PR #1, opened 2015-02-11
Complete drop and create, add JSON integration tests

osheroff:
@vanchi-zendesk I'm not sure how exactly we'll do plusones and reviews here.  Probably have to marathon it at some point...


src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
```diff
@@ -59,7 +71,8 @@ drop_column: DROP COLUMN? old_col_name;
 alter_rename_table: RENAME (TO | AS) table_name;
 
 convert_to_character_set: CONVERT TO charset_token charset_name collation?;
-default_character_set: DEFAULT? charset_token '=' charset_name ( COLLATE '=' (IDENT | STRING_LITERAL) )?;
+default_character_set: DEFAULT? charset_token '=' charset_name;
+default_collate: DEFAULT? COLLATE '=' (IDENT | STRING_LITERAL)?;
```
vanchi-zendesk: The mysql page says COLLATE can be used only with charset change..

```
  | CONVERT TO CHARACTER SET charset_name [COLLATE collation_name]
  | [DEFAULT] CHARACTER SET [=] charset_name [COLLATE [=] collation_name]
```
Also = is optional.

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
```diff
@@ -45,11 +57,11 @@ protected Table findTable(Schema schema, String dbName, String tableName) throws
 
 		MysqlParserListener listener = new MysqlParserListener(currentDB);
 
-		System.out.println("Running parse on " + sql);
+		LOGGER.debug("SQL_PARSE <- \"" + sql + "\"");
 		ParseTree tree = parser.parse();
 
 		ParseTreeWalker.DEFAULT.walk(listener, tree);
-		System.out.println(tree.toStringTree(parser));
+		LOGGER.debug("SQL_PARSE ->   " + tree.toStringTree(parser));
```
vanchi-zendesk: get an input stream, get a lexer on the input stream, get a token stream from lexer, get a parser on the token stream, get a listener on the DB, then parse into a tree, then walk the tree with the listener. 

What a function! Don't have any suggestions though. :D This code has to go somewhere.

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseCreate.java
```diff
@@ -0,0 +1,32 @@
...
+		this.encoding = encoding;
+	}
+
+	@Override
+	public Schema apply(Schema originalSchema) throws SchemaSyncError {
+		Database database = originalSchema.findDatabase(dbName);
+
+		if ( database != null && !ifNotExists) {
+			throw new SchemaSyncError("Unexpectedly asked to create existing database " + dbName);
+		}
```
vanchi-zendesk: else return newSchema?

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseDrop.java
```diff
@@ -0,0 +1,28 @@
...
+
+	public DatabaseDrop(String dbName, boolean ifExists) {
+		this.dbName = dbName;
+		this.ifExists = ifExists;
+	}
+
+	@Override
+	public Schema apply(Schema originalSchema) throws SchemaSyncError {
+		Schema newSchema = originalSchema.copy();
+		Database database = findDatabase(newSchema, dbName, ifExists);
```
vanchi-zendesk: nitpic: The interchanging usage of findDatabase within SchemaChange and within Schema is confusing.

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseCreate.java
osheroff: good point.  would end up duplicating.

src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
osheroff: ok, will fix

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseDrop.java
```diff
@@ -0,0 +1,28 @@
...
+
+		if ( database == null && !ifExists ) {
+			throw new SchemaSyncError("Can't drop missing database: " + dbName);
```
vanchi-zendesk: dead code? ifExists == ignoreMissing, so if it was false, findDatabase would have thrown?

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseDrop.java
osheroff: k.  `getDatabase()`?  `requireDatabase()`?

osheroff: I could also just move the `throw new SchemaSyncError()` stuff into `Schema` | `Database` or whatever

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
osheroff: java, not known for brevity.

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseDrop.java
```diff
@@ -0,0 +1,28 @@
...
+		}
+
+		newSchema.getDatabases().remove(database);
```
vanchi-zendesk: only if database != null.

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -7,9 +7,11 @@
 import org.antlr.v4.runtime.tree.ErrorNode;
 
 import com.zendesk.maxwell.schema.columndef.ColumnDef;
-import com.zendesk.maxwell.schema.ddl.mysqlParser.Table_nameContext;
 import com.zendesk.maxwell.schema.ddl.mysqlParser.*;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 class MaxwellSQLSyntaxRrror extends RuntimeException {
```
vanchi-zendesk: Is this a typo?

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseDrop.java
vanchi-zendesk: Oh, the comment is in the wrong place.. You could replace this https://github.com/zendesk/maxwell/pull/1/files#diff-12a4be33f9def428f8ca27f083c7b496R19 instead with findDatabase...

vanchi-zendesk: Anyway, only a nitpic.

vanchi-zendesk:
:+1:  with comment changes..

src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseDrop.java
osheroff: true.

osheroff: err, no, false.

osheroff: i think you're right and that I should sacrifice a few lines of code for increased clarity and code locality.  will re-do the findDatabase and findTable stuff.

osheroff:
take another look, all comments should be addressed

vanchi-zendesk:
checking.

vanchi-zendesk:
:+1:

PR #2, opened 2015-02-18
Save schema in mysql

osheroff:
when we record schema changes, we'll stash them in mysql.


src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -0,0 +1,197 @@
...
+	public void save() throws SQLException, IOException {
+		createStoreDatabase();
+
+		if (this.schema == null)
+			throw new RuntimeException("Uninitialized schema!");
+
+		Integer schemaId = executeInsert(schemaInsert, position.getFile(),
+				position.getOffset(), 1);
+
+		for (Database d : schema.getDatabases()) {
```
vanchi-zendesk: Is this fine? Should this whole thing not be a transaction, maybe?

vanchi-zendesk: Also, absolutely no error checks. Is it ok to fail catastrophically here if the save crashes?

osheroff: definitely needs a transaction.  probably ok to fail horribly if the save fails.

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -0,0 +1,197 @@
...
+					cRS.getString("coltype"), i++,
+					cRS.getInt("is_signed") == 1);
+			t.getColumnList().add(c);
+		}
+	}
+
+	private ResultSet findSchemaID(BinlogPosition targetPosition)
+			throws SQLException {
+		PreparedStatement s = connection
+				.prepareStatement("SELECT * from `maxwell`.`schemas` where binlog_file <= ? and binlog_position <= ? "
```
vanchi-zendesk: If the file changes, does the offset remain increasing? If not, this may be wrong... (I hope not)

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -144,6 +117,15 @@ private void renumberColumns() {
 			c.setPos(i++);
 		}
 	}
+
+	public void setDefaultColumnEncodings() {
```
vanchi-zendesk: nitpic: Just noticing an inconsistency here... Table refers to its database and falls back to its encoding. But column does not refer to its table. And table forces its default encoding wherever columns are changed or added.

vanchi-zendesk:
:+1: with some comments.

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
osheroff: when you say "offset", do you mean the binlog position?  it doesn't, it refers to a very literal file offset on disk

osheroff: oh, yeah, you're right.  it should read:

```
  (binlog_file < ?) OR (binlog_file = ? AND binlog_position <= ?)
```

right?

osheroff: *very* nice catch.

src/main/java/com/zendesk/maxwell/schema/Table.java
osheroff: yeah, good call, I think `Table` should do the same thing that `ColumnDef` does, and only fallback to the database's default encoding on table creation.

osheroff: because if a database's default encoding changes, its subordinate table's default encodings do *not* change.  I checked.

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
vanchi-zendesk: thanks!

src/main/java/com/zendesk/maxwell/schema/Table.java
vanchi-zendesk: Oh hell :/ I was hoping you would go the other way and simplify code. But,... reality.

osheroff:
updated from CR comments, will fix the broader issue of delegated encodings in another PR

vanchi-zendesk:
:+1:

osheroff:
merged this into the side branch by accident, merging it to master now

PR #3, opened 2015-02-19
implement "create table like table" parsing

osheroff:
@vanchi-zendesk


src/main/java/com/zendesk/maxwell/schema/ddl/TableCreate.java
```diff
@@ -13,6 +13,9 @@
 	public ArrayList<ColumnDef> columns;
 	public String encoding;
 
+	public String likeDB;
+	public String likeTable;
```
vanchi-zendesk: nitpic: Maybe these should be likeDBName and likeTableName to improve clarity.

vanchi-zendesk:
:+1:

PR #4, opened 2015-02-20
WIP: kafka support, as well as streaming binlogs

osheroff:
@vanchi-zendesk not much to see yet but may as well code in public

### DIFF
--- a/src/main/java/com/zendesk/maxwell/BinlogPosition.java
+++ b/src/main/java/com/zendesk/maxwell/BinlogPosition.java
@@ -5,27 +5,22 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 
 public class BinlogPosition {
-	private final int offset;
+	private final long offset;
 	private final String file;
 
-	public BinlogPosition(int offset, String file) {
-		this.offset = offset;
+	public BinlogPosition(long l, String file) {
+		this.offset = l;
 		this.file = file;
 	}
 
-	public static BinlogPosition capture(Connection c) {
+	public static BinlogPosition capture(Connection c) throws SQLException {
 		ResultSet rs;
-		try {
-			rs = c.createStatement().executeQuery("SHOW MASTER STATUS");
-			rs.next();
-            return new BinlogPosition(rs.getInt("Position"), rs.getString("File"));
-		} catch (SQLException e) {
-			// TODO be smart about permission denied here
-			return null;
-		}
+		rs = c.createStatement().executeQuery("SHOW MASTER STATUS");
+		rs.next();
+		return new BinlogPosition(rs.getInt("Position"), rs.getString("File"));
 	}
 
-	public int getOffset() {
+	public long getOffset() {
 		return offset;
 	}
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -1,0 +1,75 @@
+package com.zendesk.maxwell;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Properties;
+
+public class MaxwellConfig {
+	public String mysqlHost;
+	public int    mysqlPort;
+	public String mysqlUser;
+	public String mysqlPassword;
+
+	public String currentPositionFile;
+
+	private BinlogPosition initialPosition;
+
+	public Connection getMasterConnection() throws SQLException {
+		return DriverManager.getConnection("jdbc:mysql://" + mysqlHost + ":" + mysqlPort, mysqlUser, mysqlPassword);
+	}
+
+	public BinlogPosition getInitialPosition() throws FileNotFoundException, IOException {
+		if ( this.initialPosition != null )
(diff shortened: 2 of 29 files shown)


osheroff:
@vanchi-zendesk this is ready for review, but it didn't actually include kafka support; instead it isolated out booting up maxwell from the commandline, basic config file style stuff, and storing the current binlog position in a file.

src/main/java/com/zendesk/maxwell/BinlogPosition.java
```diff
@@ -5,27 +5,22 @@
...
-			rs = c.createStatement().executeQuery("SHOW MASTER STATUS");
-			rs.next();
-            return new BinlogPosition(rs.getInt("Position"), rs.getString("File"));
-		} catch (SQLException e) {
-			// TODO be smart about permission denied here
-			return null;
-		}
+		rs = c.createStatement().executeQuery("SHOW MASTER STATUS");
+		rs.next();
+		return new BinlogPosition(rs.getInt("Position"), rs.getString("File"));
```
osheroff: here there's not much code change, but I'm throwing the error instead of papering over it.

osheroff:
pingo

vanchi-zendesk:
​Sorry, I will get to this tomorrow. Too much firefighting this week.​

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -0,0 +1,75 @@
...
+	public int    mysqlPort;
+	public String mysqlUser;
+	public String mysqlPassword;
+
+	public String currentPositionFile;
+
+	private BinlogPosition initialPosition;
+
+	public Connection getMasterConnection() throws SQLException {
+		return DriverManager.getConnection("jdbc:mysql://" + mysqlHost + ":" + mysqlPort, mysqlUser, mysqlPassword);
```
vanchi-zendesk: This probably needs to be rewritten, (in my pr) ?

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -0,0 +1,75 @@
...
+
+		File f = new File(this.currentPositionFile);
+		FileWriter fw = new FileWriter(f);
+		try {
+			p.store(fw, "");
+			this.initialPosition = position;
+		} finally {
+			fw.close();
+		}
+	}
```
vanchi-zendesk: I am surprised this is in a file... I thought you would read from mysql... Is this stop-gap?

src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -34,24 +34,86 @@
...
 
 		TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
 		this.binlogEventListener = new MaxwellBinlogEventListener(queue);
+
+		this.replicator = new OpenReplicator();
+
+		this.replicator.setHost("127.0.0.1");
+		this.replicator.setUser("maxwell");
+		this.replicator.setPassword("maxwell");
+		this.replicator.setBinlogEventListener(this.binlogEventListener);
```
vanchi-zendesk: I am guessing, these are defaults. If they are, port is missing.

src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -117,143 +177,10 @@ private void processQueryEvent(QueryEvent event) throws SchemaSyncError {
...
-		r.registgerEventParser(new DeleteRowsEventV2Parser());
-		r.registgerEventParser(new FormatDescriptionEventParser());
-
-		return r;
-	}
-
-	private void initParser(String fileName, long position) throws Exception {
-		FileBasedBinlogParser bp = getDefaultBinlogParser();
-
-		bp.setStartPosition(position);
```
vanchi-zendesk: This is no longer needed because we are directly connecting to tcp stream?

vanchi-zendesk:
:+1:  Sorry for the delay.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: yup, eventually.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: haven't 100% decided where to store last-position.  probably mysql, but we might also just write the file/position into the kafka stream and pop the last-written off the stream to recover the position...

src/main/java/com/zendesk/maxwell/MaxwellParser.java
osheroff: I think these can go away.

src/main/java/com/zendesk/maxwell/MaxwellParser.java
osheroff: we're using the OpenReplicator's objects directly instead of implementing our own.

vanchi-zendesk: Ah ok. I was talking about setStartPosition() which is missing in the other code.. We set the FileName and FilePath in setBinlogPosition().

vanchi-zendesk: Ha, I am wrong. You are setting the offset up there. It is setBinlogFilePath which is missing now.

PR #5, opened 2015-03-01
Sets and enums

osheroff:
@vanchi-zendesk I had been avoiding sets and enum support for awhile, thinking it was hard.  It wasn't, particularly...

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
@@ -6,7 +6,10 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -101,6 +104,7 @@ public class SchemaCapturer {
 		ResultSet r = infoSchemaStmt.executeQuery();
 
 		while(r.next()) {
+			String[] enumValues = null;
 			String colName    = r.getString("COLUMN_NAME");
 			String colType    = r.getString("DATA_TYPE");
 			String colEnc     = r.getString("CHARACTER_SET_NAME");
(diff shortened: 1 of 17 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
```diff
@@ -110,8 +114,26 @@ private void captureTable(Table t) throws SQLException {
...
+			}
+
+			t.getColumnList().add(ColumnDef.build(t.getName(), colName, colEnc, colType, colPos, colSigned, enumValues));
 			i++;
 		}
 	}
+
+	private static String[] extractEnumValues(String expandedType) {
+		String[] enumValues;
+		Matcher matcher = Pattern.compile("(enum|set)\\((.*)\\)").matcher(expandedType);
```
vanchi-zendesk: what's up with the double slash?

src/main/java/com/zendesk/maxwell/schema/columndef/SetColumnDef.java
```diff
@@ -0,0 +1,40 @@
...
+		return asList(value);
+	}
+
+	private ArrayList<String> asList(Object value) {
+		ArrayList<String> values = new ArrayList<>();
+		long v = (Long) value;
+		for(int i = 0; i < enumValues.length; i++) {
+			if ( ((v >> i) & 1) == 1 ) {
+				values.add(enumValues[i]);
+			}
```
vanchi-zendesk: Oh dear god. Why does mysql have this? Do we ever use it?

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/columndef/SetColumnDef.java
osheroff: sets are useful!  Zendesk doesn't tend to use them...

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
osheroff: it allows you to pass the backslash into the regexp engine to escape the parens: the string reads literally `(enum|set)\((.*)\)`

osheroff: that is to say, a backslash in a java string will normally escape the next character.

PR #6, opened 2015-03-03
miscellaneous fixes 

osheroff:
@vanchi-zendesk in prepping for kafka, found these little things wrong with maxwell.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
@@ -24,8 +24,8 @@ public class SchemaStore {
 	private Schema schema;
 	private BinlogPosition position;
 	static final Logger LOGGER = LoggerFactory.getLogger(SchemaStore.class);
-	private final PreparedStatement schemaInsert, databaseInsert, tableInsert,
-			columnInsert;
+	private final PreparedStatement schemaInsert, databaseInsert, tableInsert;
+	private final String columnInsertSQL;
 
 	public SchemaStore(Connection connection) throws SQLException {
 		this.connection = connection;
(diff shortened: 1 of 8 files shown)


src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
```diff
@@ -18,21 +20,36 @@
...
+
+	}
+
 	public static List<SchemaChange> parse(String currentDB, String sql) {
-		ANTLRInputStream input = new ANTLRInputStream(sql);
-		mysqlLexer lexer = new mysqlLexer(input);
-		CommonTokenStream tokens = new CommonTokenStream(lexer);
-		mysqlParser parser = new mysqlParser(tokens);
+		if ( SQL_BLACKLIST.contains(sql))
+			return null;
```
vanchi-zendesk: what's this stuff? I thought the binlog stream does not have transactional stuff..? (I mean, only write after it succeeds no?)

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -100,11 +101,38 @@ private void saveSchema() throws SQLException {
...
+		}
+
+		PreparedStatement columnInsert = connection.prepareStatement(insertColumnSQL);
+		int i = 1;
+
+		for (Object o : columnData)
+			columnInsert.setObject(i++,  o);
+
+		columnInsert.execute();
+		columnData.clear();
```
vanchi-zendesk: nice shot.

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
osheroff: I'm not totally sure, but I think that for transactions that touch multiple tables you'll see a `BEGIN` at the the top of them; you have to keep things atomic from the slaves' perspective too... ? 

I don't actually know.

PR #9, opened 2015-03-05
change JSON output to what the binlog looks like.

osheroff:
this changes the data format output from

```
[
 {table: 'foo', data: {'bar': 'baz'},
 {table: 'foo', data: {'bar': 'zab'}
]
```

to

```
{
  table: 'foo',
  data: [ {'bar': 'baz'},
          {'bar': 'zab'} ]
}
```

@vanchi-zendesk


vanchi-zendesk:
:+1:

PR #10, opened 2015-03-06
Some small fixes

osheroff:
mostly stuff around getting a release together, but also a bug fix or two.

@vanchi-zendesk


osheroff:
ping

vanchi-zendesk:
The test fail is fine?
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0

at java.util.ArrayList.rangeCheck(ArrayList.java:635)
```

vanchi-zendesk:
changes look ok to me. But I am still concerned by the strange test fail.

+1

osheroff:
yeah I think it's a race condition ultimately; if the binlog parser thread hasn't gotten going by the time `.getEvent()` is called it will just return early.  I will look into something more robust later.

PR #15, opened 2015-03-09
Deal with invalid datetimes 

osheroff:
Mysql allows us to store invalid datetimes; the most common of these is '0000-00-00 00:00:00'.  In standard open-replicator, we try to bring this into a java datetime, which then *sorta* fails.  In 1.2.0 of open-replicator we preserve the underlying mysql value as a long so that we can cast it back into its original value.  And then here we use that original value instead of the parsed datetime.

mysql is ugly.

@vanchi-zendesk

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
@@ -15,6 +15,7 @@ import com.google.code.or.binlog.BinlogEventV4Header;
 import com.google.code.or.binlog.impl.event.AbstractRowEvent;
 import com.google.code.or.common.glossary.Column;
 import com.google.code.or.common.glossary.Row;
+import com.google.code.or.common.glossary.column.DatetimeColumn;
 import com.zendesk.maxwell.schema.Table;
 import com.zendesk.maxwell.schema.columndef.ColumnDef;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -193,7 +195,11 @@ public RowMap jsonMap() {
 				Column c = colIter.next();
 				ColumnDef d = defIter.next();
 
-				Object value = c.getValue();
+				if ( c instanceof DatetimeColumn )
+					value = ((DatetimeColumn) c).getLongValue();
```
vanchi-zendesk: how does this work? If the value is of type "YYYY-MM-DD HH:MM:SS" ? How are we getting a long value of this?

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
```diff
@@ -44,6 +46,18 @@ else if ( value instanceof Timestamp )
 			return "";
 	}
 
+	private String formatLong(Long value) {
+		final int second = (int)(value % 100); value /= 100;
+		final int minute = (int)(value % 100); value /= 100;
+		final int hour = (int)(value % 100); value /= 100;
+		final int day = (int)(value % 100); value /= 100;
+		final int month = (int)(value % 100);
+		final int year = (int)(value / 100);
```
vanchi-zendesk: Hm. Is the open replicator doing this encoding underneath? "2014-01-02 12:40:56" => 20140102124056

osheroff: no, this is mysql itself.

vanchi-zendesk:
:+1:

PR #16, opened 2015-03-13
Store binlog position in database

osheroff:
@vanchi-zendesk this is a first draft of storing the current binlog position in the database.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -34,7 +34,7 @@ public class Maxwell {
 	private void run(String[] args) throws Exception {
 		this.config = MaxwellConfig.buildConfig("config.properties", args);
 
-		SchemaStore.createMaxwellSchema(this.config.getMasterConnection());
+		SchemaStore.ensureMaxwellSchema(this.config.getMasterConnection());
 
 		if ( this.config.getInitialPosition() != null ) {
 			LOGGER.info("Maxwell is booting, starting at " + this.config.getInitialPosition());
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -20,8 +20,7 @@ public class Maxwell {
 	private void initFirstRun() throws SQLException, IOException {
 		Connection connection = this.config.getMasterConnection();
 
-		SchemaStore.createMaxwellSchema(connection);
-
+		LOGGER.info("Maxwell is capturing initial schema");
 		SchemaCapturer capturer = new SchemaCapturer(connection);
 		this.schema = capturer.capture();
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -35,6 +34,8 @@ private void initFirstRun() throws SQLException, IOException {
 	private void run(String[] args) throws Exception {
 		this.config = MaxwellConfig.buildConfig("config.properties", args);
 
+		SchemaStore.createMaxwellSchema(this.config.getMasterConnection());
+
```
vanchi-zendesk: Don't understand this. This needs to happen only exactly once right? So it was at it's right place in first run?

osheroff: it needs to happen before `getInitialPosition()...`

osheroff: and it's a misnomer, it should be called `createIfNotExsistsMaxwellSchema`...

src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -92,17 +92,23 @@ public void run() throws Exception {
...
-			this.config.setInitialPosition(p);
+				BinlogPosition p = eventBinlogPosition(event);
+				this.config.setInitialPosition(p);
+			}
 		}
 	}
 
+	private boolean skipEvent(MaxwellAbstractRowsEvent event) {
+		return event.getTable().getDatabase().getName().equals("maxwell");
+	}
```
vanchi-zendesk: If you think about this a bit, this is fine. 

Migration -> maxwell writes -> kafka push. (so if we follow maxwell, we get the schema changes as snapshots)

But I think it's best to not complicate at the beginning :)

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -0,0 +1,39 @@
...
+
+		return new BinlogPosition(rs.getLong("binlog_position"), rs.getString("binlog_file"));
+	}
+
+	public static void set(Connection c, Long serverID, BinlogPosition p) throws SQLException {
+		String sql = "INSERT INTO `maxwell`.`positions` set "
+				+ "server_id = ?, "
+				+ "binlog_file = ?, "
+				+ "binlog_position = ? "
+				+ "ON DUPLICATE KEY UPDATE binlog_file=?, binlog_position=?";
```
vanchi-zendesk: could we use REPLACE INTO and save lines?

src/main/java/com/zendesk/maxwell/Maxwell.java
vanchi-zendesk: right. I suspected.

vanchi-zendesk: `EnsureMaxwellSchema`

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/MaxwellParser.java
osheroff: not with the introduction of the maxwell.position table, which would keep the system in an infinite loop

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
osheroff: on duplicate key is way way faster:

http://stackoverflow.com/questions/9168928/what-are-practical-differences-between-replace-and-insert-on-duplicate-ke

PR #17, opened 2015-03-18
binlog position flush thread

osheroff:
flush the binlog position to the server every 1 second.

@vanchi-zendesk

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
@@ -58,7 +58,7 @@ public class SchemaPosition implements Runnable {
 				+ "binlog_file = ?, "
 				+ "binlog_position = ? "
 				+ "ON DUPLICATE KEY UPDATE binlog_file=?, binlog_position=?";
-		PreparedStatement s = connection.prepareStatement(sql);
+		PreparedStatement s = this.connection.prepareStatement(sql);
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -4,36 +4,104 @@
...
+	private Thread thread;
 
-		ResultSet rs = s.executeQuery();
-		if ( !rs.next() )
-			return null;
+	public SchemaPosition(Connection c, Long serverID) {
+		this.connection = c;
+		this.serverID = serverID;
+		this.lastPosition = null;
+		this.position = new AtomicReference<>();
```
vanchi-zendesk: Should this be <BinlogPosition> ? 

Does it make the entire position object access under mutex.? (not just the reference var)

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
osheroff: this is java 7 syntax-fu, it implies the type of the generic from the variable type. 

to your second question: no.  Only the `.set()` and `.get()` calls are protected, so in *theory* someone could be mutating the `BinlogPosition` object underneath us.  In practice I think we can just not do that.

PR #18, opened 2015-03-21
ignore a bunch of DDL statements we don't care about

osheroff:
@vanchi-zendesk


osheroff:
will get a later plusone on this one.

vanchi-zendesk:
:+1:

PR #19, opened 2015-03-22
force Maxwell into running under a utf8 locale

osheroff:
otherwise java is mangling strings.  I suppose there's something I'm doing wrong, but this will fix the thing.

@vanchi-zendesk


vanchi-zendesk:
:+1:

PR #22, opened 2015-03-24
rework setInitialPosition()

osheroff:
- only do the setInitialPosition() call in the producer.  
- mild refactoring

@vanchi-zendesk

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellParser.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellParser.java
@@ -93,12 +93,9 @@ public class MaxwellParser {
 			if ( !skipEvent(event)) {
 				producer.push(event);
 
-				// TODO:  this isn't quite right: we need to only stop on table map events,
-				// although it's unclear to me whether you can have two row events in a row.
-				// I think maybe you can.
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -93,12 +93,9 @@ public void run() throws Exception {
 			if ( !skipEvent(event)) {
 				producer.push(event);
 
-				// TODO:  this isn't quite right: we need to only stop on table map events,
-				// although it's unclear to me whether you can have two row events in a row.
-				// I think maybe you can.
+				// TODO:  this isn't quite right: we need to only stop on table map events
 
 				BinlogPosition p = eventBinlogPosition(event);
-				this.config.setInitialPosition(p);
```
vanchi-zendesk: do we use p after this? Maybe we can remove eventBinlogPosition as well?

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/MaxwellParser.java
osheroff: yup fixed

vanchi-zendesk:
:+1:

PR #25, opened 2015-03-25
Connection pooling

osheroff:
@vanchi-zendesk you were right.  maxwell was crashing over night when it hit a schema change, as the server had closed its connection.  

Try viewing this PR with `?w=1` -- a lot of it is just unavoidable whitespace change


vanchi-zendesk:
:+1:

PR #27, opened 2015-03-25
ensure row replication enabled before starting

vanchi-zendesk:
Check log_bin and binlog_format before starting each time.

@zendesk/rules


osheroff:
:+1: although it seems like maybe there's a use for a `MysqlMasterInformation` class...

PR #32, opened 2015-03-27
primary key handling, part 1

osheroff:
capture primary keys from the SchemaCapture module.

@zendesk/rules


vanchi-zendesk:
:+1: ( I assume part 2 will be parsing the table alteration)

PR #33, opened 2015-03-27
Primary Key storing, part two

osheroff:


src/main/java/com/zendesk/maxwell/schema/SchemaStoreMigrations.java
```diff
@@ -0,0 +1,38 @@
...
+	static final String pkAlterSQL =
+			"alter table maxwell.columns add column pk varchar(1024) NULL default NULL";
+
+	private void upgradeColumnsTable() throws SQLException {
+		ResultSet rs = connection.createStatement().executeQuery(pkTestSQL);
+		if ( rs.next() ) {
+			return;
+		} else {
+			LOGGER.info("upgrading maxwell.columns to include PK column");
+			connection.createStatement().executeQuery(pkAlterSQL);
```
vanchi-zendesk: This looks fine for now. Do we add a systematic way (list of tests/upgrade queries to run) only after we actually need it?

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -213,15 +222,21 @@ private Database restoreDatabase(int id, String name, String encoding) throws SQ
 		while (tRS.next()) {
 			String tName = tRS.getString("name");
 			String tEncoding = tRS.getString("encoding");
+			String tPKs = null;
+			try {
+				tPKs = tRS.getString("pk");
+			} catch ( SQLException e ) {}
+
```
vanchi-zendesk: Since the first thing we do when maxwell starts is: migrate maxwell db, capture the table pks, and update the current maxwell tables.. do we need a try block? You will get NULL anyway?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreMigrations.java
```diff
@@ -0,0 +1,82 @@
...
+
+	private void upgradeColumnsTable() throws SQLException, SchemaSyncError {
+		ResultSet rs = connection.createStatement().executeQuery(pkTestSQL);
+		if ( rs.next() ) {
+			return;
+		} else {
+			LOGGER.info("upgrading maxwell.tables to include PK column");
+			SchemaStore ss = verifySchemaUnchanged();
+			connection.createStatement().execute(pkAlterSQL);
+			overwriteSchema(ss);
```
vanchi-zendesk: Just to be safe, I think we should retest your testSQL here to prove that,  you know the testSQL is fine and AlterSQL negates testSQL. We should fail hard if not.

src/main/java/com/zendesk/maxwell/schema/SchemaStoreMigrations.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		SchemaCapturer capturer = new SchemaCapturer(connection);
+		Schema schema = capturer.capture();
+
+		if ( !schema.equals(lastStoredSchema.getSchema()) ) {
+			String diffs = StringUtils.join(
+					lastStoredSchema.getSchema().diff(schema, "last captured", "now captured").toArray(),
+					"\n");
+			throw new SchemaSyncError("Couldn't upgrade schema.  Please drop maxwell db and try again:\n" + diffs);
```
vanchi-zendesk: This warning should be a bit more dire. We are starting over, essentially.

src/main/java/com/zendesk/maxwell/schema/SchemaStoreMigrations.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+		lastStoredSchema.setSchema(schema);
+		return lastStoredSchema;
+	}
+
+	private void overwriteSchema(SchemaStore schemaStore) throws SQLException, SchemaSyncError {
+		try {
+			connection.setAutoCommit(false);
```
vanchi-zendesk: So, should we not update ss with the pk information before restoring ?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreMigrations.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	// If the schema change while we were down, we can't upgrade.
+	// throw SchemaSyncError in this case.
+
+	// Otherwise, return the old schema-store; we'll overwrite it definition.
+	private SchemaStore verifySchemaUnchanged() throws SQLException, SchemaSyncError  {
+		BinlogPosition pos = BinlogPosition.capture(connection);
+		SchemaStore lastStoredSchema = SchemaStore.restore(connection, pos);
```
vanchi-zendesk: So this will not have the PK info right? So it will always be different from schema down below because of that? (The part-1 PR makes sure capture can get pk info)

osheroff:
as discussed, I'm going to simplify by eliminating the upgrade path for PK storage.  stay tuned.

osheroff:
ok, removed the migration code, ping.

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -165,18 +165,29 @@ public String getEncoding() {
 		return this.pkColumnNames;
 	}
 
-	public void setPK(List<String> pkColumnNames) throws SchemaSyncError {
+	public Object getPKString() {
+		if ( this.pkColumnNames != null )
+			return StringUtils.join(pkColumnNames.iterator(), ",");
```
vanchi-zendesk: Do we need to qualify the column-names with quotes? Is it possible to have a column called "a,b" ?

vanchi-zendesk:
The open replicator jar went away, without anything new getting added. Is that fine?

vanchi-zendesk:
All else looks fine but the two points above.

vanchi-zendesk:
:+1: (add a caveat in the readme maybe)

PR #34, opened 2015-03-27
check read_only is OFF, also move mysql status to a class.

vanchi-zendesk:
@zendesk/rules 

Make code nicer.


osheroff:
cool :+1:

PR #37, opened 2015-04-08
Handle primary key tracking in create / alter table

osheroff:
@vanchi-zendesk @dadah89

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
+++ b/src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
@@ -41,6 +41,7 @@ public class MysqlParserListener extends mysqlBaseListener {
 	private ArrayList<String> pkColumns;
 
 	MysqlParserListener(String currentDatabase)  {
+		this.pkColumns = null; // null indicates no change in primary keys
 		this.schemaChanges = new ArrayList<>();
 		this.currentDatabase = currentDatabase;
 	}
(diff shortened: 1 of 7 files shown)


src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -36,6 +38,8 @@ public MaxwellSQLSyntaxError(String message) {
 
 	private final LinkedList<ColumnDef> columnDefs = new LinkedList<>();
 
+	private ArrayList<String> pkColumns;
```
vanchi-zendesk: can't we do  `= new ArrayList<>();` to avoid the ugly check above? (just nitpic)

vanchi-zendesk: right, never mind this. I think null here is a valid value if there is no change in pks in the alter.

src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
```diff
@@ -118,4 +118,17 @@ public void testDatabaseEncoding() throws SQLException, SchemaSyncError, IOExcep
...
 		testIntegration(sql);
 	}
+
+	@Test
+	public void testPKs() throws SQLException, SchemaSyncError, IOException {
+		String sql[] = {
+		   "create TABLE `test_pks` ( id int(11) unsigned primary KEY, str varchar(255) )",
+		   "create TABLE `test_pks_2` ( id int(11) unsigned, str varchar(255), primary key(id, str) )",
+		   "create TABLE `test_pks_3` ( id int(11) unsigned primary KEY, str varchar(255) )",
+		   "alter TABLE `test_pks_3` drop primary key, add primary key(str)"
```
vanchi-zendesk: TODO: can we also test simply just drop too ?

vanchi-zendesk:
There's a bit of confusion in the code, related to use of null, vs [] for the pk list. Maybe it should be explicit somehow, if pks are to be left alone (I think currently done with null), and are to be replaced with new values.

vanchi-zendesk:
:+1:

PR #38, opened 2015-04-08
go back to row-by-row output

osheroff:
since we're going to want to keep the last copy of the row after
compaction, it no longer makes sense for us to batch together bulk
inserts/updates.

This will cost more overhead in databases where that's concerned, but
it's probably an acceptable tradeoff.

@vanchi-zendesk @dadah89

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -2,6 +2,7 @@ package com.zendesk.maxwell.producer;
 
 import java.io.IOException;
 import java.sql.SQLException;
+import java.util.Iterator;
 import java.util.Properties;
 
 import com.zendesk.maxwell.MaxwellAbstractRowsEvent;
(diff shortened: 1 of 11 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -32,12 +38,14 @@ public void onCompletion(RecordMetadata md, Exception e) {
...
-					LOGGER.debug("   " + event.toJSON());
+					LOGGER.debug("   " + this.json);
 					LOGGER.debug("   " + event.getNextBinlogPosition());
 					LOGGER.debug("");
 				}
-				config.setInitialPosition(event.getNextBinlogPosition());
-			} catch (IOException | SQLException e1) {
+				if ( this.lastRowInEvent ) {
+					config.setInitialPosition(event.getNextBinlogPosition());
+				}
```
vanchi-zendesk: So this was the original reason we did this? I mean, what if it fails at this point...etc ?

Maybe this needs to be mentioned in the docs,  since maxwell may crash at this point and has to repeat all these rows.

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
osheroff: yeah, if it fails mid-way through we can repeat rows.  We should do a whole bit on the consistancy guarantees, ie what happens if you change masters, the one second flush window, this, etc...

PR #39, opened 2015-04-08
refactor-only -- keep a Map of (column name -> offsets) around

osheroff:
this PR adds no new functionality, but it speeds up `.findColumnIndex()` which we'll use to generate a PK string.

@vanchi-zendesk

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/Table.java
+++ b/src/main/java/com/zendesk/maxwell/schema/Table.java
@@ -26,7 +26,7 @@ public class Table {
 		this.columnList = list;
 		if ( pks == null )
 			pks = new ArrayList<String>();
-		this.setPK(pks);
+		this.setPKList(pks);
 		renumberColumns();
 	}
 
@@ -169,7 +169,7 @@ public class Table {
 		}
 	}
 
-	public List<String> getPK() {
+	public List<String> getPKList() {
 		return this.pkColumnNames;
 	}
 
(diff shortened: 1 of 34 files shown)


src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -180,22 +198,7 @@ public String getPKString() {
...
-
-		if ( pkColumnNames == null )
-			return;
-
-		for ( String name : pkColumnNames ) {
-			Integer i = findColumnIndex(name);
-
-			if ( i == -1 ) {
-				throw new RuntimeException("Couldn't find primary key '" + name + "' in table '" + name + "'");
-			}
```
vanchi-zendesk: don't know if this check was very important, but it's gone now.

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -38,15 +42,23 @@ public String getName() {
...
 		for(ColumnDef c : columnList) {
-			if ( c.getName().equals(name) )
-				return i;
-			i++;
+			this.columnOffsetMap.put(c.getName(), i++);
+		}
+	}
+
+	public int findColumnIndex(String name) {
+		initColumnOffsetMap();
```
vanchi-zendesk: ok, does this really speed up things? we are doing this again here..

osheroff: oh.  good point, I need to memoize.

osheroff:
memoized.

vanchi-zendesk:
:+1:

PR #40, opened 2015-04-09
ignore "flush privileges" .*

osheroff:
@vanchi-zendesk maxwell is stopped again.  

eventually I'll just make it ignore parse errors, but I want to be sure we've caught most of the real issues before I do that.


vanchi-zendesk:
:+1:

PR #42, opened 2015-04-09
primary key strings!

osheroff:
the fruition of all of last week's labor.

@vanchi-zendesk

### DIFF
--- a/docs/docs/kafka.md
+++ b/docs/docs/kafka.md
@@ -11,6 +11,18 @@ Maxwell sets the following Kafka options by default, but you can override them i
 
 Maxwell writes to a kafka topic named "maxwell" by default.  This can be changed with the `kafka_topic` option.
 
+### Kafka key
+
+Maxwell generates keys for its Kafka messages based upon a mysql row's primary key:
+
+```
+db:test_db/tbl:test_tbl/id:93/id_2:17910090
+```
+
+This key is designed to co-operate with Kafka's log compaction, which will save the last-known
+value for a key, allowing Maxwell's Kafka stream to retain the last-known value for a row and act
+as a source of truth.
+
 ### Topic and partitioning
 
 Maxwell enforces ordering on events within a logical mysql database (but not within a mysql server).  We enforce
--- a/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
@@ -8,6 +8,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang.StringUtils;
 import org.json.JSONObject;
 
 import com.google.code.or.binlog.BinlogEventV4Header;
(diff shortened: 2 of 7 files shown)


src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -242,4 +243,33 @@ public Object getData(String string) {
...
+		for ( String pk : table.getPKList() ) {
+			int idx = table.findColumnIndex(pk);
+
+			Column column = r.getColumns().get(idx);
+			ColumnDef def = table.getColumnList().get(idx);
+
+
+			pkValueList.add(def.getName() + ":" + def.asJSON(column.getValue()).toString());
+		}
+		return StringUtils.join(pkValueList, "/");
```
vanchi-zendesk: values can contain both ":" and "/", just saying.

osheroff: there's no separator they can't contain.  Only other option I could think of would be casting the whole key to json....

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -242,4 +243,33 @@ public Object getData(String string) {
 		}
 		return list;
 	}
+
+	public List<String> getPKStrings() {
+		ArrayList<String> list = new ArrayList<>();
+
+		String pkString = "db:" + table.getDatabase().getName() + "/"
+				        + "tbl:" + table.getName() + "/";
```
vanchi-zendesk: this is counter-intuitive. You expect getPKStrings to get you PK strings only, but sure.

vanchi-zendesk:
:+1: :clap:

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
vanchi-zendesk: that's fine by me.. I mean, let JSON encoding worry about skipping stuff. Just create a map and do toJSON?

vanchi-zendesk: Also, a json key is much easier to parse out.

osheroff: not quite as simple as just creating a map -- you gotta also ensure key ordering... but ok.

vanchi-zendesk: Oh, a list of [ { col: val }, { col: val } ] I guess...

docs/docs/kafka.md
```diff
@@ -11,6 +11,18 @@ Maxwell sets the following Kafka options by default, but you can override them i
...
+
+Maxwell generates keys for its Kafka messages based upon a mysql row's primary key:
+
+```
+db:test_db/tbl:test_tbl/id:93/id_2:17910090
+```
+
+This key is designed to co-operate with Kafka's log compaction, which will save the last-known
+value for a key, allowing Maxwell's Kafka stream to retain the last-known value for a row and act
+as a source of truth.
```
vanchi-zendesk: Might want to explain this for tables without primary keys.

osheroff:
@vanchi-zendesk  now with JSON keys.

vanchi-zendesk:
:+1:

PR #44, opened 2015-04-13
parse index-prefix lengths

osheroff:
note that these could, techincally, be part of the primary key and at some point in the future we should probably handle that case.

@vanchi-zendesk


vanchi-zendesk:
:+1:

PR #45, opened 2015-04-13
ignore SAVEPOINT

osheroff:
@vanchi-zendesk


vanchi-zendesk:
:+1:

vanchi-zendesk:
Maybe ignore all of these: https://dev.mysql.com/doc/refman/5.0/en/savepoint.html ?

osheroff:
yeah, probably.  It's unclear to me whether something that's rolled-back will ever make it into the binlog; it's also unclear to me why a savepoint would be replicated.  but hey.

PR #46, opened 2015-04-13
deal with DEFINER= in trigger,view,etc creation/dropping

osheroff:
@vanchi-zendesk productionizing a sql parser is unfun


vanchi-zendesk:
:+1:

PR #47, opened 2015-04-20
REFACTOR-ONLY refactor some functionality out of MaxwellConfig

osheroff:
move responsibility for some global objects out into a MaxwellContext object -- db pool, producer object, schema position controller.

@vanchi-zendesk


vanchi-zendesk:
:+1:

PR #49, opened 2015-04-27
flush schema pos on exit

osheroff:
Also:
- we add a setInitialPositionSync that lets us know for sure the position is flushed by the time the call returns; we use this when processing schema changes.  
- add log_level command line option

@vanchi-zendesk

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -25,7 +25,7 @@ public class MaxwellConfig {
 	public String kafkaTopic;
 	public String producerType;
 	public String outputFile;
-
+	public String log_level;
 
 	public MaxwellConfig() {
 		this.kafkaProperties = new Properties();
@@ -40,6 +40,7 @@ public class MaxwellConfig {
 
 	private OptionParser getOptionParser() {
 		OptionParser parser = new OptionParser();
+		parser.accepts( "log_level", "log level, one of DEBUG|INFO|WARN|ERROR" ).withRequiredArg();
 		parser.accepts( "host", "mysql host" ).withRequiredArg();
 		parser.accepts( "user", "mysql username" ).withRequiredArg();
 		parser.accepts( "password", "mysql password" ).withRequiredArg();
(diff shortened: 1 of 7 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -70,6 +69,9 @@ public void run() {
 
 
 	private void store(BinlogPosition newPosition) {
+		if ( newPosition == null )
```
vanchi-zendesk: this is not possible, right?

osheroff: i think only if we exit before we finish the initialization path?

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -95,6 +98,20 @@ public void set(BinlogPosition p) {
 		position.set(p);
 	}
 
+	public void setSync(BinlogPosition p) {
+		LOGGER.debug("syncing binlog position: " + p);
+		position.set(p);
+		while ( true ) {
+			thread.interrupt();
+			BinlogPosition s = storedPosition.get();
+			if ( p.newerThan(s) ) {
```
vanchi-zendesk: the newerThan makes sense at other places, but confused me here a bit. It's essentially (p != s) here.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -40,6 +40,7 @@ public String getConnectionURI() {
 
 	private OptionParser getOptionParser() {
 		OptionParser parser = new OptionParser();
+		parser.accepts( "log_level", "log level, one of DEBUG|INFO|WARN|ERROR" ).withRequiredArg();
```
vanchi-zendesk: this could be optional with info as default?

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
osheroff: i think we could theoretically miss the bit where "our" position is stored?  ie another thread comes along and stores a newer position than ours before we get a chance to call `storedPosition.get`.

probably unlikely in practice, though.

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
vanchi-zendesk: ok.I think so.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: it is optional, yeah.  the default is whatever gets set in the xml file.  saying `withRequiredArg` means that `--log_level` takes exactly 1 arguments, not that `--log_level` is a required argument...

vanchi-zendesk: got it.

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
vanchi-zendesk: ok.

PR #50, opened 2015-05-01
Clean old schemas

osheroff:
@vanchi-zendesk our database is growing toooooo big

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
@@ -325,6 +325,7 @@ public class SchemaStore {
 		Long toDelete = currentSchemaId - SCHEMAS_TO_KEEP;
 		while ( toDelete > 0 && schemaExists(toDelete) ) {
 			new SchemaStore(connection, toDelete).destroy();
+			toDelete--;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -297,9 +309,23 @@ public void destroy() throws SQLException {
...
+		ResultSet rs = connection.createStatement().executeQuery("select id from maxwell.schemas where id = " + schema_id);
+		return rs.next();
+	}
+
 	public BinlogPosition getBinlogPosition() {
 		return this.position;
 	}
 
+	private void deleteOldSchemas(Long currentSchemaId) throws SQLException {
+		Long toDelete = currentSchemaId - SCHEMAS_TO_KEEP;
```
vanchi-zendesk: is this a valid assumption? (That ids are incrementing properly). Also why are we not deleting everything older?

vanchi-zendesk: guessing by the name, I think we missed a toDelete-- in there.

osheroff: derp.  yeah.  missed `--`

osheroff: and yeah, I think we can assume auto-increment works

osheroff:
fixed

vanchi-zendesk:
:+1: :)

PR #51, opened 2015-05-26
Case insensitive

osheroff:
@vanchi-zendesk

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/Database.java
+++ b/src/main/java/com/zendesk/maxwell/schema/Database.java
@@ -11,7 +11,7 @@ public class Database {
 	private final String encoding;
 
 	public Database(String name, List<Table> tables, String encoding) {
-		this.name = name;
+		this.name = name.toLowerCase();
 		if ( tables == null )
 			this.tableList = new ArrayList<>();
 		else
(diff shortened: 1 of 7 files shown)


src/main/java/com/zendesk/maxwell/schema/Database.java
```diff
@@ -33,7 +33,7 @@ public Database(String name, String encoding) {
 
 	public Table findTable(String name) {
 		for ( Table t: this.tableList ) {
```
vanchi-zendesk: probably better to do name = name.toLowerCase() here?

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -50,15 +50,16 @@ private void initColumnOffsetMap() {
 		int i = 0;
 
 		for(ColumnDef c : columnList) {
-			this.columnOffsetMap.put(c.getName(), i++);
+			this.columnOffsetMap.put(c.getName().toLowerCase(), i++);
```
vanchi-zendesk: Is this to protect against existing entries? I see column defined with lowercase name always below.

vanchi-zendesk:
You might want some tests for this one. It's not clear how the existing data is saved, and how all that will interact with this code.. (Or if the check was missed somewhere.)

src/main/java/com/zendesk/maxwell/schema/Table.java
osheroff: ah, good point, that's not necessary.

osheroff:
this is actually just a mess, https://dev.mysql.com/doc/refman/5.0/en/identifier-case-sensitivity.html and I suppose I could code for it, but, eh....

PR #56, opened 2015-06-06
allow for config files passed in via cmdline

osheroff:
this led to some cleanup/refactor on the Config class.

@zendesk/rules


osheroff:
ping

dadah89:
works as expected, :+1:

PR #57, opened 2015-06-07
add a lower metadata fetch timeout

osheroff:
in dev this was just causing a horrible wall of text.

@zendesk/rules


dadah89:
:+1:

PR #58, opened 2015-06-08
add a timestamp to the row event.

osheroff:
there's some complexity here around avoiding it in the test suite, but otherwise this is pretty simple.

@zendesk/rules


dadah89:
A bit weird that it adds ts as seconds to the json hash (guess I'm used to milliseconds with js), but it works as expected, :+1: 

```js
{"database":"zendesk_development_shard_1","table":"sla_ticket_policies","type":"insert","data":{"id":1601,"sla_policy_id":46,"ticket_id":636,"updated_at":"2015-06-08 16:23:31","account_id":1,"created_at":"2015-06-08 16:23:31"},"ts":1433780611}
```

osheroff:
yeah, no way to get the millseconds back: mysql just gives us a unix timestamp

PR #59, opened 2015-06-09
fun never stops! intX are column type aliases.

osheroff:
support column type `int8`, `int4`, etc....

@zendesk/rules


vanchi-zendesk:
:+1: Nice.

PR #61, opened 2015-06-16
recover from lost cx

osheroff:
this, along with https://github.com/osheroff/open-replicator/commit/c8960ae6f9f28842316b7cfe03e936e5ef1d6452
gets us to a place where we can recover after losing the CX to the master

@vanchi-zendesk @dadah89


vanchi-zendesk:
:+1:

Issue #67, opened 2015-06-18
Maxwell shutdown on 'bool' and 'boolean'


akshayi1:
Same as #64, #65 and #66, Maxwell can't handle 'bool' and (just to be completely sure) 'boolean'. This occurs at the `create table` stage. To get Maxwell to run again, I had to `purge binary logs before now()` and `delete from maxwell.positions`.

BOOL:

```
bin/maxwell --user='root' --password='root' --host='127.0.0.1' --producer=stdout
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
10:40:30,183 INFO  Maxwell - Maxwell is capturing initial schema
10:40:30,185 DEBUG SchemaStore - Capturing schema
10:40:30,217 DEBUG SchemaStore - Finished capturing schema
10:40:30,294 INFO  TransportImpl - connecting to host: 127.0.0.1, port: 3306
10:40:30,343 INFO  TransportImpl - connected to host: 127.0.0.1, port: 3306, context: AbstractTransport.Context[threadId=31,scramble=}_EvG<6<gt16X;]DyW(],protocolVersion=10,serverHost=127.0.0.1,serverPort=3306,serverStatus=2,serverCollation=8,serverVersion=5.6.24-0ubuntu2-log,serverCapabilities=63487]
10:40:30,344 INFO  AuthenticatorImpl - start to login, user: root, host: 127.0.0.1, port: 3306
10:40:30,348 INFO  AuthenticatorImpl - login successfully, user: root, detail: OKPacket[packetMarker=0,affectedRows=0,insertId=0,serverStatus=2,warningCount=0,message=<null>]
10:40:31,175 DEBUG SchemaPosition - Writing binlog position to maxwell.positions: BinlogPosition[mysql-logs.000003:12471]
10:55:36,558 DEBUG SchemaChange - SQL_PARSE <- "create table mmf_test(col1 smallint, col2 int, col3 bigint,  col4 float, col5 date, col6 datetime, col7 decimal(10,2), col8 varchar(500), col9 text, col10 bool)"
line 1:155 no viable alternative at input 'bool'
10:55:36,590 ERROR MysqlParserListener - (data_type bool)
[...]

Issue #66, opened 2015-06-18
Maxwell shutdown on encountering 'year' datatype


akshayi1:
Similar to #64 and #65, maxwell barfs and fails when it encounters a year datatype.

    bin/maxwell --user='root' --password='root' --host='127.0.0.1' --producer=stdout
    Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
    23:29:28,708 INFO  Maxwell - Maxwell is booting, starting at BinlogPosition[mysql-logs.000003:6982]
    23:29:28,710 DEBUG SchemaStore - looking to restore schema at target position BinlogPosition[mysql-logs.000003:6982]
    23:29:28,711 INFO  SchemaStore - Restoring schema id 7 (last modified at BinlogPosition[mysql-logs.000003:6982])
    23:29:28,753 INFO  TransportImpl - connecting to host: 127.0.0.1, port: 3306
    23:29:28,785 INFO  TransportImpl - connected to host: 127.0.0.1, port: 3306, context: AbstractTransport.Context[threadId=25,scramble=@/f8$H"l|J=4*t,{KCN?,protocolVersion=10,serverHost=127.0.0.1,serverPort=3306,serverStatus=2,serverCollation=8,serverVersion=5.6.24-0ubuntu2-log,serverCapabilities=63487]
    23:29:28,785 INFO  AuthenticatorImpl - start to login, user: root, host: 127.0.0.1, port: 3306
    23:29:28,792 INFO  AuthenticatorImpl - login successfully, user: root, detail: OKPacket[packetMarker=0,affectedRows=0,insertId=0,serverStatus=2,warningCount=0,message=<null>]
    23:29:28,974 DEBUG SchemaChange - SQL_PARSE <- "create table max_test4(dt datetime, dt_y year(4))"
    line 1:45 mismatched input '(' expecting {',', ')', AUTO_INCREMENT, COLUMN_FORMAT, COMMENT, DEFAULT, NOT, NULL, ON, PRIMARY, STORAGE, UNIQUE}
[...]

Issue #65, opened 2015-06-18
Un-Graceful exit when table with 'real' datatype is created


akshayi1:
Similar to https://github.com/zendesk/maxwell/issues/64, Maxwell chokes when it encounters a field with `real` type. 

Maxwell error with real datatype:

    22:39:47,906 DEBUG SchemaChange - SQL_PARSE <- "create table max_test2 (col1 int, col2 tinyint, col3 bigint, col4 decimal, col5 float, col6 real)"
    java.lang.IllegalArgumentException: unsupported column type real
	at com.zendesk.maxwell.schema.columndef.ColumnDef.build(ColumnDef.java:73)
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.exitColumn_definition(MysqlParserListener.java:273)
	at com.zendesk.maxwell.schema.ddl.mysqlParser$Column_definitionContext.exitRule(mysqlParser.java:4362)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.exitRule(ParseTreeWalker.java:71)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:54)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:65)
	at com.zendesk.maxwell.MaxwellParser.processQueryEvent(MaxwellParser.java:206)
	at com.zendesk.maxwell.MaxwellParser.getEvent(MaxwellParser.java:192)
	at com.zendesk.maxwell.MaxwellParser.getEvent(MaxwellParser.java:198)
	at com.zendesk.maxwell.MaxwellParser.run(MaxwellParser.java:98)
	at com.zendesk.maxwell.Maxwell.run(Maxwell.java:69)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:75)
[...]

Issue #64, opened 2015-06-18
Un-Graceful exit when encountering TRUNCATE statements


akshayi1:
I issued a `truncate table` statement and Maxwell didn't like it. To the extent that it crashed, and failed every time I tried to run it. I had to use the `purge binary logs before now()` and `delete from maxwell.positions` commands to get it running again. Here is the log:

TRUNCATE table error:

    23:12:55,915 DEBUG SchemaChange - SQL_PARSE <- "truncate table max_test3"
    line 1:0 no viable alternative at input 'truncate'
    23:12:55,941 ERROR MysqlParserListener - (statement truncate table max_test3)
    23:12:55,942 ERROR SchemaChange - Error parsing SQL: 'truncate table max_test3'
    com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: truncate
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:85)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:65)
	at com.zendesk.maxwell.MaxwellParser.processQueryEvent(MaxwellParser.java:206)
	at com.zendesk.maxwell.MaxwellParser.getEvent(MaxwellParser.java:192)
	at com.zendesk.maxwell.MaxwellParser.getEvent(MaxwellParser.java:198)
	at com.zendesk.maxwell.MaxwellParser.run(MaxwellParser.java:98)
	at com.zendesk.maxwell.Maxwell.run(Maxwell.java:69)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:75)
    23:12:55,953 DEBUG SchemaPosition - Writing binlog position to maxwell.positions: BinlogPosition[mysql-logs.000003:6424]

Maxwell does not recover after truncate statement:

    bin/maxwell --user='root' --password='root' --host='127.0.0.1' --producer=stdout
    Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
[...]

PR #68, opened 2015-06-18
fix misc bugs

osheroff:
address #64, #65, #66, #67 

- YEAR type may have a length
- TRUNCATE TABLE must be ignored
- BOOL / BOOLEAN / REAL type aliases

@vanchi-zendesk @akshayi1


src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
```diff
@@ -35,6 +35,8 @@
 		SQL_BLACKLIST.add(Pattern.compile("^ANALYZE\\s+TABLE", Pattern.CASE_INSENSITIVE));
 		SQL_BLACKLIST.add(Pattern.compile("^SET\\s+PASSWORD", Pattern.CASE_INSENSITIVE));
 		SQL_BLACKLIST.add(Pattern.compile("^(CREATE|DROP|RENAME)\\s+USER", Pattern.CASE_INSENSITIVE));
+
+		SQL_BLACKLIST.add(Pattern.compile("^TRUNCATE\\s+TABLE", Pattern.CASE_INSENSITIVE));
```
vanchi-zendesk: Suggest we add this to a Caveat section, and explain our limitation there.

vanchi-zendesk:
:+1:

src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
```diff
@@ -144,7 +144,32 @@ public void testIntX() throws Exception {
 		};
 
 		testIntegration(sql);
+	}
+
+	@Test
+	public void testYearWithLength() throws Exception {
+		String sql[] = {
+			"create TABLE `test_year` ( id year(4) )"
```
vanchi-zendesk: Wondering what made them create a YEAR type. What do they do for BC years? negative? Are there no paleontologist+DBAs?

akshayi1:
I was just wondering if there is going to be support for GTID instead of
position based binlog reference for maxwell.

osheroff:
GTID: yeah, at some point, but it's not on any short-term roadmap.  it'll be quite a bit of work down in open-replicator.

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
rhamnett: Need another rule to catch truncate of partitions similar to:

                SQL_BLACKLIST.add(Pattern.compile("^ALTER\\sTABLE.*TRUNCATE"));

For SQL such as ALTER TABLE foo TRUNCATE PARTITION p4;

osheroff: sorry for the lag @rhamnett.  Yeah, maxwell needs to be able to parse `PARTITION` stuff in general.  It's currently simply not supported, and is one of the last cleanups left in the sql parser section.

PR #70, opened 2015-06-19
cast column names to lower case

osheroff:
@zendesk/rules turns out table and db names have all sorts of arcane bullshit based on the system you're on, but column names are *always* case insensitive.


vanchi-zendesk:
:+1:

PR #72, opened 2015-07-07
log to stderr, not stdtout

osheroff:
@vanchi-zendesk @dadah89

### DIFF
--- a/src/main/resources/log4j2.xml
+++ b/src/main/resources/log4j2.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Configuration status="WARN">
   <Appenders>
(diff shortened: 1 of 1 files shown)


vanchi-zendesk:
:+1:

xmlking:
wonder why logging is switched to stderr?
I was trying to debug why maxwell  is crashing overnight. I am interested only in errors as lots of logs are produced in  stdout. 
```
nohup ./bin/maxwell >/dev/null 2>>./logs/error.log &
```

osheroff:
@xmlking it logs to STDERR so as not to step on the toes of the STDOUT producer.  You can change the behavior in log4j2.xml if you like.  Just change    `SYSTEM_ERR` to `SYSTEM_OUT` (I think).

PR #74, opened 2015-07-10
use a 50 MB buffer by default

osheroff:
we're overflowing some buffer with ultra-large row events.  I can't be certain that it's this one (can't seem to repro locally, faithfully), but this seems the most likely.

@zendesk/rules


osheroff:
ping

vanchi-zendesk:
:+1:

PR #76, opened 2015-07-13
bump open-replicator to 1.2.3

osheroff:
this allows OR to handle rows greater than 16mb.

@zendesk/rules


vanchi-zendesk:
build fail looks relevant.

vanchi-zendesk:
hm. maybe not. +1 on green

PR #77, opened 2015-07-27
allow a configurable number of saved maxwell schemas to be saved.

osheroff:
this will allow us to disable schema deletion at will. 

@zendesk/rules


vanchi-zendesk:
:+1:

PR #78, opened 2015-08-04
open-replicator 1.3.0-RC1

osheroff:
@zendesk/rules 

get checksum support going.  I'll release this as an RC candidate of maxwell, too


vanchi-zendesk:
:+1:

PR #84, opened 2015-08-31
ignore the BINARY flag in string column creation

osheroff:
a quick spelunk into the mysql source code shows that this flag only
affects the collation of the column.  We don't give even the slightest
care about that, so we'll parse and ignore it.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #85, opened 2015-09-01
Fix server_id bug

osheroff:
maxwell had a bug in which schemas were always saved with server_id == 1 -- meaning someone could switch server_ids out from underneath us without us knowing.  bad news!  this fixes.

note that this means if the master flips, we will end up re-capturing the schema.  that's ok for now.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
@@ -29,11 +29,18 @@ public class SchemaStore {
 	private BinlogPosition position;
 	private Long schema_id;
 
+	public Long getSchemaID() {
+		return schema_id;
+	}
+
 	static final Logger LOGGER = LoggerFactory.getLogger(SchemaStore.class);
 	private final PreparedStatement schemaInsert, databaseInsert, tableInsert;
 	private final String columnInsertSQL;
 
-	public SchemaStore(Connection connection) throws SQLException {
+	private final Long serverID;
+
+	public SchemaStore(Connection connection, Long serverID) throws SQLException {
+		this.serverID = serverID;
 		this.connection = connection;
 		this.schemaInsert = connection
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -204,12 +210,23 @@ public static SchemaStore restore(Connection connection,
...
 		PreparedStatement p;
-		ResultSet schemaRS = findSchema(targetPosition);
+		boolean shouldResave = false;
+		ResultSet schemaRS = findSchema(targetPosition, this.serverID);
+
+		if (schemaRS == null) {
+			// old versions of Maxwell had a bug where they set every server_id to 1.
+			// try to upgrade.
+
+			LOGGER.info("found schema with server_id == 1, re-saving...");
```
vanchi-zendesk: A new server will never have it as 1 right?

vanchi-zendesk: oh. nm. dumb question.

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
osheroff: it could.  that would trigger a re-save every time.  harumph.

vanchi-zendesk: er. I thought in that case schemaRS will not be null, no?

vanchi-zendesk: I mean the second time around?

osheroff: oh.  right!  yeah.  it'll have been accidentally correct.

PR #88, opened 2015-09-03
handle COMMIT QUERY_EVENTs

osheroff:
@zendesk/rules @bfabry @baxter @chausler 

managed to even get a repro of this behavior.  Note that this means events using myISAM will have `commit: true` but no `xid`.  I don't think this matters much for your use case.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellParser.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellParser.java
@@ -188,7 +188,7 @@ public class MaxwellParser {
 				case MySQLConstants.QUERY_EVENT:
 					QueryEvent qe = (QueryEvent) v4Event;
 
-					if ( qe.getSql().toString() == "COMMIT" ) {
+					if ( qe.getSql().toString().equals("COMMIT") ) {
 						// some storage engines(?) will output a "COMMIT" QUERY_EVENT instead of a XID_EVENT.
 						// not sure exactly how to trigger this.
 						if ( !list.isEmpty() )
--- a/src/main/java/com/zendesk/maxwell/MaxwellParser.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellParser.java
@@ -187,10 +187,17 @@ public class MaxwellParser {
 					break;
 				case MySQLConstants.QUERY_EVENT:
 					QueryEvent qe = (QueryEvent) v4Event;
-					// TODO: need to handle this case.
-					// The MySQL guys say some storage engines will output a "COMMIT" QUERY_EVENT at the
-					// end of the stream.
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -187,10 +187,17 @@ private MaxwellAbstractRowsEvent processRowsEvent(AbstractRowEvent e) throws Sch
...
-					// end of the stream.
-					throw new RuntimeException("Unhandled QueryEvent: " + qe);
+
+					if ( qe.getSql().toString().equals("COMMIT") ) {
+						// some storage engines(?) will output a "COMMIT" QUERY_EVENT instead of a XID_EVENT.
+						// not sure exactly how to trigger this.
+						if ( !list.isEmpty() )
+							list.getLast().setTXCommit();
+
+						return list;
```
vanchi-zendesk: so no xid in this case? Will that cause a problem?

vanchi-zendesk: A suggestion:

Instead of this xid stamping at the end of commit, can we
1. Add a maxwell_xid when we see transactions
2. At the end, associate mysql_xid with maxwell_xid when we see XID_EVENT. (So if someone wants mysql_xid, they can search for maxwell_xid which matches it)

This way, people still have ways of grouping up transactions, and also, we don't have to do this buffering thing.

vanchi-zendesk:
:+1:

bfabry:
will this mean some transactions including ticket updates never get a commit: true with their xid? if that's the case that will matter for us. we're trying to identify "interesting" and "not interesting" at the transaction level, so we need to know when the transaction finishes

vanchi-zendesk:
@bfabry From what I read, it's the other way around. You will always get commit: true (due to setTXCommit()) but, sometimes no xid. Ben should confirm..

osheroff:
@bfabry as far as I can tell *any* innodb transaction will have an xid and `commit: true`.  Only myISAM will elide the `xid` field (but it'll still have the `commit: true`).

In practice Zendesk uses very little myISAM (only global_uid) so most everything will have both features.

src/main/java/com/zendesk/maxwell/MaxwellParser.java
osheroff: hmm, is a very interesting idea, but how would we handle it if maxwell was stopped in the middle of a transaction?  I guess we'd have to store the position of the last started transaction as well as the last row output??

bfabry:
ok, I'll just ignore events with no xid, that'll work fine. cheers

chausler:
thanks @osheroff !!

osheroff:
released as v0.11.0-PRE2

PR #92, opened 2015-09-09
handle NUMERIC columns and NUMERIC/DECIMAL(X) columns

osheroff:
@zendesk/rules 
@yester354


src/main/antlr4/imports/column_definitions.g4
```diff
@@ -77,5 +77,5 @@ nullability: (NOT NULL | NULL);
 default_value: DEFAULT (literal | NULL | CURRENT_TIMESTAMP);
 length: '(' INTEGER_LITERAL ')';
 int_flags: ( UNSIGNED | ZEROFILL );
-decimal_length: '(' INTEGER_LITERAL ',' INTEGER_LITERAL ')';
+decimal_length: '(' INTEGER_LITERAL ( ',' INTEGER_LITERAL )? ')';
```
vanchi-zendesk: might want to add a test for the case (int,int) (I see the test for (int)

vanchi-zendesk:
:+1:

osheroff:
test added

Issue #93, opened 2015-09-10
Crash if column collation specified in CREATE statement


Komnomnomnom:
Maxwell crashes if collation is specified in a CREATE statement. 

MySQL version 5.1.73
Maxwell version 0.10.0

```
15:14:12,892 DEBUG SchemaPosition - Writing binlog position to maxwell.positions: BinlogPosition[mysql-binlog.000001:438782]
15:16:15,900 DEBUG SchemaChange - SQL_PARSE <- "CREATE TABLE `maxwell_collate_test` (
	`id` INT(10) NOT NULL AUTO_INCREMENT,
	`parent_id` INT(10) NULL DEFAULT NULL,
	`model` VARCHAR(255) NULL DEFAULT NULL COLLATE 'utf8_unicode_ci',
	`foreign_key` INT(10) NULL DEFAULT NULL,
	PRIMARY KEY (`id`),
	INDEX `parent` (`parent_id`),
	INDEX `foreign_key` (`foreign_key`),
	INDEX `model` (`model`)
)
COLLATE='utf8_unicode_ci'
ENGINE=InnoDB
ROW_FORMAT=COMPACT
AUTO_INCREMENT=371"
line 4:40 extraneous input 'COLLATE' expecting {',', ')', AUTO_INCREMENT, COLUMN_FORMAT, COMMENT, DEFAULT, NOT, NULL, ON, PRIMARY, STORAGE, UNIQUE}
line 4:48 missing '=' at ''utf8_unicode_ci''
15:16:15,940 ERROR MysqlParserListener - (create_specifications ( (create_specification (column_definition (name (id `id`)) (data_type (signed_type INT (length ( 10 )))) (column_options (nullability NOT NULL)) (column_options AUTO_INCREMENT))) , (create_specification (column_definition (name (id `parent_id`)) (data_type (signed_type INT (length ( 10 )))) (column_options (nullability NULL)) (column_options (default_value DEFAULT NULL)))) , (create_specification (column_definition (name (id `model`)) (data_type (string_type VARCHAR (length ( 255 )))) (column_options (nullability NULL)) (column_options (default_value DEFAULT NULL)))) <missing ')'>)
15:16:15,940 ERROR SchemaChange - Error parsing SQL: 'CREATE TABLE `maxwell_collate_test` (
	`id` INT(10) NOT NULL AUTO_INCREMENT,
	`parent_id` INT(10) NULL DEFAULT NULL,
[...]

PR #95, opened 2015-09-10
position independence for CHARACTER SET, COLLATION

osheroff:
as it turns out CHARACTER SET and COLLATION may appear in any order in
column options, not just before the generic options (like UNISIGNED and
ZEROFILL do for numeric columns).

fixes https://github.com/zendesk/maxwell/issues/93

@zendesk/rules @Komnomnomnom


Komnomnomnom:
Awesome thanks, I can verify that this fixes #93 for me :+1:

dadah89:
changes seem fine to me, :+1:

PR #96, opened 2015-09-11
*most* table creation options have optional "equals" signs.

osheroff:
@zendesk/rules


dadah89:
seems to works fine, :+1:

PR #99, opened 2015-09-11
try to improve test reliability with getEvents()

osheroff:
we now wait until the parser has consumed all of the events that we generate in the test suite.  Hopefully this fixes travis' reliability.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellParser.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellParser.java
@@ -202,7 +202,7 @@ public class MaxwellParser {
 					max_tries--;
 					continue;
 				} else {
-					System.out.println("maxwell didn't reach the position requested.");
+					LOGGER.error("maxwell didn't reach the position requested.");
 					return;
 				}
 			}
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -196,8 +193,24 @@ public MaxwellAbstractRowsEvent getEvent(boolean stopAtNextTableMap) throws Exce
...
+	public void getEvents(EventConsumer c, BinlogPosition endPosition) throws Exception {
+		int max_tries = 10;
+		while ( true ) {
+			MaxwellAbstractRowsEvent e = getEvent();
+			if (e == null) {
+				if (max_tries > 0) {
+					max_tries--;
+					continue;
+				} else {
+					System.out.println("maxwell didn't reach the position requested.");
```
dadah89: LOGGER.error or LOGGER.warn?

dadah89:
seems fine to me, and runs fine, :+1:

vanchi-zendesk:
:+1: Let's set travis branches to master so it does not build twice.

PR #100, opened 2015-09-16
minimal master-change detection

osheroff:
when we boot up we simply destroy any schemas and positions from other server-ids.  This will keep us from the worst of the master-flip scenarios where we go from master A -> master B -> master A.

@zendesk/rules


vanchi-zendesk:
:+1: (with the comment: probably logical deletion is better considering how much we are deleting right after a master flip)

PR #101, opened 2015-09-17
Maxwell must crash when its position thread stops.

osheroff:
In the event of a master cutover, Maxwell would previously amble
blithely along its way with a totally dead position-store thread until
it called SchemaPosition#setSync() and crashed.

Now we are much more aggressive at crashing when the position thread
crashes -- the most common case for ZD is that the master it is pointing
to is read-only.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #102, opened 2015-09-17
don't allow maxwell to start when binlog_row_image == minimal

osheroff:
Some mysterious person configured a mysterious mysql server with binlog_row_image==minimal which just breaks all sorts of goddamn things.
@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellMysqlStatus.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellMysqlStatus.java
@@ -15,13 +15,17 @@ public class MaxwellMysqlStatus {
 		return "SHOW VARIABLES LIKE '" + variableName + "'";
 	}
 
-	private String getVariableState(String variableName) throws SQLException, MaxwellCompatibilityError {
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/MaxwellMysqlStatus.java
```diff
@@ -15,13 +15,17 @@ private String sqlStatement(String variableName) {
 		return "SHOW VARIABLES LIKE '" + variableName + "'";
 	}
 
-	private String getVariableState(String variableName) throws SQLException, MaxwellCompatibilityError {
+	private String getVariableState(String variableName, boolean throwOnMissing) throws SQLException, MaxwellCompatibilityError {
```
vanchi-zendesk: can we capture this into ensureVariableState("read_only", "OFF") and getVariableState("binlog_format") ? Lots of code duplication here.

vanchi-zendesk:
:+1: other than that.

src/main/java/com/zendesk/maxwell/MaxwellMysqlStatus.java
osheroff: sure will dry

PR #103, opened 2015-09-17
argh, messed up precedence of SQL operators.

osheroff:
maxwell was still finding any and all schemas and using them. 

@zendesk/rules


vanchi-zendesk:
:+1: 

This release should be named "Quick, no one saw us"

PR #104, opened 2015-09-17
Revert "minimal master-change detection"

osheroff:
This reverts commit 82fcbd4eee4807ffeeda1671b63083ac10f21714.

Too much pressure on the replication chain really.  Will bring this back once we have soft deletion + slow scavenging.

@zendesk/rules


vanchi-zendesk:
+1

PR #107, opened 2015-09-18
Soft delete schemas

osheroff:
This should get us to where we are safely and quickly removing a downed master's old schemas.  Will follow up with a set of commits in which we scavenge and slowly delete the old rows from the DB.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
@@ -9,9 +9,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
 
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -357,9 +366,50 @@ private void deleteOldSchemas(Long currentSchemaId) throws SQLException {
...
 		while ( toDelete > 0 && schemaExists(toDelete) ) {
-			new SchemaStore(connection, serverID, toDelete).destroy();
+			new SchemaStore(connection, serverID, toDelete).delete();
 			toDelete--;
 		}
 	}
 
+	/*
+		for the time being, when we detect other schemas we will simply wipe them down.
+		in the future, this is our moment to pick up where the master left off.
```
vanchi-zendesk: might want to expand this note to mention we soft delete schemas, but we hard delete old positions. (Due to master flip back problem)

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
osheroff: well we only soft delete because of the replication pressure.  it's functionally the same as a hard delete

PR #111, opened 2015-09-25
exit the main maxwell thread cleanly

osheroff:
@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellParser.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellParser.java
@@ -126,14 +126,14 @@ public class MaxwellParser {
 			replicator.setBinlogPosition(event.getHeader().getNextPosition());
 		}
 
-		this.runState = RunState.STOPPED;
-
 		try {
 			this.binlogEventListener.stop();
 			this.replicator.stop(5, TimeUnit.SECONDS);
 		} catch ( Exception e ) {
 			LOGGER.error("Got exception while shutting down replicator: " + e);
 		}
+
+		this.runState = RunState.STOPPED;
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/MaxwellParser.java
```diff
@@ -78,18 +80,48 @@ public void start() throws Exception {
...
 
-	public void stop() throws Exception {
+	public void stop() {
+		// note: we use stderr in this function as LOGGER.err() oftentimes
+		// won't flush in time, and we lose the messages.
+
 		this.binlogEventListener.stop();
-		this.replicator.stop(5, TimeUnit.SECONDS);
+		try {
+			this.replicator.stop(5, TimeUnit.SECONDS);
```
vanchi-zendesk: this is a bit weird. Considering there is a chance that L129 (https://github.com/zendesk/maxwell/pull/111/files#diff-5adf358ee2ff00689a314b88b7f2786cR129 ) may simply restart it in the main thread. 

I suggest setting the shouldStop flag here, and maybe letting the main thread do this cleanup?

vanchi-zendesk: Or atleast check shouldStop before L129 ?

vanchi-zendesk:
Think the overall approach is fine. :+1: but for the comment above.

vanchi-zendesk:
ready for review again?

osheroff:
nope, working on a bigger refactor that should hopefully clean some of this crap up.  a bit.

osheroff:
Ok, this is ready for review now.  hopefully cleaner.

vanchi-zendesk:
:+1: nice

PR #114, opened 2015-09-28
4 bugfixes around shutdown

osheroff:
apparently all the shutdown code sucked.   

-  don't set runState to STOPPED until we shutdown the replicator too
-  terminate context thread *after* stopping main thread -- we were missing the last position update
- the parser thread should always set runState to STOPPED, even in the case of an exception
- fix inverted null check in context thread shutdown

@zendesk/rules 

NOTE: this branch is targeting ben/exit_softer.  merge manually.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -68,12 +68,12 @@ public class Maxwell {
 		Runtime.getRuntime().addShutdownHook(new Thread() {
 			@Override
 			public void run() {
-				context.terminate();
 				try {
 					p.stop();
 				} catch (TimeoutException e) {
 					System.err.println("Timed out trying to shutdown maxwell parser thread.");
 				}
+				context.terminate();
 			}
 		});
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -68,12 +68,12 @@ private void run(String[] argv) throws Exception {
 		Runtime.getRuntime().addShutdownHook(new Thread() {
 			@Override
 			public void run() {
-				context.terminate();
 				try {
 					p.stop();
 				} catch (TimeoutException e) {
 					System.err.println("Timed out trying to shutdown maxwell parser thread.");
 				}
+				context.terminate();
```
vanchi-zendesk: aha!. It did cross my mind if we want to do this first or last.

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -66,8 +66,9 @@ public void run() {
 			} catch (InterruptedException e) { }
 		}
 
-		if ( this.exception != null )
+		if ( exception == null ) {
```
vanchi-zendesk: hm. We see an exception on the main thread, we *don't* save the last position?

osheroff: no, this isn't the main thread's exception, it's the position thread's exception.

PR #115, opened 2015-09-29
bunch of fixes around replication positioning.

osheroff:
1. rename setInitialPosition() -> setPosition
2. only store our position into mysql if the event has the "commit"
   flag.  This will prevent us from ending up stopped in the middle of a
   transaction, which in some cases caused us to miss a table-map event,
   and thusly miss rows.
3. make sure the replicator thread is running, even inside
   getTransactionEvents() -- the replicator thread was getting
   disconnected from out from under us, leading to a process that would
   never make progress.
4. always setup the replicator's position after processing an event.
   ensure that if we need to restart the thread it'll always be in the
   right place.

I think I really need to refactor the MaxwellParser class.  It's huge.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -29,7 +29,7 @@ public class Maxwell {
 		SchemaStore store = new SchemaStore(connection, this.context.getServerID(), this.schema, pos);
 		store.save();
 
-		this.context.setInitialPosition(pos);
+		this.context.setPosition(pos);
 	}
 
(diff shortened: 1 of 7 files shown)


osheroff:
ping here, we need the `ensureReplicatorThread()` stuff pretty bad, maxwell in production keeps having stoppages.

dadah89:
So you moved the replicator position update within the loop, and added the check for commit producers push events, it seems fine to me. Tested locally and seems to run just fine, :+1:

vanchi-zendesk:
:+1: took a while to figure this one out. Definitely need a big refactor soon.

PR #116, opened 2015-10-01
Docker dev mode

osheroff:
move maxwell's root into /app

this lets Maxwell work properly with the zdi dev-mode scripts.
@zendesk/rules


vanchi-zendesk:
:+1:

PR #118, opened 2015-10-05
Refactor thread loop

osheroff:
both `MaxwellParser` and `SchemaPosition` were doing the same thing, running in a loop and stoppable from the outside.  Here we refactor them to use the same code. 

as with all refactor PRs, this one is pretty ugly, and probably best viewed with ?w=1

@zendesk/rules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,6 +82,11 @@
     	<artifactId>dbpool</artifactId>
     	<version>7.0-jdk7</version>
     </dependency>
+    <dependency>
+        <groupId>com.djdch.log4j</groupId>
+        <artifactId>log4j-staticshutdown</artifactId>
+        <version>1.1.0</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
@@ -4,18 +4,20 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -82,6 +82,11 @@
     	<artifactId>dbpool</artifactId>
     	<version>7.0-jdk7</version>
     </dependency>
+    <dependency>
+        <groupId>com.djdch.log4j</groupId>
+        <artifactId>log4j-staticshutdown</artifactId>
+        <version>1.1.0</version>
```
osheroff: this is a nice little jar that improves the behavior of log4j2 -- with this, you can actually log while the jvm is going down.

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -36,39 +38,46 @@ public SchemaPosition(ConnectionPool pool, Long serverID) {
...
 		thread.start();
 	}
 
-	public void stop() {
-		this.run.set(false);
-
+	public void stopLoop() throws TimeoutException {
+		this.requestStop();
 		thread.interrupt();
+		super.stopLoop();
```
osheroff: here, because I need to send `thread.interrupt()`, I had to override `stopLoop`.

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -36,39 +38,46 @@ public SchemaPosition(ConnectionPool pool, Long serverID) {
...
-		while ( run.get() && this.exception == null ) {
-			BinlogPosition newPosition = position.get();
+		try {
+			runLoop();
+		} catch ( Exception e ) {
+			// this code should never be hit.  There is, I suppose,
+			// a design flaw in my code, but at a certain point
+			// the whole inheritance + exception handling thing
+			// just started to drive me nuts.
+			LOGGER.error("Hit unexpected exception in SchemaPosition thread: " + e);
```
osheroff: sorry for the long rant.  I don't think I'm really much of a fan of checked exceptions.

vanchi-zendesk:
ok :+1:

PR #120, opened 2015-10-08
Bit Type Support Added

Tanmayshetty:
Add Bit Type Support While Maxwell Starts Reading Database Schema


osheroff:
hey, thanks for the contribution, and sorry about this, but I had been working on https://github.com/zendesk/maxwell/pull/121 for the last few days, which is not too different from your PR, but it converts the BIT columns back into raw numbers -- your solution would have needed more work, as JSON won't allow you to put any-old-string in it (has to be valid utf8).  

Thanks again, hope to see more contributions!

PR #121, opened 2015-10-08
add BIT column support

osheroff:
we export bit columns back as numeric values.  Note that we need to pick
up a new version of open-replicator, as the old one had a bug around
parsing longer BIT columns.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #125, opened 2015-10-08
Added database and table filter

csenki:
Added database and table filter


osheroff:
have a look at `org.apache.commons.lang.StringUtils.split`.  but otherwise, cool, keep going!

PR #128, opened 2015-10-12
add "SchemaScavenger" thread

osheroff:
this is a lazy-delete thread that checks for logically deleted entries in maxwell.schemas and physically DELETEs them, 500 rows at a time.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
@@ -1,0 +1,92 @@
+package com.zendesk.maxwell.schema;
+
+import com.zendesk.maxwell.RunLoopProcess;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import snaq.db.ConnectionPool;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SchemaScavenger extends RunLoopProcess implements Runnable {
+	static final Logger LOGGER = LoggerFactory.getLogger(SchemaStore.class);
+	private static final int MAX_ROWS_PER_SECOND = 500;
+	private final ConnectionPool connectionPool;
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
```diff
@@ -0,0 +1,92 @@
...
+	private final ConnectionPool connectionPool;
+
+	public SchemaScavenger(ConnectionPool pool) {
+		this.connectionPool = pool;
+	}
+
+	private List<Long> getDeletedSchemas() throws SQLException {
+		ArrayList<Long> list = new ArrayList<>();
+		try ( Connection connection = connectionPool.getConnection() ) {
+			ResultSet rs = connection.createStatement().executeQuery("select id from `maxwell`.`schemas` where deleted = 1");
```
vanchi-zendesk: A limit here is nice. I don't know how many schemas we accumulated.

src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
```diff
@@ -0,0 +1,92 @@
...
+		List<Long> schemaIds = getDeletedSchemas();
+
+		for ( Long id : schemaIds) {
+			boolean finishedDelete = false;
+
+			while ( toDelete > 0 && !finishedDelete ) {
+				for ( String tName : tables ) {
+					long deleted = deleteRows(id, tName, toDelete);
+
+					if (tName.equals("databases") && deleted < toDelete) {
```
vanchi-zendesk: this logic is needlessly complex, no? I had to think real hard about deleted < toDelete there. (It's the same as toDelete > 0 , by the way, if you hoist the toDelete -= deleted here.)  Maybe we don't have to be so accurate about our deletion speed? (maybe just sleep inside deleteRows() and make time configurable.

vanchi-zendesk: Note: By doing toDelete > 0, we can take deleteSchema out of the for loop and put it below it.

PR #129, opened 2015-10-13
Scavenge thread take 2

osheroff:
@vanchi-zendesk here's a different approach at the scavenge thread.  it's more readable I guess, but I had to expose some guts of RunLoopProcess as well as make it somewhat less accurate.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
@@ -13,7 +13,7 @@ import java.util.List;
 
 public class SchemaScavenger extends RunLoopProcess implements Runnable {
 	static final Logger LOGGER = LoggerFactory.getLogger(SchemaStore.class);
-	private static final int MAX_ROWS_PER_SECOND = 500;
+	private static final long MAX_ROWS_PER_SECOND = 500;
 	private final ConnectionPool connectionPool;
 
 	public SchemaScavenger(ConnectionPool pool) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		return list;
+	}
+
+	public void deleteSchema(Long id, Long maxRowsPerSecond) throws SQLException {
+		String[] tables = { "columns", "tables", "databases" };
+
+		try ( Connection connection = connectionPool.getConnection() ) {
+			connection.createStatement().execute("delete from `maxwell`.`schemas` where id = " + id);
```
vanchi-zendesk: i am a bit worried about deleting the id before we get all the rows. Can't this be done after?

src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
```diff
@@ -0,0 +1,83 @@
...
+					long nDeleted = connection.createStatement().executeUpdate(
+						"DELETE FROM maxwell." + tName +
+							" WHERE schema_id = " + id +
+							" LIMIT " + maxRowsPerSecond
+					);
+
+					if (isStopRequested())
+						return;
+
+					if (nDeleted == 0)
```
vanchi-zendesk: you could put it here, and then break.

src/main/java/com/zendesk/maxwell/schema/SchemaScavenger.java
osheroff: yeah.  that was definitely just a coding error when I did the rewrite of this.

osheroff:
@vanchi-zendesk updated

vanchi-zendesk:
:+1:

PR #132, opened 2015-10-23
Performance work

osheroff:
switching out to the jackson json libraries gets us from 70k rows per second to 100k rows per second on my laptop.  It adds a significant amount of ugliness as we end up hand-building the JSON, but this is *the* hot loop of maxwell and so probably worth it.  Suggestions on how to make the code prettier welcome.

@zendesk/rules @lileeyao


src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -196,6 +197,22 @@ public String toSQL() {
 	class RowMap extends HashMap<String, Object> {
 		private final HashMap<String, Object> data;
 
+		@Override
+		public Set<String> keySet() {
```
ggrossman: Is this in the hot loop?

Seems like `keySet` only has 4 possible values that are basically immutable

Constructing objects in Java creates garbage, increased GC pressure, etc.

Could we create four LinkedHashSets:

```
keySets[0] = database, table, type, ts
keySets[1] = database, table, type, ts, xid
keySets[2] = database, table, type, ts, commit
keySets[3] = database, table, type, ts, xid, commit
```
(expand to needed Java boilerplate)

Then

```
int xidOffset = this.containsKey("xid") ? 1 : 0;
inf commitOffset = (this.containsKey("commit") && (boolean) this.get("commit") == true)) ? 2 : 0;
return keySets[xidOffset + commitOffset];
```

ggrossman: Does this even need to be a `LinkedHashSet`? Why not just `ArrayList` or even dumb `String[]`?

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -275,21 +292,59 @@ public Object getData(String string) {
...
 		for ( RowMap map : jsonMaps() ) {
-			list.add(new MaxwellJSONObject(map).toString());
+			ByteArrayOutputStream b = new ByteArrayOutputStream();
+			JsonGenerator g = createJSONGenerator(b);
+
+			try {
+				rowMapToJSON(g, map);
+				g.flush();
+				list.add(b.toString());
+			} catch ( IOException e ) { /* never gonna happen. */ }
```
ggrossman: C'mon empty catch clauses is the "goto" of Java
If it ain't never gonna happen, then adding `logger.error` won't hurt anything :)

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -275,21 +292,59 @@ public Object getData(String string) {
...
+		g.writeEndObject();
+		g.writeEndObject();
 	}
 
 	public List<String> toJSONStrings() {
 		ArrayList<String> list = new ArrayList<>();
 
 		for ( RowMap map : jsonMaps() ) {
-			list.add(new MaxwellJSONObject(map).toString());
+			ByteArrayOutputStream b = new ByteArrayOutputStream();
```
ggrossman: Instead of new-ing up a ByteArrayOutputStream each time, how about `b.reset()`

osheroff: thanks for the push here.  I had this originally, but the `JsonGenerator` was adding spaces in an annoying way.  Found the work-around and this is worth another 10% or so of overall perf.

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
osheroff: yeah I simplified all of this immensely.

osheroff:
ping @ggrossman, if you're gonna start a code-review you gotta finish it :)

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -275,21 +279,75 @@ public Object getData(String string) {
...
+		try {
+			return factory.createGenerator(b);
+		} catch (IOException e) {
+			LOGGER.error("Caught exception while creating JSON generator: " + e);
 		}
-		return list;
+		return null;
+	}
+
+	static String[] keyOrder = {"database", "table", "type", "ts"};
```
ggrossman: Could throw in a `final` here

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -275,21 +279,75 @@ public Object getData(String string) {
...
 
 		for ( RowMap map : jsonMaps() ) {
-			list.add(new MaxwellJSONObject(map).toString());
+			try {
+				rowMapToJSON(g, map);
+				b.reset();
+				g.flush();
+				list.add(b.toString());
+			} catch ( IOException e ) {
+				LOGGER.error("Caught IOException while generating JSON: " + e);
```
ggrossman: Can do `LOGGER.error("Caught IOException while generating JSON: " + e, e);` instead of the `printStackTrace`

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -27,6 +30,7 @@
 // and ultimately outputs arrays of json objects representing each row.
 
 public abstract class MaxwellAbstractRowsEvent extends AbstractRowEvent {
+	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellAbstractRowsEvent.class);
```
ggrossman: `ALL CAPS LOGGER IS WEIRD`

src/main/java/com/zendesk/maxwell/producer/ProfilerProducer.java
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	@Override
+	public void push(MaxwellAbstractRowsEvent e) throws Exception {
+		if ( this.nullOutputStream == null ) {
+			this.nullOutputStream = new FileOutputStream(new File("/dev/null"));
+		}
+
+		if ( this.startTime == 0)
+			this.startTime = System.currentTimeMillis();
```
ggrossman: I hear `Clock.systemUTC().millis()` or `System.nanotime()` is the new hotness

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
osheroff: borrowed the pattern from open-replicator.  seemed sorta sane enough to me.

ggrossman:
:+1:

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
ggrossman: Indeed, they do do that.
I find it odd.. I've only ever seen that for constants in Java of the `public static final int READ_ONLY=0` variety, but good enough for me

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -275,21 +279,75 @@ public Object getData(String string) {
 		return list;
 	}
 
-	public List<JSONObject> toJSONObjects() {
-		ArrayList<JSONObject> list = new ArrayList<>();
+	static final JsonFactory factory = new JsonFactory();
```
ggrossman: Maybe call this `jsonFactory` to distinguish it from all the other factories that might spring into being at any moment, this being Java?

ggrossman: Can this be private?

PR #133, opened 2015-10-27
"replay" mode for maxwell -- intended to help performance testing

osheroff:
this adds a "--replay" and "--init-position" flag to maxwell that allows maxwell to re-tread already processed events, saving neither position nor new schemas to the database.  I've been using it to benchmark maxwell, but it could have some other marginal value.

@zendesk/rules


vanchi-zendesk:
:+1:

Issue #134, opened 2015-10-28
Maxwell database becomes large if you make lots of table creates


davidsheldon:
In our app, there are a selection of tables (one per day), and they get created every minute or so. We have several servers, and they each create these tables using "CREATE ... IF NOT EXISTS". Normally this is fine, as it will do nothing in mysql. These DDL lines still get replicated, and maxwell parses them, and adds the 'new' table to the schema each time it is seen.

<pre>
> select count(*) from tables;
+----------+
| count(*) |
+----------+
|   386694 |
+----------+

> select count(distinct name) from tables;
+----------------------+
| count(distinct name) |
+----------------------+
|                  824 |
+----------------------+

</pre>

This is after running Maxwell for less than an hour.

osheroff:
@davidsheldon let me know if you want to take on the work to handle no-op DDL statements properly, otherwise I knock it out.

osheroff:
fixed in https://github.com/zendesk/maxwell/releases/tag/v0.13.0

PR #135, opened 2015-10-28
Fix for #134 - duplicate tables being created.

davidsheldon:
If the CREATE ... IF NOT EXISTS ran on the upstream server, then we wont already know it. If it failed, or didn't create the table, then leave the schema alone.

### DIFF
--- a/src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
+++ b/src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
@@ -27,7 +27,7 @@ create_table:
       | create_like_tbl
     );
     
-create_table_preamble: CREATE TEMPORARY? TABLE (IF NOT EXISTS)? table_name;
+create_table_preamble: CREATE TEMPORARY? TABLE if_not_exists? table_name;
 create_specifications: '(' create_specification (',' create_specification)* ')'; 
 
 create_specification: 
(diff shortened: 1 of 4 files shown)


osheroff:
nice catch.  a couple of things:

- can you write a test that exposes this issue?  `DDLIntegrationTest` looks like about the right place to put an issue like this.
- there's an `ifNotExists` pattern here: https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/schema/ddl/DatabaseCreate.java#L19 which is a little bit safer (crash if asked to create an existing table, unless if-not-exists is true), can you add that to your patch?

thanks!
-ben

davidsheldon:
In DDLIntegration test I couldn't check for the exact case (table defined the same multiple times, as the schema diff functions don't seem to mind), but if we try to redefine the table to a different schema it does detect that both of them are in the schema and different.

davidsheldon:
Whilst this solves the bug of several tables being created in each schema, it still allocates a new schemaId, and writes a whole new schema to the database whenever we process one of these messages.

Would it be sensible to somehow check if the new schema is different, rather than just count the potential changes in https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java#L282

osheroff:
:+1: on the changes, looks like the schema-diff procedure needs to check the number of tables in each database, eh?  

Discussed the schema check in the other PR; I think it's good enough to say that if we get back the same object we started with, it's a no-op and we don't need to re-save the schema.

PR #137, opened 2015-10-29
Fix table name in error message for create table

davidsheldon:
sourceTable is always null, and doesn't have a toString method. This should be the name of the table that was looked up.


osheroff:
looks good :+1:

PR #138, opened 2015-10-29
Fix potential NPE in TableDrop

davidsheldon:
I suspect that this will not get through a working MySQL server, but if we've just tested the database for being null, calling a method on it is quite dangerous.

This code will still return "newSchema" if d == null && ifExists, but will use the existing return.

Incidentally, is there any difference between returning originalSchema or newSchema in this case? Should we return originalSchema if nothing's changed?


osheroff:
:+1: on the change. 

I think it'd be fine to return `originalSchema` if we don't end up making any changes.  The only thing I wanted to avoid was mutating the incoming schema (always make a copy if you make changes).  I think I wanted to do that because the plan was originally to do a schema-diff between the original schema and the new one.

Checking to see if they're the same object (and making the changes to enforce that) would also probably be fine.  There's a couple of examples of DDL that actually apply, but don't create changes: (`create table foo like bar; rename table foo to bar, bar to foo`), but I think they're corner cases enough to just say "if the `SchemaChange` returns the same object we started with, it's a no-op.

PR #141, opened 2015-10-29
Don't write new schema on noop DDL

davidsheldon:
When the DDL doesn't make any changes, return the same Schema object, and don't make database changes.

Related to #134.


osheroff:
nice.

osheroff:
:+1:

PR #143, opened 2015-10-31
WIP: productionize filter code

osheroff:
@csenski @joshcanfield @zendesk/rules

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -6,29 +6,55 @@ Maxwell can be configured by command line or a java "properties" file.
 
 ### Command line options
 
-option                          | description
-------------------------------- | -----------
---user USER                     | mysql username
---password PASSWORD             | mysql password
---host HOST                     | mysql host
---port PORT                     | mysql port
---producer stdout,kafka         | where shall we send these rows, sir?
---kafka.bootstrap.servers       | list of kafka nodes, listed as HOST:PORT[,HOST:PORT]
---kafka_topic                   | provide a topic for maxwell to write to. Default will be "maxwell".
+option                                        | description
+--------------------------------------------- | -----------
+--config FILE                                 | location of `config.properties` file
+--log_level DEBUG|INFO|WARN|ERROR             | log level
+--user USER                                   | mysql username
+--password PASSWORD                           | mysql password
+--host HOST                                   | mysql host
+--port PORT                                   | mysql port
+--producer stdout,kafka,file,profiler         | what type of producer to use.  default: stdout
+--output_file                                 | path of the file for the 'file' producer to write to
+--include_dbs PATTERN                         | only send updates from these databases
+--exclude_dbs PATTERN                         | ignore updates from these databases
+--include_tables PATTERN                      | only send updates from tables named like PATTERN
+--exclude_tables PATTERN                      | ignore updates from tables named like PATTERN
+--kafka.bootstrap.servers                     | list of kafka brokers, listed as HOST:PORT[,HOST:PORT].
+--kafka_topic                                 | provide a topic for maxwell to write to. Default will be "maxwell".
+--max_schemas                                 | how many old schemas maxwell should leave lying around in maxwell.schemas.  default: 5
+--init_position FILE:POSITION                 | ignore the information in maxwell.positions and start at the given binlog position.  Not available in config.properties.
+--replay                                      | enabled maxwell's read-only "replay" mode.  Not available in config.properties.
 
-### Configuration file options
+### Configuration file
 
-If maxwell finds the file `config.properties` in $PWD it will use it.
+If maxwell finds the file `config.properties` in $PWD it will use it.  Any
+command line options (except init_position and replay) may be given as
+"key=value" pairs.
 
-option                        | description
------------------------------ | -----------
-user=USER                     | mysql username
-password=PASSWORD             | mysql password
-host=HOST                     | mysql host
-port=PORT                     | mysql port
-producer=stdout,kafka,        | where shall we send these rows, sir?
-kafka.*=XXX                   | any options prefixed with 'kafka.' will be passed into the kafka producer library
-kafka_topic                   | provide a topic for maxwell to write to. Default will be "maxwell".
+Additionally, any configuration file options prefixed with 'kafka.' will be
+passed into the kafka producer library, after having 'kafka.' stripped off the
+front of the key.  So, for example if config.properties contains
+
+```
+kafka.batch.size=16384
+```
+
+Maxwell will send `batch.size=16384` to the kafka producer library.
+
+### row filters
+
+The options `include_dbs`, `exclude_dbs`, `include_tables`, and `exclude_tables` control whether
+Maxwell will send an update for a given row to its producer.  All the options take a single value PATTERN,
+which may either be a literal table/database name, given as `option=name`, or a regular expression,
+given as `option=/regex/`.  The options are evaluated as follows:
+
+1. only accept databases in `include_dbs`
+1. reject databases in `exclude_dbs`
+1. only accept tables in `include_tables`
+1. reject tables in `exclude_tables`
+
+So an example like `--include_dbs=/foo.*/ --exclude_tables=bar` will match `footy.baz` but not `footz.bar`.
 
 
 <script>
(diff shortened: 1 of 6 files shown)


src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -33,6 +32,28 @@ private void initFirstRun(Connection connection) throws SQLException, IOExceptio
 		this.context.setPosition(pos);
 	}
 
+	private void setupFilter(MaxwellFilter filter, MaxwellConfig config) {
+		if (!config.include_databases.isEmpty()) {
+			for ( String d : config.include_databases.split(",") )
```
joshcanfield: Is there value in supporting spaces in the string? 
`.split("\\s*,\\s*")`

e.g. `include_databases=one, two, three`

osheroff:
@joshcanfield yeah just pushed latest with regexs & space handling and general better hygene, take a look.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -1,57 +1,106 @@
...
-			return includeDatabases.contains(dbName);
+	private Pattern compile(String name) throws MaxwellInvalidFilterException {
+		if ( name.startsWith("/") ) {
+			if ( !name.endsWith("/") ) {
+				throw new MaxwellInvalidFilterException("Invalid regular expression: " + name);
+			}
+			return Pattern.compile(name.substring(1, name.length() - 1));
 		} else {
-			return !excludeDatabases.contains(dbName);
+			return Pattern.compile("^" + name.trim() + "$");
```
osheroff: here is where we're handling leading/trailing spaces.  I guess we should trim the string before we start, could get regexs with leading whitespace...

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -60,11 +60,18 @@ private OptionParser getOptionParser() {
...
 		parser.accepts( "kafka.bootstrap.servers", "at least one kafka server, formatted as HOST:PORT[,HOST:PORT]" ).withRequiredArg();
 		parser.accepts( "kafka_topic", "optionally provide a topic name to push to. default: maxwell").withOptionalArg();
+
 		parser.accepts( "max_schemas", "how many old schema definitions maxwell should keep around.  default: 5").withOptionalArg();
 		parser.accepts( "init_position", "initial binlog position, given as BINLOG_FILE:POSITION").withRequiredArg();
 		parser.accepts( "replay", "replay mode, don't store any information to the server");
+
+		parser.accepts( "includeDatabases", "include thes databases, formatted as includeDatabases=db1,db2").withOptionalArg();
+		parser.accepts( "excludeDatabases", "exclude thes databases, formatted as excludeDatabases=db1,db2").withOptionalArg();
+		parser.accepts( "includeTables", "include thes tables, formatted as includeTables=db1,db2").withOptionalArg();
```
joshcanfield: Typo - "thes" 
Should be updated with the regexp change too.

joshcanfield:
Cool. That will be nice for excluding our history/audit tables.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -1,57 +1,107 @@
...
-			return !excludeTables.contains(tblName);
+	public void addRowConstraint(String field, Integer value) {
+		rowFilter.put(field, value);
+	}
+
+	private boolean matchesIncludeExcludeList(List<Pattern> includeList, List<Pattern> excludeList, String name) {
+		if ( includeList.size() > 0 ) {
+			boolean found = false;
+			for ( Pattern p : includeList ) {
+				found = found || p.matcher(name).find();
```
vanchi-zendesk: break if found?

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -1,57 +1,107 @@
...
+			}
+			if ( !found )
+				return false;
+		}
+
+		for ( Pattern p : excludeList ) {
+			if ( p.matcher(name).find() )
+				return false;
```
vanchi-zendesk: Might want to add a note somewhere that excludes will override includes.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
vanchi-zendesk: er. Is this even possible? Cmd line arguments themselves are separated with spaces. The more natural assumption would be that db/table names might have spaces in them (if mysql supports that), and assume users will provide db names as --include_databases='db name','another db'

Will the option parser strip out the quotes?

joshcanfield: Table names can also include commas...

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
osheroff: did an update pass at the configuration documentation.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
osheroff: @vanchi-zendesk this code is here to handle the natural form: 

```--include_dbs="foo, bar"``` 

note that for spaces ```foo bar, baz` will work just fine.

Comment (osheroff) [inline src/main/java/com/zendesk/maxwell/MaxwellFilter.java]: @joshcanfield it's true what you say about commas, but as I've told @vanchi-zendesk in other PRs, I refuse to code around people who act like sociopaths and put commas in their table names. :)

**Comment by osheroff** [comment]
@vanchi-zendesk I think I addressed all the CR concerns.

**Comment by vanchi-zendesk** [comment]
:+1:

File: docs/docs/config.md [hunk 6]
```diff
@@ -6,29 +6,55 @@ Maxwell can be configured by command line or a java "properties" file.
...
+Maxwell will send an update for a given row to its producer.  All the options take a single value PATTERN,
+which may either be a literal table/database name, given as `option=name`, or a regular expression,
+given as `option=/regex/`.  The options are evaluated as follows:
+
+1. only accept databases in `include_dbs`
+1. reject databases in `exclude_dbs`
+1. only accept tables in `include_tables`
+1. reject tables in `exclude_tables`
+
+So an example like `--include_dbs=/foo.*/ --exclude_tables=bar` will match `footy.baz` but not `footz.bar`.
```
joshcanfield: In the example the exclude isn't a regexp, so it will only match exactly bar, not 'footz.bar', right?

osheroff: in the example the database is `footz` (which matches the include) but the table is `bar`, so it will be excluded.

joshcanfield: Haha.. I blame Monday. Thanks.

docs/docs/config.md
```diff
@@ -6,29 +6,55 @@ Maxwell can be configured by command line or a java "properties" file.
...
+Maxwell will send an update for a given row to its producer.  All the options take a single value PATTERN,
+which may either be a literal table/database name, given as `option=name`, or a regular expression,
+given as `option=/regex/`.  The options are evaluated as follows:
+
+1. only accept databases in `include_dbs` if non-empty
+1. reject databases in `exclude_dbs`
+1. only accept tables in `include_tables` if non-empty
+1. reject tables in `exclude_tables`
+
+So an example like `--include_dbs=/foo.*/ --exclude_tables=bar` will include `footy.zab` and exclude `footy.bar`
```
osheroff: rewrote the example a bit to be clearer, maybe?

The other thought I had was to try to boil it down to *only* include/exclude, so you could do:

```
 --include DB_NAME.*
 --include *.TBL_NAME
 --include /REGEX/ (where the regex is run against the fully qualified name.  you'd end up writing /.*\.prefix.*/ though )
```

whether that is really clearer though, I don't know.  It's a bit more flexible as you could tell maxwell to include tables `A.1` and `B.2` but *not* `A.2`, but feels slightly more cumbersome to work with.

I'm pretty sure Zendesk's filtering needs are handled entirely by `exclude_tables="/_.*_new/, /.*_stats$/"` but I could see the argument for more advanced cases.

PR #147, opened 2015-11-16
Change mv to cp, rm

seancaffery:
I was trying to build an image with 0.13.1 and kept getting several errors like:

```
mv: cannot create hard link './lib/antlr4-runtime-4.5.jar' to './docs/jacoco': Operation not permitted
```

It looks like [this is happening](http://superuser.com/questions/839662/mv-creates-a-hard-link-instead-of-move-or-rename-file-or-directory), but we couldn't find any evidence.

Changing to `cp` seemed to do the job.

@jeffreytheobald @osheroff

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN mkdir /app
 WORKDIR /app
 
 RUN curl -sLo - https://github.com/zendesk/maxwell/releases/download/v0.13.1/maxwell-0.13.1.tar.gz | tar zxvf -
-RUN mv maxwell-*/* .
+RUN cp -r maxwell-*/* . && rm -rf maxwell-*
 
 ADD REVISION /
 


jeffreytheobald:
:+1: same thing happened on my vanilla install, thanks for fixing this!

Dockerfile
```diff
@@ -6,7 +6,8 @@ RUN mkdir /app
 WORKDIR /app
 
 RUN curl -sLo - https://github.com/zendesk/maxwell/releases/download/v0.13.1/maxwell-0.13.1.tar.gz | tar zxvf -
-RUN mv maxwell-*/* .
+RUN cp -r maxwell-*/* .
+RUN rm -rf maxwell-*
```
kunday: It's better to do it in the same step, something like, 
```
RUN cp -r maxwell-*/* . && \
rm -rf maxwell-*
```
Doing in two steps means, the maxwell tar ball is left in the previous layer which from the rm i assume you are trying to not leave around.

seancaffery: Fair enough. Done.

kunday:
For folks who are interested in why this is happening, we encountered this a year or so back and we raised a ticket in docker repos, there are two possibilities https://github.com/docker/docker/issues/4570 and this https://github.com/docker/machine/issues/1876 , i believe this issue is more of the latter, which is a vboxfs limitation - https://www.virtualbox.org/ticket/818 

cc @jeffreytheobald

kunday:
LGTM :+1:

kunday:
@wushujames Yes, good catch, @seancaffery This oneliner does the trick,

```Docerfile
RUN curl -sLo - https://github.com/zendesk/maxwell/releases/download/v0.13.1/maxwell-0.13.1.tar.gz | tar --strip-components=1 -zxvf -
```

Thing of significance is the --strip-components and -zxvf part.

seancaffery:
```
tar (child): --strip-components=1: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
```
Because it is not saving to a file first?
I think it's fine to leave the tar around. My main concern is building a new image.

osheroff:
:+1: feel free to merge whenever you have something you like.

PR #148, opened 2015-11-17
apparently it's OK to name your primary keys.

osheroff:
The name is stored exactly nowhere, but the mysql parser is nothing if not liberal.

@bfabry  @zendesk/rules


bfabry:
:+1: cheers Ben

PR #151, opened 2015-11-19
Maxwell should capture `mysql` database schema

dadah89:
Changes on mysql db would throw exceptions.

Also will support crazy stuff like
```mysql
RENAME TABLE mysql.foo TO test.bar;
```

/cc @zendesk/rules


osheroff:
:+1: yay!

PR #152, opened 2015-11-21
Fix date/datetime tests.

wushujames:
I noticed that some of the methods in ColumnDefTest were missing the @Test annotation, which means they weren't running. Added the annotation, and fixed up the tests.

This is my first pull request ever, so let me know if I did anything wrong (either in the code, or in the Github workflow/pull request thing).


src/test/java/com/zendesk/maxwell/ColumnDefTest.java
```diff
@@ -181,17 +184,20 @@ public void TestDateTime() throws ParseException {
 
 		assertTrue(d.matchesMysqlType(MySQLConstants.TYPE_DATE));
 
-		Date date = new SimpleDateFormat().parse("1979-10-01 19:19:19");
-		assertThat(d.toSQL(date), is("'1979-10-01 19:19:19'"));
+        SimpleDateFormat simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        simpleDateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+        Date date = simpleDateFormat.parse("1979-10-01 19:19:19");
+        assertThat(d.toSQL(date), is("'1979-10-01 19:19:19'"));
```
osheroff: some whitespace issues here -- use tabs instead of spaces please.

src/test/java/com/zendesk/maxwell/ColumnDefTest.java
```diff
@@ -172,7 +175,7 @@ public void TestDate() {
 		assertTrue(d.matchesMysqlType(MySQLConstants.TYPE_DATE));
 
 		Date date = new GregorianCalendar(1979, 10, 1).getTime();
-		assertThat(d.toSQL(date), is("1979-10-01"));
+		assertThat(d.toSQL(date), is("''1979-11-01''"));
```
osheroff: this is odd -- is `toSQL` returning a string with two single-quote characters?

osheroff:
thanks, James!  BTW been meaning to talk to you about what you'd need as far as consuming Maxwell properly as a library...

src/test/java/com/zendesk/maxwell/ColumnDefTest.java
wushujames: Yes, it is returning with two single-quote characters, which I found weird, but I didn't want to change it without asking you about it first.

See https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java#L37

src/test/java/com/zendesk/maxwell/ColumnDefTest.java
wushujames: Okay, will do.

wushujames:
Yeah, I'm still gradually figuring out what I'll need to use Maxwell as a library. I was going to get-it-working, and then revisit all my changes to see if there were better ways to do things.

I have some work-in-progress if you want to take a look.  See https://github.com/wushujames/maxwell/tree/column-type as well as https://github.com/wushujames/maxwell/tree/wushujames/libraryize.

src/test/java/com/zendesk/maxwell/ColumnDefTest.java
osheroff: single single-quotes I think -- the premise is to get the value back to a string that can be put directly into SQL.

wushujames:
I fixed the whitespace issues, as well as the toSQL() for Dates to return single single-quotes. Still trying to figure out Git(hub), but it looks like my commits automatically updated the pull request too?

wushujames:
Travis CI failed, but I can't reproduce locally. It says it failed in "mvn test -B", but I ran that locally and it passed. Any ideas? Could it be an intermittent failure in Travis CI, or is there something more I could try locally?

osheroff:
yeah, you just push new changes to the branch, they'll come into the PR.

re: the tests, travis itself just often isn't all that reliable of a platform, especially when pushing it hard by doing things like downloading and bootstrapping a separate mysql server.  Anyway, I restarted the build, seems fine.  

merging...

osheroff:
:+1:

wushujames:
Thanks! Now that the branch is merged, does that mean I can delete the fix-datetime-tests branch?

osheroff:
yup.

PR #153, opened 2015-11-23
Move RowMap into its own file

osheroff:
part one of a bit of RowMap refactoring.  This is a small refactor towards buffering large transactions to disk -- I need something that can serialize well to disk, and RowMap is that.  

@zendesk/rules


dadah89:
:+1:

PR #155, opened 2015-11-23
move RowMap JSON conversion into RowMap class

osheroff:
Most this is just code movement, but we use  a couple of tricky ThreadLocal variables in order to save the cost of instantiating a new JsonGenerator for each row.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #157, opened 2015-11-23
fixes for replay mode dealing with the schema updates

osheroff:
in replay mode, we still need to process schema updates, we just don't
persist them back to the maxwell schema tables.

@zendesk/rules


vanchi-zendesk:
:+1:

Issue #158, opened 2015-11-23
Maxwell needs to parse SQL inside /*!  */ "comments"


jpeyper:
I have been seeing this error while evaluating Maxwell in a Fedora VM.
```
09:25:53,554 DEBUG SchemaChange - SQL_PARSE <- "CREATE DATABASE /*! IF NOT EXISTS */ error"
09:25:53,556 DEBUG SchemaChange - SQL_PARSE ->   (parse (statement (create_database CREATE DATABASE (name (id error)))) <EOF>)
com.zendesk.maxwell.schema.ddl.SchemaSyncError: Unexpectedly asked to create existing database error
	at com.zendesk.maxwell.schema.ddl.DatabaseCreate.apply(DatabaseCreate.java:25)
	at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:280)
	at com.zendesk.maxwell.MaxwellReplicator.getEvent(MaxwellReplicator.java:253)
	at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:89)
	at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:31)
	at com.zendesk.maxwell.Maxwell.run(Maxwell.java:89)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:95)
09:25:53,558 INFO  SchemaPosition - Storing final position: null
```
A similar error occurs at the table level.
```
10:24:04,086 DEBUG SchemaChange - SQL_PARSE <- "create table /*! if not exists */ `_work`.`cdctest3` (
  `id` int (10) unsigned not null auto_increment,
  primary key (`id`)
)"
10:24:04,090 DEBUG SchemaChange - SQL_PARSE ->   (parse (statement (create_table (create_table_preamble create table (table_name (db_name (name (id `_work`))) . (name (id `cdctest3`)))) (create_specifications ( (create_specification (column_definition (name (id `id`)) (data_type (signed_type int (length ( 10 )) (int_flags unsigned) (column_options (nullability not null)) (column_options auto_increment))))) , (create_specification (index_definition (index_type_pk primary key (index_column_list ( (index_columns (index_column (name (id `id`)))) ))))) )))) <EOF>)
com.zendesk.maxwell.schema.ddl.SchemaSyncError: Unexpectedly asked to create existing table cdctest3
[...]

PR #159, opened 2015-11-24
parse mysql-escaped style nested comments

osheroff:
Parses and honors SQL fragments inside `/*! ... */` blocks.  

Technically we really should put some parsing code in for the maximum mysql version that we support, 
but this will do until things fall down further...

Addresses #158

### DIFF
--- a/src/main/antlr4/imports/mysql_idents.g4
+++ b/src/main/antlr4/imports/mysql_idents.g4
@@ -9,7 +9,11 @@ string: (IDENT | STRING_LITERAL);
 integer: INTEGER_LITERAL;
 charset_name: (IDENT | STRING_LITERAL | QUOTED_IDENT);
 
-SQL_COMMENT: '/*' (.)*? '*/' -> skip;
+SQL_UPGRADE_COMMENT: '/*!' [0-9]* -> skip;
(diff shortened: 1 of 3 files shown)


src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -9,7 +9,11 @@ string: (IDENT | STRING_LITERAL);
 integer: INTEGER_LITERAL;
 charset_name: (IDENT | STRING_LITERAL | QUOTED_IDENT);
 
-SQL_COMMENT: '/*' (.)*? '*/' -> skip;
+SQL_UPGRADE_COMMENT: '/*!' [0-9]* -> skip;
+SQL_UPGRADE_ENDCOMMENT: '*/' -> skip;
+
+SQL_COMMENT: '/*' ~'!' (.)*? '*/' -> skip;
```
osheroff: the behavior of antlr's lexer is to match the longest token first, so we have to make sure that `SQL_COMMENT` dodges `/*!`, otherwise the greedy little thing that antlr is will gobble it.

vanchi-zendesk:
:+1:

PR #160, opened 2015-11-24
use instance variables instead of a HashMap for builtin json fields

osheroff:
this wins another ~10-15% of performance and some clarity.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/RowMap.java
@@ -6,15 +6,20 @@ import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -6,15 +6,20 @@
...
 import java.io.IOException;
-import java.io.InputStream;
-import java.math.BigDecimal;
-import java.math.BigInteger;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Set;
 
-public class RowMap extends HashMap<String, Object> {
+public class RowMap {
```
vanchi-zendesk: TODO: Maybe RowMap is not the best name for this anymore.

vanchi-zendesk:
:+1:

Issue #166, opened 2015-11-26
No ObjectCodec defined for the generator, ... (type passed java.sql.Time)


io41:
I'm experimenting a little with maxwell, following the getting started guide and appear to have hit an issue while processing an event generated by the following row:

```
mysql> select * from events where event_id = 5;
+----------+--------------+---------------------+----------------------------------+-----------+---------+----------------------------------+--------------+
| event_id | business_ref | when                | what_ref                         | what_type | action  | who_ref                          | deleted_name |
+----------+--------------+---------------------+----------------------------------+-----------+---------+----------------------------------+--------------+
|        5 |            1 | 2015-11-25 23:59:00 | 785d226973bf4eae9bfc903f0b9f2033 | Task      | updated | d4ed50fe84ba473fb2690b149e7f2933 | NULL         |
+----------+--------------+---------------------+----------------------------------+-----------+---------+----------------------------------+--------------+`
```

From the stack trace it appears that maxwell has trouble with the `when` column, which is of type `datetime`:

```
08:54:57,083 ERROR MaxwellReplicator - Got an unexpected row-event: WriteRowsEvent[header=BinlogEventV4HeaderImpl[timestamp=1448495940000,eventType=23,serverId=1,eventLength=290,nextPosition=2705491,flags=0,timestampOfReceipt=1448528097073],tableId=332,reserved=1,columnCount=10,usedColumns=1111111111,rows=[Row[columns=[560, 1, Wed Nov 25 23:59:00 UTC 2015, 785d226973bf4eae9bfc903f0b9f2033, Task, d4ed50fe84ba473fb2690b149e7f2933, webapp, null, BlobColumn[value=[B@7c417213], b4a678826df0408d869346ff67275627]]]]
java.lang.IllegalStateException: No ObjectCodec defined for the generator, can only serialize simple wrapper types (type passed java.sql.Time)
	at com.fasterxml.jackson.core.JsonGenerator._writeSimpleObject(JsonGenerator.java:1725)
	at com.fasterxml.jackson.core.base.GeneratorBase.writeObject(GeneratorBase.java:327)
	at com.fasterxml.jackson.core.JsonGenerator.writeObjectField(JsonGenerator.java:1415)
	at com.zendesk.maxwell.RowMap.toJSON(RowMap.java:91)
	at com.zendesk.maxwell.MaxwellAbstractRowsEvent.toJSONStrings(MaxwellAbstractRowsEvent.java:257)
[...]

PR #168, opened 2015-11-26
some Time fixes / improvements

osheroff:
- Support millisecond precisions in at least the schema (it not the output).
- Stop crashing when we see a TIME column

Addresses #166


osheroff:
@zendesk/rules

io41:
Awesome, fixed! Works locally for me, thank you

vanchi-zendesk:
:+1:

PR #170, opened 2015-11-30
Move JSON-key generation into RowMap

osheroff:
the next in a long series of refactors that will culminate in letting maxwell spill big transactions over to disk.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
@@ -202,13 +202,14 @@ public abstract class MaxwellAbstractRowsEvent extends AbstractRowEvent {
 		Object value;
 		for ( Iterator<Row> ri = filteredRows().iterator() ; ri.hasNext(); ) {
 			Row r = ri.next();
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -266,30 +267,14 @@ private JsonGenerator createJSONGenerator(ByteArrayOutputStream b) {
 	public List<String> getPKStrings() {
 		ArrayList<String> list = new ArrayList<>();
 
-		for ( Row r : filteredRows()) {
-			list.add(new MaxwelllPKJSONObject(table.getDatabase().getName(),
-										      table.getName(),
-										      getPKMapForRow(r)).toString());
+		for ( RowMap r : jsonMaps() ) {
```
vanchi-zendesk: did we lose filtering somehow?

vanchi-zendesk: no. nm.

vanchi-zendesk:
much better. +1

PR #172, opened 2015-11-30
Overflow large transactions to disk

osheroff:
And for the last act, we have the actual feature:  serialize RowMap rows to disk in a data structure that keeps N elements in memory and then starts spooling out to disk.

we run at about 1/10th the speed after spooling to disk, so making the memory buffer big "enough" is important.  If I knew how to reflect on java's max-heap size I'd make it some multiple of that, maybe.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
@@ -208,7 +208,6 @@ public abstract class MaxwellAbstractRowsEvent extends AbstractRowEvent {
 					getDatabase().getName(),
 					getTable().getName(),
 					getHeader().getTimestamp() / 1000,
-					getXid(),
 					table.getPKList(),
 					this.getNextBinlogPosition());
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
@@ -2,16 +2,7 @@ package com.zendesk.maxwell;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.StringWriter;
 import java.util.*;
-
-import com.fasterxml.jackson.core.JsonFactory;
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.code.or.common.glossary.column.BitColumn;
-import org.apache.commons.lang.StringUtils;
-import org.json.JSONObject;
 
 import com.google.code.or.binlog.BinlogEventV4Header;
 import com.google.code.or.binlog.impl.event.AbstractRowEvent;
@@ -34,9 +25,6 @@ public abstract class MaxwellAbstractRowsEvent extends AbstractRowEvent {
 	private final MaxwellFilter filter;
 	private final AbstractRowEvent event;
 
-	private Long xid;
-	private boolean txCommit; // whether this row ends the transaction
-
 	protected final Table table;
 	protected final Database database;
 
@@ -46,8 +34,6 @@ public abstract class MaxwellAbstractRowsEvent extends AbstractRowEvent {
 		this.header = e.getHeader();
 		this.table = table;
 		this.database = table.getDatabase();
-		this.txCommit = false;
-		this.xid = null;
 		this.filter = f;
 	}
 
@@ -94,23 +80,6 @@ public abstract class MaxwellAbstractRowsEvent extends AbstractRowEvent {
 		else
 			return null;
 	}
-
-	public Long getXid() {
-		return xid;
-	}
-
-	public void setXid(Long xid) {
-		this.xid = xid;
-	}
-
-	public void setTXCommit() {
-		this.txCommit = true;
-	}
-
-	public boolean isTXCommit() {
-		return txCommit;
-	}
-
 
 	@Override
 	public String toString() {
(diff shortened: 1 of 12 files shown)


wushujames:
Ben, what do you need this feature for?

vanchi-zendesk:
When Maxwell encounters a transaction, it keeps accumulating all the events until the XID is discovered and the transaction ends. Then it unspools all those events into Kafka. It turns out, we may hit memory limits by keeping so many events in memory. This PR, I think, keeps that list of events on memory + disk.

wushujames:
Can you unspool those events into Kafka as they come in? Then you wouldn't have to buffer into memory. Or, are you triggering of the XID event for some other purpose?

vanchi-zendesk:
In the binlog the XID goes last, and on Maxwell, we want the XID to go with every json event.

wushujames:
Ah, I see.

Is the goal to send the XID with every JSON event? Or is the goal to identify a set of JSON events as being part of a transaction?

If the primary goal is to mark JSON events as part of a transaction, then you could instead use the binlog position of the "start" of the transaction as your identifier, and add that to every JSON event. For example, if a transaction started at file "mysql.00004 position 12" you could add that to every JSON event.

osheroff:
@wushujames The original intention was to identify events as part of a transaction.  
We investigated a bunch of solutions; Maxwell could make up its own transaction ID (it sees the "BEGIN" message, but just doesn't know mysql's XID until the COMMIT event), or we could use the start-position like you said, or we could even just mark the first row BEGIN and the last row END.  In the end I'm being an annoying stickler for details here; somehow throwing away mysql's transaction ID in order to optimize a (rare-ish?) Maxwell path felt wrong.  

Ultimately I think I'd like the XID stuff to be a feature that you turn on, rather than something that's on by default -- so in those cases we could just skip the buffer entirely and push straight to the producer.

wushujames:
Makes sense to me. And I understand the desire not to lose data that MySQL was already giving you. Thanks for the explanations.

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
```diff
@@ -208,7 +208,6 @@ public String toSQL() {
 					getDatabase().getName(),
 					getTable().getName(),
 					getHeader().getTimestamp() / 1000,
-					getXid(),
```
vanchi-zendesk: where are we setting this ?

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -152,11 +150,11 @@ private MaxwellAbstractRowsEvent processRowsEvent(AbstractRowEvent e) throws Sch
 		return ew;
 	}
 
-	private LinkedList<MaxwellAbstractRowsEvent> getTransactionEvents() throws Exception {
+	private RowMapBuffer getTransactionEvents() throws Exception {
```
vanchi-zendesk: change name?

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -152,11 +150,11 @@ private MaxwellAbstractRowsEvent processRowsEvent(AbstractRowEvent e) throws Sch
...
 		BinlogEventV4 v4Event;
 		MaxwellAbstractRowsEvent event;
 
-		LinkedList<MaxwellAbstractRowsEvent> list = new LinkedList<>();
+		RowMapBuffer list = new RowMapBuffer(MAX_TX_ELEMENTS);
```
vanchi-zendesk: maybe list -> buffer.

src/main/java/com/zendesk/maxwell/MaxwellAbstractRowsEvent.java
vanchi-zendesk: may not need https://github.com/zendesk/maxwell/pull/172/files#diff-6af601f94c7136d68fd09c12665d16a1R215 because of setTxCommit below when we encounter XID

src/main/java/com/zendesk/maxwell/util/ListWithDiskBuffer.java
```diff
@@ -0,0 +1,76 @@
...
+			if ( elementsInFile == 0 )
+				LOGGER.debug("Overflowed in-memory buffer, spilling over into " + file);
+
+			os.writeObject(this.list.removeFirst());
+			elementsInFile++;
+		}
+	}
+
+	public boolean isEmpty() {
+		return (elementsInFile == 0) && this.list.size() == 0;
```
vanchi-zendesk: use this.size()

src/test/java/com/zendesk/maxwell/MaxwellIntegrationTest.java
```diff
@@ -128,14 +129,13 @@ public void testIncludeTable() throws Exception {
...
 		assertThat(list.size(), is(1));
 
-		e = list.get(0);
-		assertThat(e.getTable().getName(), is("minimal"));
+		assertThat(list.get(0).getTable(), is("minimal"));
 	}
 
 	@Test
 	public void testExcludeTable() throws Exception {
 		MaxwellAbstractRowsEvent e;
```
vanchi-zendesk: remove the unused 'e's

src/test/java/com/zendesk/maxwell/RowMapBufferTest.java
```diff
@@ -0,0 +1,26 @@
...
+public class RowMapBufferTest {
+	@Test
+	public void TestOverflowToDisk() throws Exception {
+		RowMapBuffer buffer = new RowMapBuffer(2);
+
+		RowMap r;
+		buffer.add(new RowMap("insert", "foo", "bar", 1L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1")));
+		buffer.add(new RowMap("insert", "foo", "bar", 2L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1")));
+		buffer.add(new RowMap("insert", "foo", "bar", 3L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1")));
+
```
vanchi-zendesk: maybe assert that we are only holding 2 in memory.

vanchi-zendesk:
:+1:

PR #176, opened 2015-12-04
Support MySQL online schema change DDL syntax, e.g. something like:

aquarapid:
ALTER TABLE foobar ADD foobar_column CHAR(2) DEFAULT NULL, ALGORITHM=INPLACE, LOCK=NONE;

Consult the MySQL 5.6 manual:

http://dev.mysql.com/doc/refman/5.6/en/alter-table.html

This is still missing some tests.


src/main/antlr4/imports/mysql_literal_tokens.g4
```diff
@@ -12,6 +12,7 @@ INT8: I N T '8';
 ACTION: A C T I O N;
 ADD: A D D;
 AFTER: A F T E R;
+ALGORITHM: A L G O R I T M;
```
osheroff: this file is generated by the script at src/main/antlr4/imports/generate_tokens.rb.  Can you add the tokens in there?  I'll update it to add a note so others don't fall into the same trap :)

osheroff:
thanks for the PR!  Can you add a test for the changes?  Something around the neighborhood of https://github.com/zendesk/maxwell/blob/d459b13f42d356f261cfa7af2a4ca864dc878a77/src/test/java/com/zendesk/maxwell/schema/ddl/DDLParserTest.java#L166 should be fine.

aquarapid:
For some reason the tests don't work for me locally;  the mysql onetimeserver starts up, but no tests actually run against it.  In addition, even if it did work, it uses 5.5.45;  that's not going to work for online schema change.

osheroff:
harumph, that's the second report of onetimeserver just hanging.  Mind sending me the output of `src/test/onetimeserver --mysql-version=5.5 --log-bin=master --binlog_format=row --innodb_flush_log_at_trx_commit=1 --server_id=123123 --debug`?

Meanwhile, I'll merge your PR.  Can write a test for it pretty quick.

btw, the whole test suite runs under both 5.5 and 5.6, with certain tests only running on 5.6.

Issue #178, opened 2015-12-04
Maxwel Crashing when Altering table with ADD INDEX  specified by Order


danfiza:
It looks like Maxwell is crashing when I ALTER a table with an ADD INDEX clause specified with an ORDER.

Ex:

ALTER TABLE `test` 
ADD COLUMN `test_col` ENUM('Y', 'N') NULL DEFAULT 'N',
ADD INDEX `test_col` (`test_col_index` ASC);

Results in an error stack like the below:
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: ASC
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:85)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:67)
	at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:273)
	at com.zendesk.maxwell.MaxwellReplicator.getEvent(MaxwellReplicator.java:253)
	at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:89)
[...]

PR #180, opened 2015-12-05
add ASC, DESC to index column definition

osheroff:
also allow for index comments.

addresses #178 
@zendesk/rules


dadah89:
that is new to me, but guess makes sense if you want to rely on index natural order being desc
:+1:

Issue #144, opened 2015-10-31
Emitting diffs in addition to the full row.


joshcanfield:
It would be useful to support a configuration that emits a diff record, only including the columns that have changed.

The old column data appears to be available in update events. I'm interested to know if this is something that has been considered? And if so is there a roadmap or design?

osheroff:
yup, this is on the list, and is actually a fairly simply change.  I had figured it would look like:

```
  { 
     "type": "update", 
     "data": { "id": 1, "txt": "abc", "foo": "bar"  },  // the complete row image 
     "old": { "txt": "old_text" } // only the rows that changed
  }
```

that way parsing an update stays the same as parsing an insert (`data` has the full row image), but you can access what the row was if you like. 

I don't have a particular time frame on implementing this, of course I'm super patches welcome, just ping this issue if you start on code so I don't duplicate effort.

osheroff:
I should probably pull together a ROADMAP.md at some point too.

joshcanfield:
This is great to hear, and thanks for being responsive!

I have a hackday tomorrow and I'm going to be integrating Maxwell into our system and I need column level change notification. If someone isn't already implementing this feature then I'll spend some time creating a branch.

osheroff:
great to hear, josh.  No one's working on it at the moment as far as I know.

jpeyper:
This is something I would be super keen for. 

Has anyone looked at this yet?

joshcanfield:
I wasn't able to get to it, but kristiankaufmann pull request looks good! :+1:

osheroff:
Fixed in 0.16.0

PR #184, opened 2015-12-09
Emit diffs for updateRowsEvent

kristiankaufmann:
Implements #144

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
@@ -88,25 +88,25 @@ public class MaxwellUpdateRowsEvent extends MaxwellAbstractRowsEvent {
 			Iterator<Column> befIter = before.getColumns().iterator();
 			Iterator<ColumnDef> defIter = table.getColumnList().iterator();
 			while ( aftIter.hasNext() && defIter.hasNext() && befIter.hasNext() ) {
-				Column a = aftIter.next();
-				ColumnDef d = defIter.next();
-				Column b = befIter.next();
+				Column afterColumn = aftIter.next();
+				ColumnDef columnDef = defIter.next();
+				Column beforeColumn = befIter.next();
 
 				avalue = valueForJson(afterColumn);
 
 				bvalue = valueForJson(beforeColumn);
 
 				if (avalue != null) {
-					avalue = d.asJSON(avalue);
-					if ( !avalue.equals(d.asJSON(bvalue)) ) {
-						bvalue = bvalue != null ? d.asJSON(bvalue) : bvalue;
-						rowMap.putOldData(d.getName(), bvalue);
+					avalue = columnDef.asJSON(avalue);
+					if ( !avalue.equals(columnDef.asJSON(bvalue)) ) {
+						bvalue = bvalue != null ? columnDef.asJSON(bvalue) : bvalue;
+						rowMap.putOldData(columnDef.getName(), bvalue);
 					}
 				} else if (bvalue != null) {
-					bvalue = d.asJSON(bvalue);
-					rowMap.putOldData(d.getName(), bvalue);
+					bvalue = columnDef.asJSON(bvalue);
+					rowMap.putOldData(columnDef.getName(), bvalue);
 				}
-				rowMap.putData(d.getName(), avalue);
+				rowMap.putData(columnDef.getName(), avalue);
 			}
 			list.add(rowMap);
 		}
--- a/src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
@@ -2,12 +2,16 @@ package com.zendesk.maxwell;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.*;
 
 import com.google.code.or.binlog.impl.event.UpdateRowsEvent;
 import com.google.code.or.binlog.impl.event.UpdateRowsEventV2;
+import com.google.code.or.common.glossary.Column;
 import com.google.code.or.common.glossary.Pair;
 import com.google.code.or.common.glossary.Row;
+import com.google.code.or.common.glossary.column.DatetimeColumn;
 import com.zendesk.maxwell.schema.Table;
+import com.zendesk.maxwell.schema.columndef.ColumnDef;
 
 public class MaxwellUpdateRowsEvent extends MaxwellAbstractRowsEvent {
 	private final UpdateRowsEvent event;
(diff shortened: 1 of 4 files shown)


osheroff:
sweet, thanks for the submission!  some comments will follow.

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,79 @@ public String sqlOperationString() {
...
+		performedBeforeAndAfterFilter = true;
+		return filteredRowsBeforeAndAfter;
+	}
+
+	@Override
+	public List<RowMap> jsonMaps() {
+		ArrayList<RowMap> list = new ArrayList<>();
+		Object avalue;
+		Object bvalue;
+		for (Iterator<Pair<Row>> p = filteredRowsBeforeAndAfter().iterator(); p.hasNext(); ) {
```
osheroff: I think you can just use:
```

File: src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java [hunk 2]
```diff
@@ -48,4 +52,79 @@ public String sqlOperationString() {
...
+			Row after = rowpair.getAfter();
+			Row before = rowpair.getBefore();
+
+			RowMap rowMap = new RowMap(
+					getType(),
+					getDatabase().getName(),
+					getTable().getName(),
+					getHeader().getTimestamp() / 1000,
+					table.getPKList(),
+					this.getNextBinlogPosition());
```
osheroff: this looks exactly the same as the call in the parent class.  Move this to a `buildRowMap()` method on `MaxwellAbstractRowEvent`

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,79 @@ public String sqlOperationString() {
...
+
+			Iterator<Column> aftIter = after.getColumns().iterator();
+			Iterator<Column> befIter = before.getColumns().iterator();
+			Iterator<ColumnDef> defIter = table.getColumnList().iterator();
+			while ( aftIter.hasNext() && defIter.hasNext() && befIter.hasNext() ) {
+				Column a = aftIter.next();
+				ColumnDef d = defIter.next();
+				Column b = befIter.next();
```
osheroff: there's a lot of variables going on in this function now, probably best to call them `beforeColumn`, `afterColumn`, `columnDef`

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,79 @@ public String sqlOperationString() {
...
+
+				if ( a instanceof DatetimeColumn) {
+					avalue = ((DatetimeColumn) a).getLongValue();
+				} else {
+					avalue = a.getValue();
+				}
```
osheroff: some DRY stuff here -- maybe move this check to a `valueForJSON(column, value)` function?

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,79 @@ public String sqlOperationString() {
...
+					bvalue = ((DatetimeColumn) b).getLongValue();
+				} else {
+					bvalue = b.getValue();
+				}
+
+				if (avalue != null) {
+					avalue = d.asJSON(avalue);
+					if ( !avalue.equals(d.asJSON(bvalue)) ) {
+						bvalue = bvalue != null ? d.asJSON(bvalue) : bvalue;
+						rowMap.putOldData(d.getName(), bvalue);
```
osheroff: this is a little unclear.  It looks as though you're putting `null` into the `old: {}` section, but so far the convention has been to simply elide NULL fields.  In any case, I think I might write this as spelling out all three cases:

```
   if ( avalue != null && bvalue != null ) { 
       // compare, maybe add to Old
   } else if ( avalue == null && bvalue != null ) {
       //  add to old
   } // else both are null, do nothing.
```

The other possibility to clean up code here is to simply make sure all the fields probably handle `.asJSON(null)`, and then strip null fields later.  But that's a bigger bit of lifting.

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -128,7 +130,31 @@ public String toJSON() throws IOException {
...
+					for (String s : stringList) {
+						g.writeString(s);
+					}
+					g.writeEndArray();
+				} else {
+					g.writeObjectField(key, data);
+				}
+			}
+			g.writeEndObject(); // end of 'old: { }'
+		}
```
osheroff: more DRY stuff here, put this into a `writeMapToJSON()` function

kristiankaufmann:
Ready for another look now

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,64 @@ public String sqlOperationString() {
...
+			Iterator<Column> befIter = before.getColumns().iterator();
+			Iterator<ColumnDef> defIter = table.getColumnList().iterator();
+			while ( aftIter.hasNext() && defIter.hasNext() && befIter.hasNext() ) {
+				Column afterColumn = aftIter.next();
+				ColumnDef columnDef = defIter.next();
+				Column beforeColumn = befIter.next();
+
+				avalue = valueForJson(afterColumn);
+
+				bvalue = valueForJson(beforeColumn);
```
osheroff: may as well bring `avalue` and `bvalue` into line -- `afterValue`, `beforeValue`

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,64 @@ public String sqlOperationString() {
...
+
+				if (avalue != null) {
+					avalue = columnDef.asJSON(avalue);
+					if ( !avalue.equals(columnDef.asJSON(bvalue)) ) {
```
osheroff: 1.  this code still feels a bit hard to read. 
2.  I think `columnDef.asJSON(bvalue)` will probably crash if `bvalue` is null -- you need the null check before this.

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -107,28 +136,12 @@ public String toJSON() throws IOException {
...
-				g.writeArrayFieldStart(key);
-				for ( String s : stringList )  {
-					g.writeString(s);
-				}
-				g.writeEndArray();
-			} else {
-				g.writeObjectField(key, data);
-			}
+		if ( !this.oldData.isEmpty()) {
+			writeMapToJSON("old", this.oldData, true);
```
osheroff: hmm.  Ok, yeah, I guess I buy that if the value changed from NULL -> something we should explicitly say `old: { old_field: null }`.

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
kristiankaufmann: How about this with the addition of a couple lines of comments?

```
afterValue = valueForJson(afterColumn) != null ? columnDef.asJSON(valueForJson(afterColumn)) : null;

beforeValue = valueForJson(beforeColumn) != null ? columnDef.asJSON(valueForJson(beforeColumn)) : null;
//Add beforeValue to rowMap.oldData if beforeValue is different from afterValue
if ( (afterValue != null && !afterValue.equals(beforeValue))
	|| (beforeValue != null && afterValue == null) ) {
    rowMap.putOldData(columnDef.getName(), beforeValue);
}
```

osheroff: nope, still too complex.  Really, just try the code as 3 if statements:

```
  if ( avalue != null && bvalue != null ) { 
    // run comparison
  } else if ( bvalue != null ) { 
     // avalue is null, bvalue is not -- cast bvalue to json and put in oldData
  }  else if ( avalue != null ) { 
     // bvalue is null, avalue is not -- put null in oldData
 } // otherwise, both are null, do nothing
```

kristiankaufmann:
any other changes?

Should this be an configurable for example.

src/main/java/com/zendesk/maxwell/MaxwellUpdateRowsEvent.java
```diff
@@ -48,4 +52,71 @@ public String sqlOperationString() {
...
+				if ( afterValue != null ) {
+					afterValue = columnDef.asJSON(afterValue);
+				}
+
+				beforeValue = valueForJson(beforeColumn);
+				if ( beforeValue != null) {
+					beforeValue = columnDef.asJSON(beforeValue);
+				}
+
+				if ( afterValue != null && beforeValue != null ) {
```
joshcanfield: https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object)

I think you can replace this all with the null-safe Objects compare.
```java
if ( !Objects.equals(beforeValue, afterValue) ) {
  rowMap.putOldData(columnDef.getName(), beforeValue);
}
```

kristiankaufmann: Great! done.

joshcanfield:
Looking forward to getting it up and running!

osheroff:
@kristiankaufmann nope this looks good.  I need to do a larger pass at enabling/disabling features at some point, but I think most people want this on by default.  Thanks much!  I will test locally in a bit and then merge.

osheroff:
:+1:

osheroff:
released as 0.16.10

wushujames:
If I'm reading this correctly, the only impact this has on the output data format is that there is a new key called "old"?

PR #185, opened 2015-12-09
handle VIEW ddl better

osheroff:
It's hard to blanklist `CREATE VIEW` and `ALTER VIEW` statements with a regex, so we run them through ANTLR, which parses them enough to recognize them as view DDL, and then ignores them. 

@zendesk/rules


dadah89:
:+1: , moving the EVENT change to another PR

src/main/antlr4/imports/mysql_view.g4
```diff
@@ -0,0 +1,26 @@
...
+
+create_view:
+  CREATE (OR REPLACE)? view_options* VIEW name;
+
+view_options:
+    ALGORITHM '=' (UNDEFINED | MERGE | TEMPTABLE)
+  | DEFINER '=' (user | CURRENT_USER)
+  | SQL SECURITY ( DEFINER | INVOKER );
+
+
```
dadah89: :space_invader:

Issue #186, opened 2015-12-10
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: SELECT


jpeyper:
I had this error occur today.

```
line 4:0 no viable alternative at input 'SELECT'
line 5:3 token recognition error at: '*\r'
10:58:19,973 ERROR MysqlParserListener - (create_table (create_table_preamble CREATE TEMPORARY TABLE (table_name (name (id 172898_16841_transmem)))) SELECT t . FROM map . transmem AS t WHERE t . deleted = 'N' AND IFNULL ( t . trnnumrev , 0 ) = 0 AND t . locked = 'Y' AND t . lockedby = 'AUTOSTL' AND t . allocated = 'N' AND t . special = 'N' AND t . invnum_d = 172898 AND t . imnum = 16841)
10:58:19,974 ERROR SchemaChange - Error parsing SQL: 'CREATE
TEMPORARY
TABLE 172898_16841_transmem
SELECT
 t.*
FROM map.transmem AS t
WHERE t.deleted = 'N'
AND IFNULL(t.trnnumrev, 0) = 0
AND t.locked = 'Y'
AND t.lockedby = 'AUTOSTL'
AND t.allocated = 'N'
AND t.special = 'N'
AND t.invnum_d = 172898
AND t.imnum = 16841'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: SELECT
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:85)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:67)
	at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:274)
	at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:254)
	at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:87)
[...]

PR #187, opened 2015-12-10
blacklist CREATE TEMPORARY TABLE

osheroff:
that wasn't supposed to replicate.  but ok.
@zendesk/rules 

addresses #186


dadah89:
:+1:

PR #190, opened 2015-12-10
Add 'commit' to Kafka message key

seancaffery:
Allows filtering from the message key based on a table list or if there is a commit.

During load spikes, our consumer needs to discard a large number of messages. We will be able to reduce the logic required to discard messages (and hopefully speed it up) if we can whitelist a couple of tables along with 'commit' events.

Happy to take suggestions if this is a silly way to do it :)

@osheroff @chausler


chausler:
:+1: this will make pacman happy

osheroff:
unfortunately this will do bad things: the intention of the kafka key is to be unique per row (to allow for topic compaction), if we do this we'll end up with at least two (random) copies of each row.

PR #192, opened 2015-12-11
take another try at preventing memory issues.

osheroff:
ObjectInputStream.readObject and ObjectOutputStream.writeObject do a
semi-pathological thing where they retain references to every object
that you've written into them (for efficiency purposes).  This means
that my whole premise about using them to flush ram out to disk was
flawed.

I also threw some buffering onto the streams as that improves perf a bit.

Really I'm finally about to admit that trying to preserve mysql's transaction ID is flawed.  

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
@@ -155,11 +155,11 @@ public class MaxwellReplicator extends RunLoopProcess {
 
 		RowMapBuffer buffer = new RowMapBuffer(MAX_TX_ELEMENTS);
 
-		// currently to satisfy the test interface, the contract is to return null
-		// if the queue is empty.  should probably just replace this with an optional timeout...
-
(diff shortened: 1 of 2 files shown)


wushujames:
Another option is to read the binlog, ignoring data changes and just looking for the transaction id, and remember it when you find it. When you are done, rewind back in the binlogs to the start of the transaction and then use the remembered transaction to apply to the data change events.

It's kind of horrible and you'll be using double the network in exchange for reduced memory. But maybe you only do that for large transactions?

Anyway, just throwing that idea out there.

osheroff:
@dadah89 I moved the initialization of `is` to somewhere that makes a bit more sense.

osheroff:
@wushujames Yeah, that's a good option, it's problematic from a code-standpoint though; currently the open-replicator is decoupled fairly heavily from the rest of the stack; trying to intertwine them here would be hard.

dadah89:
:+1:

PR #194, opened 2015-12-11
[Jdk7] try-with-resource close resource and Files.readAllLines make SQL String

darionyaphet:
Simple Improment : 

1. Statement and ResultSet should close every time after used . 
2. FIles.readAllLines to collect file content.

thank you  :)


osheroff:
hi darion, 
I've poked around the documentation, and I don't see any reason to wrap `prepareStatement()` calls in a `try{}` statement.  What is this code intended to do?  generally that syntax indicates that there's some hidden action at the end of the `try {}` block that must happen -- for instance, if you have a pool of database connections, you might reserve one in the `try()` statement and then check it back in at the end.

darionyaphet:
@osheroff 

some resource after used should be close , before JDK7 it's write as this structure:

```
        try{
            resource.doSomeThing();
        }finally{
            resource.close();
        }
```

Since JDK 7, a new "try-with-resources" is introduced. When the `try` block is end, it will close or release the resource automatically.

```
The try-with-resources statement is a try statement that declares one or more resources. 

A resource is an object that must be closed after the program is finished with it. 

The try-with-resources statement ensures that each resource is closed at the end of the statement. 

Any object that implements java.lang.AutoCloseable, which includes all objects which implement java.io.Closeable, can be used as a resource.
```

Here are Oracle's Java Doc [The try-with-resources Statement](http://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

osheroff:
@darionyaphet I know the patttern, but I don't think `prepareStatement()` creates resources that need to be closed.

darionyaphet:
`PreparedStatement` is not necessary . It's  a precompiled SQL statement for execute SQL many times . Furthermore  , it's same with `Statement` . But in the method just once execute SQL .

Also I found a discuss on stackoverflow [When should a java PreparedStatement be closed?](http://stackoverflow.com/questions/6353046/when-should-a-java-preparedstatement-be-closed)  It's say you should close it when DBMS operation completed .

osheroff:
Yeah I read that thread on stack overflow.  I'm not convinced.  I don't think prepared statements or result sets need to be closed, they can just get garbage collected like anything.

darionyaphet:
yep they will recovery by GC automatic but I don't know it will make some strange case which without close .  So  I always close them by myself .

osheroff:
nah, that's just not how it works.  There's nothing server-side when dealing with a prepared statement.

darionyaphet:
I don't think recovery resource dependence on GC is a good idea

osheroff:
That's correct, *for objects that have external resources.*. So yes, you tend to want to close file handles and server connections rather than letting gc handle it.

But prepared staments don't have any such resources.  So there's nothing to clean up.

darionyaphet:
thank you for you relay 

maybe I should know what happened on `PreparedStatement ` and `ResultSet` close :)

darionyaphet:
I check up `mysql-connector-j` and found close() just remove it from `Map<Statement,Statement>`

In StatementImpl.realClose

```
if (this.connection != null) {
  if (!this.connection.getDontTrackOpenResources()) {
    this.connection.unregisterStatement(this);
  }
}
```

darionyaphet:
@osheroff   I close this PR and will re-open it if necessary

PR #197, opened 2015-12-11
support CREATE TABLE SELECT * FROM `foo`

osheroff:
with row-based replication this gets transformed to:

BEGIN
CREATE table t ( ... )
ROWS
COMMIT

so we have to support CREATE TABLE ddl inside a transaction.  Note that
this means we could process the CREATE TABLE, flush the binlog position,
crash, and then restart inside a transaction, missing the BEGIN event.
For this reason we downgrade the started-replication-outside-transaction
case to a warning.

@zendesk/rules


dadah89:
:+1:

wushujames:
Is it actually possible to reconnect in the middle of a transaction? From what I've seen, I didn't think that was possible. See https://github.com/shyiko/mysql-binlog-connector-java/issues/53 that I filed against mysql-binlog-connector-java.

PR #199, opened 2015-12-13
Support binlog_row_image=MINIMAL

osheroff:
The bulk of the work in this changes was in creating a unified iterator that can 
traverse Columns, ColumnDefinitions, and the "which columns are actually present" bitmap in 
one go.

@zendesk/rules

### DIFF
--- a/docs/docs/compat.md
+++ b/docs/docs/compat.md
@@ -9,7 +9,17 @@
 ### Unsupported
 
 - Mysql 5.7 is untested with Maxwell, and in particular GTID replication is unsupported as of yet.
-- `binlog_row_image=MINIMAL` is not supported and will break Maxwell in a variety of amusing ways.
+
+### binlog_row_image=MINIMAL
+
+As of 0.17.0, Maxwell supports binlog_row_image=MINIMAL, but it may not be what you want.  It will differ
+from normal Maxwell operation in that:
+
+- INSERT statements will no longer output a column's default value
+- UPDATE statements will be incomplete; Maxwell outputs as much of the row as given in the binlogs,
+  but `data` will only include what is needed to perform the update (generally, id columns and changed columns).
+  The `old` section may or may not be included, depending on the nature of the update.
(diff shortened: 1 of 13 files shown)


docs/docs/compat.md
```diff
@@ -9,7 +9,17 @@
 ### Unsupported
 
 - Mysql 5.7 is untested with Maxwell, and in particular GTID replication is unsupported as of yet.
-- `binlog_row_image=MINIMAL` is not supported and will break Maxwell in a variety of amusing ways.
+
+### binlog_row_image=MINIMAL
```
wushujames: How does binlog_row_image=MINIMAL work with log compaction? Or, does it not work with log compaction? I guess log compaction isn't technically a Maxwell concern, it's a Kafka concern, so it's up to the user if they care about it.

src/test/resources/sql/json/test_minimal
```diff
@@ -1,36 +1,24 @@
...
+insert into test_table set a='avalue2', b='bvalue2'
 
-->  {"database": "test_db_1", "table": "test_table", "type": "insert", "data": { "id": 2, "textcol": "test_col_2"} }
+->  {"database": "test", "table": "test_table", "type": "insert", "data": {"id": 2, "a": "avalue2", "b": "bvalue2"} }
 
-alter table test_table add column `datecol` datetime NULL default null AFTER id
+update test_table set a='avalue3' where id = 2
 
-insert into test_table set textcol='test_col_2', datecol='1979-10-01 00:00:00'
+->  { "database": "test", "table": "test_table", "type": "update", "data": {"id": 2, "a": "avalue3"} }
```
vanchi-zendesk: suggested test:
```
update test_table set a='avalue3' where a='avalue2'
```

vanchi-zendesk: nm.

vanchi-zendesk:
:+1:

Issue #200, opened 2015-12-14
Parse table names that look like floating point numbers


jpeyper:
Parse error showed up in my log
```
2015-12-14T13:40:57,653 ERROR com.zendesk.maxwell.schema.ddl.MysqlParserListener: (alter_specification .20151214 _TSP_16819__CGPos_0_C7D1A82D5B6243259A60D3CAD54A4B61 ADD KEY `TRNNUM` ( `trnnum`)
2015-12-14T13:40:57,653 ERROR com.zendesk.maxwell.schema.ddl.SchemaChange: Error parsing SQL: 'ALTER TABLE cgt.20151214_TSP_16819__CGPos_0_C7D1A82D5B6243259A60D3CAD54A4B61	asis-uat3.hq.local
```

osheroff:
@jpeyper, was that the full error message?  seems truncated.

osheroff:
Also, your environment appears to have a knack to find *all* the bugs.

osheroff:
oh, I think I see.  I think it hates underscores.

osheroff:
wait that makes no sense.

osheroff:
ok.  the numeric table name, after the dot.

jpeyper:
haha, yes, we have a lot of people here running a lot of things against these environments. if there ever was going to be somewhere that would find the obscure edge cases, we are it.

Umm, the log may have been truncated, I can hunt down the proper log if you need it.

osheroff:
Nah, I sussed it out. Parser was confused about what was a floating point and what was a table name.

jpeyper:
Is this going into a release soon?

Thanks for all the bug fixes for us :)

osheroff:
@jeyper, yeah!  I wanted to get it out there with the MINIMAL binlog format fixes, but those need to go through a code-review cycle here.  I'll get those CR'ed, merged and will release a new version tomorrow

osheroff:
Fixed in https://github.com/zendesk/maxwell/releases/tag/v0.16.2

PR #201, opened 2015-12-14
fix parser's ability to handle DATABASE.12345_TBL

osheroff:
the problem was that the string '.12345restoftablename' would match
FLOAT_LITERAL + IDENT.  The fix is to shorten the tokens, so that the
'.' character is no longer part of a longer token, and must always match
independently.  We then reconstruct the floating point parsing rules out
of INTEGER_LITERAL tokens.

addresses lucky issue #200.  
@zendesk/rules


dadah89:
:+1:

PR #202, opened 2015-12-14
use ObjectOutputStream#reset to constrain memory

osheroff:
writeUnshared was (a) terribly slow, and (b) didn't actually prevent
ObjectOutputStream from holding onto references.  .reset() appears to
actually do what we want, which is to release references to the RowMap.

@zendesk/rules @dadah89

### DIFF
--- a/src/main/java/com/zendesk/maxwell/util/ListWithDiskBuffer.java
+++ b/src/main/java/com/zendesk/maxwell/util/ListWithDiskBuffer.java
@@ -37,9 +37,12 @@ public class ListWithDiskBuffer<T extends Serializable> {
 			if ( elementsInFile == 0 )
 				LOGGER.debug("Overflowed in-memory buffer, spilling over into " + file);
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/util/ListWithDiskBuffer.java
```diff
@@ -37,9 +37,12 @@ public void add(T element) throws IOException {
 			if ( elementsInFile == 0 )
 				LOGGER.debug("Overflowed in-memory buffer, spilling over into " + file);
 
-			os.writeUnshared(this.list.removeFirst());
+			os.writeObject(this.list.removeFirst());
 
 			elementsInFile++;
+
+			if ( elementsInFile % maxInMemoryElements == 0 )
```
dadah89: this is faster than `elementsInFile == maxInMemoryElements`?

dadah89:
:+1:

Issue #204, opened 2015-12-16
Zendesk crashes if there is period in column in Create Table statment


tomarkuntal:
When executing following SQL Statement, maxwell crashes if there is dot (period) in column names. If I modify to test.CreatedBy to test_CreatedBy everything works fine. I understand having period in column name is not the right way but that is the convention that is being used right now.

SQL statement:

CREATE TABLE `person_5654e0fa181c13df04000020` (
  `_id` varchar(45) NOT NULL,
  `test.CreatedBy` varchar(45) DEFAULT NULL,
  `firstName` varchar(255) DEFAULT NULL,
  `lastName` varchar(255) DEFAULT NULL,
  `startDate` datetime DEFAULT NULL,
  `isActive` int(1) DEFAULT NULL,
  PRIMARY KEY (`_id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;

Error in maxwell:

03:00:28,073 DEBUG SchemaChange - SQL_PARSE <- "CREATE TABLE `person_5654e0fa181c13df04000020` (
  `_id` varchar(45) NOT NULL,
  `test.CreatedBy` varchar(45) DEFAULT NULL,
  `firstName` varchar(255) DEFAULT NULL,
  `lastName` varchar(255) DEFAULT NULL,
  `startDate` datetime DEFAULT NULL,
  `isActive` int(1) DEFAULT NULL,
  PRIMARY KEY (`_id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1"
03:00:28,084 ERROR MysqlParserListener - (data_type ` varchar(45) DEFAULT NULL
03:00:28,085 ERROR SchemaChange - Error parsing SQL: 'CREATE TABLE `person_5654e0fa181c13df04000020` (
  `_id` varchar(45) NOT NULL,
  `test.CreatedBy` varchar(45) DEFAULT NULL,
  `firstName` varchar(255) DEFAULT NULL,
  `lastName` varchar(255) DEFAULT NULL,
  `startDate` datetime DEFAULT NULL,
  `isActive` int(1) DEFAULT NULL,
  PRIMARY KEY (`_id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: ` varchar(45) DEFAULT NULL,
  `
[...]

PR #205, opened 2015-12-16
be more permissive about what's inside a quote character

osheroff:
really, it's anything these days.  I must have followed "Before MySQL
5.1.6, database and table names cannot contain /, \, ., or characters
that are not permitted in file names."

dumb me for targeting an older mysql release.

@zendesk/rules 

addresses #204


dadah89:
:+1:

PR #211, opened 2015-12-18
mysql-test, part 1

osheroff:
Further fixes in the parser.

- support `NOW()` as an alias for CURRENT_TIMESTAMP
- support `long varbinary` and others as column type aliases.

this could be a sinkhole
@zendesk/rules


vanchi-zendesk:
:+1:

PR #212, opened 2015-12-18
mysql-test, part 2

osheroff:
DON'T CLICK MERGE.  merge by hand into master instead.

- `DOUBLE PRECISION` as alias for `DOUBLE`
- `ASCII` as alias for `character set latin1`
- `SIGNED` integer flag
- BINARY is a valid character set name
- handle b'101', x'fkljlkj' style strings
- CREATE AGGREGATE FUNCTION
- unquoted, higher bit utf8
@zendesk/rules


src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -38,13 +38,13 @@ SQL_COMMENT: '/*' ~'!' (.)*? '*/' -> skip;
 
 SQL_LINE_COMMENT: ('#' | '--') (~'\n')* ('\n' | EOF) -> skip;
 
-STRING_LITERAL: TICK ('\\\'' | '\'\'' | ~('\''))* TICK;
+STRING_LITERAL: [a-zA-Z]? TICK ('\\\'' | '\'\'' | ~('\''))* TICK;
```
vanchi-zendesk: isn't this a bit too relaxed? What else can come here other than b and x? maybe o for octal?

vanchi-zendesk: I am assuming x'blah' is hex and b'blah' is binary. Maybe I am wrong?

vanchi-zendesk:
:+1:

src/main/antlr4/imports/mysql_idents.g4
osheroff: it's actually not liberal enough, https://dev.mysql.com/doc/refman/5.7/en/literals.html shows there's some stuff I'm still missing (`_utf8'string'`).  But ok, I'll see what's available.

osheroff:
merged by hand.

PR #213, opened 2015-12-18
mysql-test, part 3

osheroff:
DO NOT PRESS MERGE.  MERGE INTO MASTER INSTEAD.

support the crazy nonsense of a NATIONAL charset

before the days of the sanity and light that is utf8, there were people who thought it would be good to have a character set that changed depending on what country you might be in.  40 years later, I spend 3
hours of wasted time on that mistake.  Interesting thing, time and history.

Also, LONG is an alias for MEDIUMTEXT.  Which is also kinda baffling.

@zendesk/rules


src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDef.java
```diff
@@ -105,6 +111,8 @@ static public String unalias_type(String type, boolean longStringFlag) {
 			case "real":
 			case "numeric":
 				return "double";
+			case "long":
+				return "mediumtext";
```
vanchi-zendesk: nice touch thinking of writing this function two prs ago.

vanchi-zendesk:
:+1:

osheroff:
merged by hand.

PR #214, opened 2015-12-19
mysql-test, part 4

osheroff:
DO NOT MERGE.  MERGE INTO MASTER.

here we mostly deal with string literals.  we fix parsing of double-backticks and add support for double-quoted literals.

We also filter out successes and failures from mysql-test into their own files.

@zendesk/rules


src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -39,15 +39,17 @@ SQL_COMMENT: '/*' ~'!' (.)*? '*/' -> skip;
 SQL_LINE_COMMENT: ('#' | '--') (~'\n')* ('\n' | EOF) -> skip;
 
 STRING_LITERAL: [a-zA-Z]? TICK ('\\\'' | '\'\'' | ~('\''))* TICK;
-
+DBL_STRING_LITERAL: DBL ('""' | ~('"'))+ DBL;
```
vanchi-zendesk: Ok, this twisted my brain a bit. Might be good to add a comment here.

vanchi-zendesk:
:+1:

osheroff:
merged by hand.

PR #215, opened 2015-12-19
mysql-test, part 5

osheroff:
NO MERGEY MERGE

- bring in `DROP KEY` from another PR
- bring in `STATS_AUTO_RECALC`
- remove a bunch of garbagey test data

@zendesk/rules


lileeyao:
why are we not using codecov in other repos, this tool looks really decent.

vanchi-zendesk:
:+1: 

How do you plan to maintain this test data? It's unreviewable. Could this be generated on the fly instead?

osheroff:
@vanchi-zendesk, it is generated dynamically.  I have three files:

- mysql-test.sql -- this contains the DDL I captured from a run of mysql-test
- mysql-test-errors.sql -- that which is still in error
- mysql-test-fixed.sql -- that which is fixed.

but generated on the fly from mysql-test?  No.  That script is a huge pain, I had to hack it and munge it to generate the file in the first place, and it still output junk data here & there.

osheroff:
@lileeyao we certainly could use codecov elsewhere.  I'm not totally sure I like how it comments on every PR, but it's not bad.  The only thing is that to do anything with our non-public repos people have to sign things, legal has to get involved; dumb stuff like that.

osheroff:
merged manually.

PR #217, opened 2015-12-20
add bootstrapping support

nmaquet:
Add bootstrapping support.

### Using the maxwell-bootstrap utility

You can use the `maxwell-bootstrap` utility to bootstrap tables from the command-line.

option                                        | description
--------------------------------------------- | -----------
--log_level LOG_LEVEL             | log level (DEBUG, INFO, WARN or ERROR)
--user USER                                   | mysql username
--password PASSWORD                           | mysql password
--host HOST                                   | mysql host
--port PORT                                   | mysql port
--database DATABASE                           | mysql database containing the table to bootstrap
--table TABLE                                 | mysql table to bootstrap

### Using the maxwell.bootstrap table

Alternatively you can insert a row in the `maxwell-bootstrap` table to trigger a bootstrap.

```
mysql> insert into maxwell.bootstrap (database_name, table_name) values ('fooDB', 'barTable');
```

### Bootstrapping Data Format

* a bootstrap starts with a document with `type = "bootstrap-start"`
* then documents with `type = "insert"` (one per row in the table)
* then one document per `INSERT`, `UPDATE` or `DELETE` that occurred since the beginning of bootstrap
* finally a document with `type = "bootstrap-complete"`

Here's a complete example:
```
mysql> create table barTable(txt varchar(255));
mysql> insert into barTable (txt) values ("hello"), ("bootstrapping!");
mysql> insert into maxwell.bootstrap (database_name, table_name) values ("fooDb", "barTable");
```
And here's the corresponding replication stream output of table `fooDB`:
```
{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
{...,"type":"bootstrap-start","ts":1450557744340,"data":{}}
{...,"type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
{...,"type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
{...,"type":"bootstrap-complete","ts":1450557744362,"data":{}}
```

### DIFF
--- a/docs/docs/bootstrapping.md
+++ b/docs/docs/bootstrapping.md
@@ -44,14 +44,14 @@ mysql> create table barTable(txt varchar(255));
 mysql> insert into barTable (txt) values ("hello"), ("bootstrapping!");
 mysql> insert into maxwell.bootstrap (database_name, table_name) values ("fooDb", "barTable");
 ```
-And here's the corresponding replication stream output of table `fooDB`:
+Corresponding replication stream output of table `fooDB`:
 ```
-{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
-{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
-{...,"type":"bootstrap-start","ts":1450557744340,"data":{}}
-{...,"type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
-{...,"type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
-{...,"type":"bootstrap-complete","ts":1450557744362,"data":{}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
+{"{"database":"fooDb","table":"barTable","type":"bootstrap-start","ts":1450557744340,"data":{}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
+{"{"database":"fooDb","table":"barTable","type":"bootstrap-complete","ts":1450557744362,"data":{}}
 ```
 
 <script>
--- a/docs/docs/bootstrapping.md
+++ b/docs/docs/bootstrapping.md
@@ -24,6 +24,13 @@ Alternatively you can insert a row in the `maxwell-bootstrap` table to trigger a
 mysql> insert into maxwell.bootstrap (database_name, table_name) values ('fooDB', 'barTable');
 ```
 
+### Async vs Sync bootstrapping
+
+The Maxwell replicator is single threaded; events are captured by one thread from the binlog and replicated to Kafka one message at a time.
+When running Maxwell with `--bootstrapper=sync`, the same thread is used to do bootstrapping, meaning that all binlog events are blocked until bootstrapping is complete.
+Running Maxwell with `--bootstrapper=async` however, will make Maxwell spawn a separate thread for bootstrapping.
+In this async mode, non-bootstrapped tables are replicated as normal by the main thread, while the binlog events for bootstrapped tables are queued and sent to the replication stream at the end of the bootstrap process.
+
 ### Bootstrapping Data Format
 
 * a bootstrap starts with a document with `type = "bootstrap-start"`
--- a/docs/docs/bootstrapping.md
+++ b/docs/docs/bootstrapping.md
@@ -1,0 +1,54 @@
+<div class="content-title">Bootstrapping</div>
+
+***
+
+### Using the maxwell-bootstrap utility
+
+You can use the `maxwell-bootstrap` utility to bootstrap tables from the command-line.
+
+option                                        | description
+--------------------------------------------- | -----------
+--log_level LOG_LEVEL                         | log level (DEBUG, INFO, WARN or ERROR)
+--user USER                                   | mysql username
+--password PASSWORD                           | mysql password
+--host HOST                                   | mysql host
+--port PORT                                   | mysql port
+--database DATABASE                           | mysql database containing the table to bootstrap
+--table TABLE                                 | mysql table to bootstrap
+
+### Using the maxwell.bootstrap table
+
+Alternatively you can insert a row in the `maxwell-bootstrap` table to trigger a bootstrap.
+
+```
+mysql> insert into maxwell.bootstrap (database_name, table_name) values ('fooDB', 'barTable');
+```
+
+### Bootstrapping Data Format
+
+* a bootstrap starts with a document with `type = "bootstrap-start"`
+* then documents with `type = "insert"` (one per row in the table)
+* then one document per `INSERT`, `UPDATE` or `DELETE` that occurred since the beginning of bootstrap
+* finally a document with `type = "bootstrap-complete"`
+
+Here's a complete example:
+```
+mysql> create table barTable(txt varchar(255));
+mysql> insert into barTable (txt) values ("hello"), ("bootstrapping!");
+mysql> insert into maxwell.bootstrap (database_name, table_name) values ("fooDb", "barTable");
+```
+And here's the corresponding replication stream output of table `fooDB`:
+```
+{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
+{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
+{...,"type":"bootstrap-start","ts":1450557744340,"data":{}}
+{...,"type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
+{...,"type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
+{...,"type":"bootstrap-complete","ts":1450557744362,"data":{}}
+```
+
+<script>
+  jQuery(document).ready(function () {
+    jQuery("table").addClass("table table-condensed table-bordered table-hover");
+  });
+</script>
--- a/docs/docs/bootstrapping.md
+++ b/docs/docs/bootstrapping.md
@@ -46,12 +46,12 @@ mysql> insert into maxwell.bootstrap (database_name, table_name) values ("fooDb"
 ```
 Corresponding replication stream output of table `fooDB`:
 ```
-{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
-{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
-{"{"database":"fooDb","table":"barTable","type":"bootstrap-start","ts":1450557744340,"data":{}}
-{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
-{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
-{"{"database":"fooDb","table":"barTable","type":"bootstrap-complete","ts":1450557744362,"data":{}}
+{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
+{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
+{"database":"fooDb","table":"barTable","type":"bootstrap-start","ts":1450557744340,"data":{}}
+{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
+{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
+{"database":"fooDb","table":"barTable","type":"bootstrap-complete","ts":1450557744362,"data":{}}
 ```
 
 <script>
--- a/src/main/java/com/zendesk/maxwell/MaxwellBootstrap.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellBootstrap.java
@@ -1,0 +1,154 @@
+package com.zendesk.maxwell;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import snaq.db.ConnectionPool;
+
+import java.io.Console;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public class MaxwellBootstrap {
+	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellBootstrap.class);
+
+	private static final long UPDATE_PERIOD_MILLIS = 250;
+	private static final long DISPLAY_PROGRESS_WARMUP_MILLIS = 5000;
+	private static Console console = System.console();
+
+	private void run(String[] argv) throws Exception {
+		MaxwellBootstrapConfig config = new MaxwellBootstrapConfig(argv);
+		if ( config.log_level != null ) {
+			MaxwellLogging.setLevel(config.log_level);
+		}
+		ConnectionPool connectionPool = getConnectionPool(config);
+		try ( Connection connection = connectionPool.getConnection() ) {
+			int rowCount = getRowCount(connection, config);
+			long rowId = insertBootstrapStartRow(connection, config);
+			boolean isComplete = false;
+			int insertedRowsCount = 0;
+			Long startedTimeMillis = null;
+			while ( !isComplete ) {
+				if ( insertedRowsCount < rowCount ) {
+					if ( startedTimeMillis == null && insertedRowsCount > 0 ) {
+						startedTimeMillis = System.currentTimeMillis();
+					}
+				 	insertedRowsCount = getInsertedRowsCount(connection, rowId);
+				} else {
+					isComplete = getIsComplete(connection, rowId);
+				}
+				displayProgress(rowCount, insertedRowsCount, startedTimeMillis);
+				Thread.sleep(UPDATE_PERIOD_MILLIS);
+			}
+			displayLine("");
+		} catch ( SQLException e ) {
+			LOGGER.error("Failed to connect to mysql server @ " + config.getConnectionURI());
+			LOGGER.error(e.getLocalizedMessage());
+			System.exit(1);
+		}
+	}
+
+	private int getInsertedRowsCount(Connection connection, long rowId) throws SQLException {
+		String sql = "select inserted_rows from maxwell.bootstrap where id = ?";
+		PreparedStatement preparedStatement = connection.prepareStatement(sql);
+		preparedStatement.setLong(1, rowId);
+		ResultSet resultSet = preparedStatement.executeQuery();
+		resultSet.next();
+		return resultSet.getInt(1);
+	}
+
+	private boolean getIsComplete(Connection connection, long rowId) throws SQLException {
+		String sql = "select is_complete from maxwell.bootstrap where id = ?";
+		PreparedStatement preparedStatement = connection.prepareStatement(sql);
+		preparedStatement.setLong(1, rowId);
+		ResultSet resultSet = preparedStatement.executeQuery();
+		resultSet.next();
+		return resultSet.getInt(1) == 1;
+	}
+
+	private ConnectionPool getConnectionPool(MaxwellBootstrapConfig config) {
+		String name = "MaxwellBootstrapConnectionPool";
+		int maxPool = 10;
+		int maxSize = 0;
+		int idleTimeout = 10;
+		String connectionURI = config.getConnectionURI();
+		String mysqlUser = config.mysqlUser;
+		String mysqlPassword = config.mysqlPassword;
+		return new ConnectionPool(name, maxPool, maxSize, idleTimeout, connectionURI, mysqlUser, mysqlPassword);
+	}
+
+	private int getRowCount(Connection connection, MaxwellBootstrapConfig config) throws SQLException {
+		LOGGER.info("counting rows");
+		String sql = String.format("select count(*) from %s.%s", config.databaseName, config.tableName);
+		PreparedStatement preparedStatement = connection.prepareStatement(sql);
+		ResultSet resultSet = preparedStatement.executeQuery();
+		resultSet.next();
+		return resultSet.getInt(1);
+	}
+
+	private long insertBootstrapStartRow(Connection connection, MaxwellBootstrapConfig config) throws SQLException {
+		LOGGER.info("inserting bootstrap start row");
+		String sql = "insert into maxwell.bootstrap (database_name, table_name) values(?, ?)";
+		PreparedStatement preparedStatement = connection.prepareStatement(sql);
+		preparedStatement.setString(1, config.databaseName);
+		preparedStatement.setString(2, config.tableName);
+		preparedStatement.execute();
+		ResultSet generatedKeys = preparedStatement.getGeneratedKeys();
+		generatedKeys.next();
+		return generatedKeys.getLong(1);
+	}
+
+	private void displayProgress(int total, int count, Long startedTimeMillis) {
+		if ( startedTimeMillis == null ) {
+			displayLine("Waiting for bootstrap to start... ");
+		}
+		else if ( count < total) {
+			long currentTimeMillis = System.currentTimeMillis();
+			long elapsedMillis = currentTimeMillis - startedTimeMillis;
+			long predictedTotalMillis = ( long ) ((elapsedMillis / ( float ) count) * total);
+			long remainingMillis = predictedTotalMillis - elapsedMillis;
+			String duration = prettyDuration(remainingMillis, elapsedMillis);
+			displayLine(String.format("%d / %d (%.2f%%) %s", count, total, ( count * 100.0 ) / total, duration));
+		} else {
+			displayLine("Waiting for bootstrap to stop... ");
+		}
+	}
+
+	private String prettyDuration(long millis, long elapsedMillis) {
+		if ( elapsedMillis < DISPLAY_PROGRESS_WARMUP_MILLIS ) {
+			return "";
+		}
+		long d = (millis / (1000 * 60 * 60 * 24));
+		long h = (millis / (1000 * 60 * 60)) % 24;
+		long m = (millis / (1000 * 60)) % 60;
+		long s = (millis / (1000)) % 60;
+		if ( d > 0 ) {
+			return String.format("- %d days %02dh %02dm %02ds remaining ", d, h, m, s);
+		} else if ( h > 0 ) {
+			return String.format("- %02dh %02dm %02ds remaining ", h, m, s);
+		} else if ( m > 0 ) {
+			return String.format("- %02dm %02ds remaining ", m, s);
+		} else if ( s > 0 ) {
+			return String.format("- %02ds remaining ", s);
+		} else {
+			return "";
+		}
+	}
+
+	private void displayLine(String line) {
+		if ( console != null ) {
+			System.out.print("\u001b[2K\u001b[G" + line);
+			System.out.flush();
+		}
+	}
+
+	public static void main(String[] args) {
+		try {
+			new MaxwellBootstrap().run(args);
+		} catch ( Exception e ) {
+			e.printStackTrace();
+			System.exit(1);
+		}
+	}
+}
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -116,6 +116,8 @@ public class MaxwellConfig {
 			this.producerType = (String) options.valueOf("producer");
 		if ( options.has("bootstrapper"))
 			this.bootstrapperType = (String) options.valueOf("bootstrapper");
+		if ( options.has("bootstrapper_fetch_size"))
+			this.bootstrapperBatchFetchSize = Integer.valueOf((String) options.valueOf("bootstrapper_fetch_size"));
 
 		if ( options.has("kafka.bootstrap.servers"))
 			this.kafkaProperties.setProperty("bootstrap.servers", (String) options.valueOf("kafka.bootstrap.servers"));
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -265,6 +265,10 @@ public class MaxwellConfig {
 			this.maxwellMysql.host = "localhost";
 		}
 
+		if ( this.replicationMysql.host != null && !this.bootstrapperType.equals("none") ) {
+			usage("please specify --bootstrapper=none when specifying a replication host");
+		}
+
 		if ( this.replicationMysql.port == null )
 			this.replicationMysql.port = 3306;
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -27,6 +27,9 @@ public class MaxwellConfig {
 	public final Properties kafkaProperties;
 	public String kafkaTopic;
 	public String producerType;
+	public String bootstrapperType;
+	public Integer bootstrapperBatchFetchSize;
+
 	public String outputFile;
 	public String log_level;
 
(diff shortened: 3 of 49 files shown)


nmaquet:
@osheroff The tests are failing on Travis due to JSON key ordering issues... All tests pass in my local environment though. Not sure how to best fix this.

wushujames:
Woohoo! I'm super excited about this functionality! I'm going to try to find time on Monday to do a code review. I don't have commit rights or anything, but I'm just super interested. Thanks for doing this.

osheroff:
Ok, cool.  Let's start high level and then maybe zoom in closer.  

- Overall code quality looks good! 
- I was pretty lazy at the beginning about doing decent packaging, but I've been trying to get better with new code.  Maybe we can put all new code in `com.zendesk.maxwell.bootstrap`?
- I'm a bit unclear on why `RowMapBufferByTable` was introduced, when it seems like the bootstrapper can only do one table at a time.  Is there a future design in your mind where the async-bootstrapper has multiple sync-bootstrappers and runs more than one table dump?
- Not sure I'm totally sold on `getObjectFromResultSet` -- did you have a reason in mind why that approach is better than just `Object o = rs.getObject(index); columnDef.asJSON(o)`?  
- there was some odd reason I set the timezone to utc.  But I don't remember it, I'll have to look.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -225,6 +233,18 @@ private void setDefaults() {
...
+		if ( this.bootstrapperType == null ) {
+			this.bootstrapperType = "async";
+		} else if ( !this.bootstrapperType.equals("async")
+				&& !this.bootstrapperType.equals("sync")
+				&& !this.bootstrapperType.equals("none") ) {
+			usage("please specify --bootstrapper=async|sync|none");
+		}
+
+		if ( this.bootstrapperBatchFetchSize  == null ) {
+			this.bootstrapperBatchFetchSize = 64000;
```
osheroff: it looks like we're never consuming the value of the `bootstrapper_fetch_size` flag.  But more than that, it's kinda unclear to me that the subsequent `Statement#setFetchSize` call actually does the right thing.  Have you tested this with really big tables to ensure that we're not buffering the entire table into memory?

docs/docs/bootstrapping.md
```diff
@@ -0,0 +1,54 @@
...
+And here's the corresponding replication stream output of table `fooDB`:
+```
+{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
+{...,"type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
+{...,"type":"bootstrap-start","ts":1450557744340,"data":{}}
+{...,"type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
+{...,"type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
+{...,"type":"bootstrap-complete","ts":1450557744362,"data":{}}
+```
+
```
osheroff: let's add a note about what `--bootstrap=async` does vs `--bootstrap-sync` does.

src/main/java/com/zendesk/maxwell/MaxwellBootstrap.java
```diff
@@ -0,0 +1,154 @@
...
+			System.exit(1);
+		}
+	}
+
+	private int getInsertedRowsCount(Connection connection, long rowId) throws SQLException {
+		String sql = "select inserted_rows from maxwell.bootstrap where id = ?";
+		PreparedStatement preparedStatement = connection.prepareStatement(sql);
+		preparedStatement.setLong(1, rowId);
+		ResultSet resultSet = preparedStatement.executeQuery();
+		resultSet.next();
```
osheroff: I really shoulda chosen some lightweight ORM :) Sorry about that.

src/main/java/com/zendesk/maxwell/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,200 @@
...
+			lastInsertedRowsUpdateTimeMillis = now;
+		}
+	}
+
+	protected Connection getConnection( ) throws SQLException {
+		return context.getConnectionPool().getConnection();
+	}
+
+	private RowMap replicationStreamBootstrapStartRow(Table table, BinlogPosition position) {
+		return replicationStreamBootstrapRow("bootstrap-start", table, position);
```
osheroff: ok, so the `bootstrap-start` and `bootstrap-complete` events *do* have a db/table name.  That's not documented in bootstrap.md, can you fix?

src/main/java/com/zendesk/maxwell/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	private RowMap replicationStreamBootstrapCompletedRow(Table table, BinlogPosition position) {
+		return replicationStreamBootstrapRow("bootstrap-complete", table, position);
+	}
+
+	private RowMap replicationStreamBootstrapRow(String type, Table table, BinlogPosition position) {
```
osheroff: suggest renaming this suite of functions `bootstrapEventRowMap()` and similar

src/main/java/com/zendesk/maxwell/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,200 @@
...
+		return database;
+	}
+
+	private void ensureTable(String tableName, Database database) {
+		findTable(tableName, database);
+	}
+
+	private ResultSet getAllRows(String databaseName, String tableName, Connection connection) throws SQLException, InterruptedException {
+		Statement statement = createBatchStatement(connection);
+		return statement.executeQuery(String.format("select * from %s.%s", databaseName, tableName));
```
osheroff: let's go ahead and order by the table's primary key here.  If you leave off any ordering it's just "database order" which is largely undefined, although often it can be ordered by the PK.

src/main/java/com/zendesk/maxwell/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	private void setRowValues(RowMap row, ResultSet resultSet, Table table) throws SQLException, IOException {
+		Iterator<ColumnDef> columnDefinitions = table.getColumnList().iterator();
+		int columnIndex = 1;
+		while ( columnDefinitions.hasNext() ) {
+			ColumnDef columnDefinition = columnDefinitions.next();
+			row.putData(columnDefinition.getName(), columnDefinition.getObjectFromResultSet(resultSet, columnIndex));
+			++columnIndex;
+		}
```
osheroff: let's do a sanity check (maybe just once per ResultSet) that the columnDefinitions match up with the resultSet; there's a race possible where the schema that's in-memory at the time we start the bootstrap does not match the schema that's active at the moment on the server.  ie think about:

- stop maxwell
- enqueue a bootstrap row
- add a column
- start maxwell

now the schema present at the time the row is enqueued is different than what SELECT * will give.  So we should catch this issue.

osheroff: this actually brings up a larger, thornier issue.  If maxwell falls behind in the replication stream, a bootstrap process could output old data to the stream.  think about:
 
- there's a row that has had the following done:
```
   INSERT INTO t set a = 'first_value'; # -> id = 1
   UPDATE t set a = 'second_value' where id = 1;
```
- maxwell is 6 days behind, with a bootstrap row in the queue
- maxwell starts up, and starts the bootstrapping process. 

now the following will result:
```
  -> start-bootstrap
  -> { type: "insert", data: { a: "second_value" } }
  -> complete-bootstrap
  -> { type: "insert", data: { a: "first_value" } }
  -> { type: "update", data: { a: "second_value" } }
```

which isn't too too bad as the values converge, but it's a bit ugly.

osheroff: aha.  here you go:  capture the binlog position right before you start the `SELECT *`.  throw out any `insert/update` rows who have a binlog position lesser than when you started the SELECT.  I'm not sure how nasty that is code-wise.

osheroff:
that's enough CR for now.  Thanks so much for this contribution, this will be a great addition to Mawell once it lands.

nmaquet:
Re `RowMapBufferByTable`: not sure if it's the best way to go but currently when a table is queued for bootstrapping, all further events for that table are "skipped", i.e. not sent to the replication stream but stored in a disk-backed buffer instead (see `AsyncBootstrapper.shouldSkip`). At the end of the bootstrapping these events are then sent to the replication stream and events are no longer skipped. Since multiple tables can be queued at the same time, `RowMapBufferByTable` allows to have one buffer for each.

src/main/java/com/zendesk/maxwell/bootstrap/AsynchronousBootstrapper.java
```diff
@@ -0,0 +1,142 @@
...
+		if ( bootstrappedRow != null && haveSameTable(row, bootstrappedRow) ) {
+			skippedRows.add(row);
+			return true;
+		}
+		for ( RowMap queuedRow : queue ) {
+			if ( haveSameTable(row, queuedRow) ) {
+				skippedRows.add(row);
+				return true;
+			}
+		}
```
osheroff: ok, I missed that we queue up a bunch of bootstraps to do and then skip these too.

src/main/java/com/zendesk/maxwell/bootstrap/AsynchronousBootstrapper.java
```diff
@@ -0,0 +1,142 @@
...
+		String databaseName = ( String ) bootstrapStartRow.getData("database_name");
+		String tableName = ( String ) bootstrapStartRow.getData("table_name");
+		return row.getDatabase().equals(databaseName) && row.getTable().equals(tableName);
+	}
+
+	@Override
+	public void startBootstrap(final RowMap startBootstrapRow, final Schema schema, final AbstractProducer producer, final OpenReplicator replicator) throws Exception {
+		queueRow(startBootstrapRow);
+		if (thread == null) {
+			final RowMap row = bootstrappedRow = queue.remove();
```
osheroff: took me awhile to grasp what's going on here.  Could we just pass the row into the thread as a local?  (or if there's already a thread, queue it up).

Another maybe more involved approach would be to always have a thread running in the background that's just doing 
```
while ( true ) { 
   RowMap row = queue.poll()
   if ( row == null )  { sleep(..); continue; }
   ...
}
```

and then just push stuff into its queue.

docs/docs/bootstrapping.md
```diff
@@ -0,0 +1,61 @@
...
+Alternatively you can insert a row in the `maxwell-bootstrap` table to trigger a bootstrap.
+
+```
+mysql> insert into maxwell.bootstrap (database_name, table_name) values ('fooDB', 'barTable');
+```
+
+### Async vs Sync bootstrapping
+
+The Maxwell replicator is single threaded; events are captured by one thread from the binlog and replicated to Kafka one message at a time.
+When running Maxwell with `--bootstrapper=sync`, the same thread is used to do bootstrapping, meaning that all binlog events are blocked until bootstrapping is complete.
```
wushujames: @osheroff, if Maxwell is blocked, what happens to the events that are coming in from OpenReplicator? Does open replicator load everything into memory and then blocked when sending them to maxwell? Or is open replicator smart enough to only pull events from the mysql server as it processes them?

I guess the general question is, is there any point where things are being loaded into memory without limit?

src/main/java/com/zendesk/maxwell/bootstrap/AsynchronousBootstrapper.java
```diff
@@ -0,0 +1,140 @@
...
+		return synchronousBootstrapper.isBootstrapRow(row);
+	}
+
+	protected SynchronousBootstrapper getSynchronousBootstrapper( ) {
+		return new SynchronousBootstrapper(context);
+	}
+
+	@Override
+	public boolean shouldSkip(RowMap row) throws SQLException, IOException {
+		// The main replication thread skips rows of the currently bootstrapped
```
wushujames: I'm trying to understand the async bootstrapper and how it interacts with the kafka checkpoints. 

I understand the part where rows from bootstrapped tables are saved into memory as the bootstrap is happening and then are emitted.

When rows from non-bootstrapped tables are sent to the producer, won't the producer checkpoint those rows? The checkpoint for those rows might be beyond the point of the "maxwell.bootstrap" row, and upon crash, they would not be restarted.

And it looks like when you send the INSERTs of a bootstrap, won't the producer.send() end up checkpointing the rows for those INSERTS?

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrap.java
```diff
@@ -0,0 +1,155 @@
+package com.zendesk.maxwell.bootstrap;
```
wushujames: It looks like this file doesn't actually DO the bootstrap, but it inserts a row into maxwell.bootstrap so that Maxwell can do the bootstrap? And it also lets you monitor the progress of the bootstrap?

From the filename, I thought that this command was actually performing a bootstrap. Maybe the file should be named MaxwellBootstrapAdmin or MaxwellBootstrapController or something?

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,212 @@
...
+		Database database = findDatabase(schema, databaseName);
+		Table table = findTable(tableName, database);
+		OpenReplicator openReplicator = replicator.getOpenReplicator();
+		BinlogPosition position = new BinlogPosition(openReplicator.getBinlogPosition(), openReplicator.getBinlogFileName());
+		producer.push(startBootstrapRow);
+		producer.push(bootstrapStartRowMap(table, position));
+		LOGGER.info(String.format("bootstrapping started for %s.%s, binlog position is %s", databaseName, tableName, position.toString()));
+		try ( Connection connection = getConnection() ) {
+			setBootstrapRowToStarted(startBootstrapRow, connection);
+			ResultSet resultSet = getAllRows(databaseName, tableName, schema, connection);
```
wushujames: I'm pretty sure I understand, but I just wanted to check:

This does a *single* "select *" command, right? And then JDBC (may) chunk it into bootstrap_fetch_size chunks internally?

It doesn't do multiple "select *" commands? Because that wouldn't work, consistency-wise. You want the bootstrap to be a single consistent snapshot.

nmaquet: Correct, there's one single `select *` and we're relying on the JDBC driver to get the fetch size into account so that the entire resultset isn't stored in memory.

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrap.java
```diff
@@ -0,0 +1,155 @@
...
+		}
+		ConnectionPool connectionPool = getConnectionPool(config);
+		try ( Connection connection = connectionPool.getConnection() ) {
+			int rowCount = getRowCount(connection, config);
+			long rowId = insertBootstrapStartRow(connection, config);
+			boolean isComplete = false;
+			int insertedRowsCount = 0;
+			Long startedTimeMillis = null;
+			while ( !isComplete ) {
+				if ( insertedRowsCount < rowCount ) {
```
wushujames: getRowCount() may return a different number of rows than the actual bootstrap is seeing. Consider the following scenario:

1. 1 million records in table.
2. getRowCount() returns 1 million
3. insertBootstrapStartRow()
4. 1 million rows deleted by outside process
5. Maxwell sees the bootstrap start row.
6. Maxwell does a "select *" and inserts 0 rows into the stream.
7. Maxwell marks the bootstrap row as completed.

insertedRowsCount will always be 0. This while loop will never set isComplete and will block forever.

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrap.java
```diff
@@ -0,0 +1,155 @@
...
+		} else if ( s > 0 ) {
+			return String.format("- %02ds remaining ", s);
+		} else {
+			return "";
+		}
+	}
+
+	private void displayLine(String line) {
+		if ( console != null ) {
+			System.out.print("\u001b[2K\u001b[G" + line);
```
wushujames: What do these escape codes do?

src/main/java/com/zendesk/maxwell/SynchronousBootstrapper.java
wushujames: @osheroff, the "thornier issue" that you describe is actually the core of it: do you want a consistent stream, or an eventually consistent stream? The approach taken here, as well as the approach currently being implemented for mypipe (in https://github.com/mardambey/mypipe/commit/9c1cfd8420c5fcc1d14e89c7866c2c6fa3e5e4f8) is to make the stream eventually-consistent.

See discussion here: https://github.com/mardambey/mypipe/issues/28#issuecomment-166033049

wushujames: Agree about the sanity check. In order for this to work, the users of mysql would have to make not to do any DDL statements while the bootstrap is in progress.

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -93,11 +93,12 @@ public void work() throws Exception {
 		if (row == null)
 			return;
 
-		if (!skipRow(row)) {
+		if ( !bootstrapper.shouldSkip(row) && !isMaxwellRow(row) ) {
 			producer.push(row);
+		} else {
+			bootstrapper.work(row, producer, this);
```
wushujames: This depends on the fact that the maxwell.bootstrap is writing to the same binlogs as Maxwell is listening to.

There is work in https://github.com/zendesk/maxwell/pull/222 that is making it possible to move the maxwell.schemas table to another mysql host. @osheroff, does that also move all the other maxwell tables to the other host? If so, then these two pull requests would conflict with each other.

nmaquet:
Re `Statement.setFetchSize`: from the Oracle docs:
> Gives the JDBC driver a hint as to the number of rows that should be fetched from the database when more rows are needed for ResultSet objects genrated by this Statement.
So there's unfortunately no garantee that the JDBC drive will do the right thing.

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
```diff
@@ -38,10 +35,10 @@ public boolean matchesMysqlType(int type) {
 	private String formatValue(Object value) {
 		if ( value instanceof Long && getType().equals("datetime") )
 			return formatLong((Long) value);
-		else if ( value instanceof Date )
-			return getDateTimeFormatter().format((Date) value);
 		else if ( value instanceof Timestamp )
 			return getDateTimeFormatter().format((Timestamp) value);
+		else if ( value instanceof Date )
```
wushujames: Any particular reason you changed the ordering here? Is one a superclass of the other? Does it have any impact on existing tests/behavior?

src/test/java/com/zendesk/maxwell/BootstrapIntegrationTest.java
```diff
@@ -0,0 +1,181 @@
...
+import static org.junit.Assert.assertThat;
+
+public class BootstrapIntegrationTest extends AbstractMaxwellTest {
+
+	@Test
+		public void testSingleRowBootstrap( ) throws Exception {
+		List<RowMap> list;
+		String input[] = {
+			"insert into minimal set account_id = 1, text_field='hello'",
+			"insert into maxwell.bootstrap set database_name = 'shard_1', table_name = 'minimal'"
```
wushujames: Is it worth adding a test for bootstraps of multiple tables? Although I suppose it might get hard if there are ordering issues (multiple bootstraps going on simultaneously).

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
nmaquet: `java.sql.Date` and `java.sql.Timestamp` are both subclasses of `java.util.Date` so the if statement `if ( value instanceof Timestamp )` was previous unreachable.

docs/docs/bootstrapping.md
osheroff: open-replicator feeds into a LinkedBlockingQueue of a few hundred events, and will just start blocking if maxwell isn't there to get them.  no biggie.

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
osheroff: hm, good point.  Ok, I think we say that for #222 it's supported to run the replication thread off the slave and keep the schema store on the master (one seemingly popular config), but if you want to disconnect them entirely then bootstrapping won't work.

osheroff:
@nmaquet re: `setFetchSize`, I don't think we've got the incantations right to trigger the cursor code path in mysql.  I started a local bootstrap of a table with 7mm rows in it, dumped the heap, and:

image

I think if you look at https://github.com/mysql/mysql-connector-j/blob/fe541c166cec739c74cc727c5da96c1028b4834a/src/com/mysql/jdbc/MysqlIO.java#L432 we're probably missing some of the cursor conditions.

src/test/java/com/zendesk/maxwell/BootstrapIntegrationTest.java
nmaquet: Done. Like you said it was a bit tricky to do because of the ordering but you're right it's better to have that additional test.

nmaquet:
@osheroff Sorry for all the Travis CI failures... I hope this doesn't spam your inbox. Testing multithreaded coded is tricky > >

wushujames:
Pretty cool to see @osheroff sending a pull request to get something merged into @nmaquet's branch which itself is currently on a pull request back to @osheroff .

osheroff:
hmm @nmaquet, got a chance to play with the bootstrap code on a large local table tonight, and I got myself into a weird situation:

- I inserted some rows into maxwell.bootstrap with the utility.
- I ended up creating like 3 bootstrap rows, as the utility didn't seem to want to re-use existing rows. 
- Then I decided to delete the rows to start over.
- Then I found myself in a bit of a fubar'ed state, where the replication stream still had rows to process, but I had deleted the rows out of the database (so there wasn't much state reporting), and there didn't seem to be anything that could be done that would cancel the process save for letting the bootstrap continue.

Now this of course is a bit of a dumb thing operationally and a normal person might just let the thing go, but I think there may be issues with trying to co-ordinate the two pieces of state (the replication stream and the row itself at once).  Consider again the problem of the replication stream lagging behind by a day: if I send in a "start bootstrap" command, I still probably want it to start *now*, not once maxwell is caught up.  This may or may not be a totally likely scenario, but I could see:

- take maxwell down, delete the kafka topics so that maxwell starts fresh-and-clean
- enqueue bootstraps
- bring maxwell back up

The expectation is probably that the very first thing maxwell does is to start bootstrapping, not to play the replication stream and then start.

Can we mull over what it might look like to use maxwell.boostraps as a "work queue" table that the bootstrap threads poll regularly, rather than trying to do edge-triggered notifications with the replication stream?  What are the trade-offs of each?  Running out the door.  More later.

wushujames:
Up until now, I don't think we've ever needed to "control" Maxwell. You just set it up and it did its thing. 

A more drastic idea than "maxwell.bootstraps as work queue": what if we add a control API to Maxwell? A REST interface that you use to add/remove/query bootstrap jobs. You could expand to also doing other admin-y type stuff like pausing Maxwell, querying binlog location, controlled shutdown, etc.

jpeyper:
I would actually love an API from maxwell that could be used to interact with it and show some "admin-y" type stuff.

I know its not a definite thing, but if such an enhancement was considered, (very) roughly how long would it take to develop.

osheroff:
> A more drastic idea than "maxwell.bootstraps as work queue": what if we add a control API to Maxwell?

I'm not necessarily for-or-against the idea, although stuffing a REST server in there is quite a lot of work, and I think you'd need a strong case for doing so.  In either case, I *think* you'd want to persist queued bootstraps into a table like @nmaquet has going, rather than having them be ephermal and die along with the process. 

So, hand wavy, maybe-if-someone-contributed on the HTTP server, but I think it's definitely outside the scope of this PR, and I don't want to block getting bootstrapping going behind getting a fully-fledged admin server going.

docs/docs/bootstrapping.md
```diff
@@ -0,0 +1,61 @@
...
+mysql> insert into maxwell.bootstrap (database_name, table_name) values ("fooDb", "barTable");
+```
+Corresponding replication stream output of table `fooDB`:
+```
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"hello"}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557598,"xid":13561,"data":{"txt":"bootstrapping!"}}
+{"{"database":"fooDb","table":"barTable","type":"bootstrap-start","ts":1450557744340,"data":{}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744355,"data":{"txt":"hello"}}
+{"{"database":"fooDb","table":"barTable","type":"insert","ts":1450557744356,"data":{"txt":"bootstrapping!"}}
+{"{"database":"fooDb","table":"barTable","type":"bootstrap-complete","ts":1450557744362,"data":{}}
```
osheroff: data-format typos here:  should read 
```
  { "database"
```
not

```
{"{"database"
```

src/test/java/com/zendesk/maxwell/BootstrapIntegrationTest.java
```diff
@@ -0,0 +1,259 @@
...
+		};
+		String expectedJSON[] = {
+			"{\"database\":\"shard_1\",\"table\":\"minimal\",\"type\":\"insert\",\"ts\":0,\"xid\":0,\"data\":{\"id\":0,\"account_id\":1,\"text_field\":\"hello\"}}",
+			"{\"database\":\"maxwell\",\"table\":\"bootstrap\",\"type\":\"insert\",\"ts\":0,\"xid\":0,\"data\":{\"id\":0,\"database_name\":\"shard_1\",\"table_name\":\"minimal\",\"is_complete\":0}}",
+			"{\"database\":\"shard_1\",\"table\":\"minimal\",\"type\":\"bootstrap-start\",\"ts\":0,\"data\":{}}",
+			"{\"database\":\"shard_1\",\"table\":\"minimal\",\"type\":\"insert\",\"ts\":0,\"data\":{\"id\":0,\"account_id\":1,\"text_field\":\"hello\"}}",
+			"{\"database\":\"maxwell\",\"table\":\"bootstrap\",\"type\":\"update\",\"ts\":0,\"xid\":0,\"data\":{\"id\":0,\"database_name\":\"shard_1\",\"completed_at\":\"\",\"table_name\":\"minimal\",\"started_at\":\"\",\"is_complete\":1},\"old\":{\"completed_at\":null,\"is_complete\":0}}",
+			"{\"database\":\"shard_1\",\"table\":\"minimal\",\"type\":\"bootstrap-complete\",\"ts\":0,\"data\":{}}"
```
osheroff: generally I've been liking to keep JSON as raw files on-disk, just so it doesn't look so hideous with all the backslashes.  no big deal tho.

osheroff:
@nmaquet any thoughts on the control channel stuff?  fwiw I don't think it's necessarily a blocker to getting this stuff merged, the rest of it looks like it's in pretty good shape to go.

wushujames:
Yeah, I agree that a rest interface is big enough for a different PR and should not block this. 

And yeah, a REST interface would still need to persist the changes somewhere. The advantage would be that the REST interface would be the only supported way of interacting with that "work queue" table and so it would be able to appropriately stop jobs before deleting a row from the table. So it wouldn't run into the problem you got into (row gone but bootstrap still keeps running).

nmaquet:
@osheroff Re: multiple bootstrap rows and a fu-bared state: this situation happened to us also. We didn't really think twice about it and just dropped the maxwell schema when it happened.
However, I understand that this may come up during ops tasks so we should try to make this better.
What do you think of the following:
* when CTRL-C-ing out of `maxwell-bootstrap`, it deletes the bootstrap row and warns the user the bootstrap has been cancelled
* when the bootstrapper updates the `inserted_rows` count, it also checks whether the row exists and cancels the bootstrap accordingly
* when the bootstrapper starts a new bootstrap (whether due to capturing the binlog event or proceeding to the next queued bootstrap), it also checks for the presence of the bootstrap row and does not start bootstrapping if the row is absent. 

Re: bootstrapping *now* vs bootstrapping when maxwell has caught up: we've thought about this quite a bit and it feels quite tricky. It would definitely be an improvement over the current implementation but we feel that it should probably warrant its own PR / effort.
Happy to give a go at this after the Christmas break.

Re: REST interface: at this stage we wouldn't really benefit from this directly. If maxwell gets a REST interface there would definitely be some value in hooking up the bootstrap code to it though.

osheroff:
@nmaquet that sounds like a good-for-now kind of solution.  Maybe we shoot for being able to cancel a bootstrap in this PR and try for a larger control rework in a subsequent one?  

Happy Christmas or whenever it is you say in New Zealand.

nmaquet:
@osheroff Happy holidays. I'm away until the 5th of Jan, but I'll finish this ASAP when I get back.

nmaquet:
@osheroff Happy New Year. Bootstrap canceling is now implemented.
Also, I've fixed a crash in the bootstrapping code that occurrend on improperly formatted dated.
I had to add `zeroDateTimeBehavior=convertToNull` to the connection string.

docs/docs/bootstrapping.md
```diff
@@ -0,0 +1,61 @@
+<div class="content-title">Bootstrapping</div>
+
+***
```
osheroff: I can fix up documents to my tweak-level after merge, but I'd like a summary of what bootstrapping is/does.

src/main/java/com/zendesk/maxwell/bootstrap/AsynchronousBootstrapper.java
```diff
@@ -0,0 +1,162 @@
...
+
+	@Override
+	public boolean isCompleteBootstrapRow(RowMap row) {
+		return synchronousBootstrapper.isCompleteBootstrapRow(row);
+	}
+
+	@Override
+	public boolean isBootstrapRow(RowMap row) {
+		return synchronousBootstrapper.isBootstrapRow(row);
+	}
```
osheroff: since these functions are side-effect free and there's only one implementation (in synchronousBootstrapper), let's move them to the abstract class.

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,236 @@
...
+
+	@Override
+	public void completeBootstrap(RowMap completeBootstrapRow, AbstractProducer producer, MaxwellReplicator replicator) throws Exception {
+		String databaseName = ( String ) completeBootstrapRow.getData("database_name");
+		String tableName = ( String ) completeBootstrapRow.getData("table_name");
+		Database database = findDatabase(replicator.getSchema(), databaseName);
+		ensureTable(tableName, database);
+		Table table = findTable(tableName, database);
+		OpenReplicator openReplicator = replicator.getOpenReplicator();
+		BinlogPosition position = new BinlogPosition(openReplicator.getBinlogPosition(), openReplicator.getBinlogFileName());
```
osheroff: oh, this isn't quite correct.  the open-replicator's binlog position may be ahead of maxwell's, as it runs in an asynchronous thread.  What you really want to do is attach a `BinlogPosition` object to the `RowMap`.

src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
```diff
@@ -12,12 +11,10 @@ public DateColumnDef(String tableName, String name, String type, int pos) {
 	}
 
 	private static SimpleDateFormat dateFormatter;
-	private static final TimeZone tz = TimeZone.getTimeZone("UTC");
 
 	protected static SimpleDateFormat getDateFormatter() {
 		if ( dateFormatter == null ) {
 			dateFormatter = new SimpleDateFormat("yyyy-MM-dd");
-			dateFormatter.setTimeZone(tz);
```
osheroff: I remembered why I had done this: https://github.com/zendesk/open-replicator/issues/14 -- otherwise open-replicator will pick up maxwell's timezone, which is probably not what the user wants (they probably want the data interpreted "raw").

osheroff:
@nmaquet, happy new years.  Once we round-up this last few issues I'll be ready to merge this into master, although I think what I will do is to release a v0.17.0 with all of the parser and bug-fixes I've worked on over the holidays, then merge bootstrapping and release a v1.0-PRE1 with it included.  It's a big enough change that I'd like to have it burned in in someone's production before I release it GA.  sound good?

thanks!

nmaquet:
@osheroff Re merging in v1.0-PRE1: that sounds great actually. It will give us the time to do more testing with realistic loads and iron out the remaining kinks.

src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
nmaquet: @osheroff Do you mind handling that change when merging in? I'm a bit confused about this. All our servers always run in UTC so for our use it doesn't matter either way. I can look deeper if you'd  prefer.

osheroff: yeah I can look and add commits as need when merging.

osheroff:
ok, I think we're good to go here.  Anything else you want to fix up?  If not I'll take a final look PST/AM and probably merge.  :+1:

nmaquet:
@osheroff All good! I've tested it heavily for the past few days and haven't found more bugs thus far. We've measured the performance of bootstrapping and we get a very stable average throughput of 3 million rows / minute (50k/s). I'm curious to hear whether you get faster results.

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -0,0 +1,221 @@
...
+		try ( Connection connection = getConnection() ) {
+			setBootstrapRowToStarted(startBootstrapRow, connection);
+			ResultSet resultSet = getAllRows(databaseName, tableName, schema, connection);
+			int insertedRows = 0;
+			while ( resultSet.next() ) {
+				RowMap row = new RowMap(
+						"insert",
+						databaseName,
+						tableName,
+						System.currentTimeMillis(),
```
osheroff: hey @nmaquet, I realize post-merge that the `ts` row is inconsistent -- the bootstrapped rows use milliseconds, whereas the replicated rows use seconds.  can you fix to use seconds in a separate PR?

PR #218, opened 2015-12-20
mysql-test, part 6

osheroff:
NO M.
NOOOOO M.

- parse `drop view`, it's easy enough
- alter table force
- the "to" in "rename to" is optional
- support for token-type names in AFTER column

@zendesk/rules


vanchi-zendesk:
:+1:

osheroff:
merged by hand.

PR #220, opened 2015-12-21
mysql-test, part 8

osheroff:
- BLOB and TEXT take optional sizes that turn them into other column types
- DROP FOREIGN KEY
- current_timestamp also takes an optional size.


src/main/antlr4/imports/column_definitions.g4
```diff
@@ -67,7 +69,7 @@ column_options:
 	  nullability
 	| default_value
 	| primary_key
-	| ON UPDATE ( CURRENT_TIMESTAMP | now_function )
+	| ON UPDATE ( CURRENT_TIMESTAMP length? | now_function )
```
vanchi-zendesk: this also seems to work:
```
mysql> select NOW(1);
+---------------------+
| NOW(1)              |
+---------------------+
| 2015-12-22 09:37:17 |
+---------------------+
1 row in set (0.00 sec)
```

vanchi-zendesk: http://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_now

vanchi-zendesk:
:+1:

src/main/antlr4/imports/column_definitions.g4
osheroff: yes, but doesn't seem to be valid for alters:

```
mysql> create table t ( i timestamp default now() on update now(1) );
ERROR 1294 (HY000): Invalid ON UPDATE clause for 'i' column
mysql> create table t ( i timestamp default now() on update now() );
Query OK, 0 rows affected (0.02 sec)
```

osheroff:
actually am going to merge these all into a large topic branch.

PR #222, opened 2015-12-21
Different mysql for schema

### LINKED ISSUE #208: Allow for separation of schema store and replication stream
When I'm running maxwell, I get the following error:

WARN  MaxwellReplicator - Unhandled QueryEvent inside transaction: QueryEvent[header=BinlogEventV4HeaderImpl[timestamp=1450339656000,eventType=2,serverId=10,eventLength=259,nextPosition=684603801,flags=0,timestampOfReceipt=1450344595074],threadId=2935445014,elapsedTime=0,databaseNameLength=[],errorCode=0,statusVariablesLength=26,statusVariables=[QFlags2Code[flags=0], QSQLModeCode[sqlMode=0], QCatalogNzCode[catalogName=std], QCharsetCode[characterSetClient=8,collationConnection=8,collationServer=8]],databaseName=******,sql= *****AN INSERT QUERY****]

seems like eventType=2 in the above object is not handled in the code.

Please help.

kristiankaufmann:
This pull requests allows maxwell to read binlogs from a separate mysql instance from the mysql instance used to store maxwell's state. It adds new config options
  * replication_host
  * replication_port
  * replication_user
  * replication_password

resolves #208


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -23,6 +23,12 @@
 	public Integer mysqlPort;
 	public String  mysqlUser;
 	public String  mysqlPassword;
+
+	public String mysqlSchemaHost;
+	public Integer mysqlSchemaPort;
```
kristiankaufmann: I think shoudl make a MySQLConfig class and have two instances of the new class rather than the repeated code you see here.

osheroff: sure, go for it.  the other thing I was thinking about was that maybe we could accept uris: 

```
  --mysql='mysql://user:password@server:port'
  --schema_mysql='...'
```

wushujames:
Will you be required to specify both the schema_host as well as the normal mysql_host? Or can you leave off schema_host and have it fall back automatically to storing it in the mysql_host?

Not sure which is better. Automatic fallback can cause problems if you are not expecting it.

kristiankaufmann:
at the moment the code falls back automatically from --schema_host to --mysql_host, but I am open to both. Perhaps inverting the fallback so that if the target host for replication falls back to the schema host instead of the way that I have would be better.

jpeyper:
is there any space in this PR to also be able to specify the actual schema maxwell uses? 

for example,
config setting
`maxwell_schema=maxwell_dev`
will create and populate
`maxwell_dev.positions`
`maxwell_dev.schema`
etc..

With the current changes to allow Maxwell to store its positions and stuff on a different database to the one it is monitoring, its not a huge leap for someone to want to centralize all their environments to have
`maxwell_dev`
`maxwell_qa`
`maxwell_my_super_awesome_environment_name`
all writing to the same server.

osheroff:
@kristiankaufmann @wushujames I think this code has the right approach in general; the intention is that mostly people will just configure a single maxwell host and that'll be enough, which will be the one configured in `--host=`

kristiankaufmann:
@osheroff I came to the same conclusion as I was writing the latest updates. So host now points to the mysql where maxwell keeps state. And the new optional parameters make it possible to point to a different replication_host that acts a master.

I've tested on my local box and you can start up maxwell with just the --password specified on the command line.

kristiankaufmann:
@jpeyper I'll look into adding a parameter to control Maxwells database, but I think it should go into another PR. Can you add a new issue requesting this feature?

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -99,14 +102,36 @@ private void parse(String [] argv) {
...
+			this.maxwellMysql.mysqlPassword = (String) options.valueOf("password");
 		if ( options.has("user"))
-			this.mysqlUser = (String) options.valueOf("user");
+			this.maxwellMysql.mysqlUser = (String) options.valueOf("user");
 		if ( options.has("port"))
-			this.mysqlPort = Integer.valueOf((String) options.valueOf("port"));
+			this.maxwellMysql.mysqlPort = Integer.valueOf((String) options.valueOf("port"));
+
+		if ( options.has("replication_host"))
+			this.replicationMysql.mysqlHost = (String) options.valueOf("replication_host");
```
osheroff: this could be DRY'ed, something like 
``
  class MaxwellMysqlConfig {  
    public void parseOptions(String prefix, Options options) { 
      if ( options.has(prefix + "host") ) {
         ...
      }
   }
 }
```

File: src/main/java/com/zendesk/maxwell/MaxwellConfig.java [hunk 3]
```diff
@@ -99,14 +102,36 @@ private void parse(String [] argv) {
...
+			this.replicationMysql.mysqlUser = (String) options.valueOf("replication_user");
+
+		if ( options.has("replication_port"))
+			this.replicationMysql.mysqlPort = Integer.valueOf((String) options.valueOf("replication_port"));
+
+		if ( this.replicationMysql.mysqlPassword == null && !options.has("replication_password")) {
+			this.replicationMysql = new MaxwellMysqlConfig(this.maxwellMysql.mysqlHost,
+															this.maxwellMysql.mysqlPort,
+															this.maxwellMysql.mysqlUser,
+															this.maxwellMysql.mysqlPassword);
```
osheroff: whitespace issues here;  Also I think the rule should probably be that if you specify `replication_host` you need to specify `replication_user` and `replication_password` -- here if you specify `replication_host` but not `replication_password` you'd end up just going back to the standard config.

osheroff: And also the code where we let the replica fallback to the schema server should probably be in `setDefaults`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -225,16 +231,43 @@ private void setDefaults() {
...
+			if (this.replicationMysql.mysqlHost != null
+					|| this.replicationMysql.mysqlUser != null
+					|| this.replicationMysql.mysqlPassword != null) {
+				usage("Specified a replication option but missing one of the following options: replication_host, replication_user, replication_password.");
+			}
+			
+			this.replicationMysql = new MaxwellMysqlConfig(this.maxwellMysql.mysqlHost,
+															this.maxwellMysql.mysqlPort,
+															this.maxwellMysql.mysqlUser,
+															this.maxwellMysql.mysqlPassword);
```
kristiankaufmann: I'll fix the white space here

kristiankaufmann:
@osheroff this is ready for another look

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -45,20 +47,21 @@ public MaxwellConfig(String argv[]) {
...
 		parser.accepts( "config", "location of config file" ).withRequiredArg();
 		parser.accepts( "log_level", "log level, one of DEBUG|INFO|WARN|ERROR" ).withRequiredArg();
-		parser.accepts( "host", "mysql host" ).withRequiredArg();
-		parser.accepts( "user", "mysql username" ).withRequiredArg();
+		parser.accepts( "replication_host", "mysql host to treat as master" ).withRequiredArg();
+		parser.accepts( "replication_user", "mysql username with replication privileges on replication_host" ).withRequiredArg();
 		parser.accepts( "output_file", "output file for 'file' producer" ).withRequiredArg();
-		parser.accepts( "password", "mysql password" ).withRequiredArg();
-		parser.accepts( "port", "mysql port" ).withRequiredArg();
+		parser.accepts( "replication_password", "mysql password for replication_user" ).withRequiredArg();
```
osheroff: let's have these say "(if using a separate connection)"

src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Created by kaufmannkr on 12/23/15.
+ */
+public class MaxwellMysqlConfig {
+
+	public String mysqlHost;
+	public Integer mysqlPort;
+	public String mysqlUser;
+	public String mysqlPassword;
```
osheroff: no need to reiterate "mysql" here I think.  These fields can be just "host, port, user, password"

osheroff:
cool.  :+1: after nitty little changes

kristiankaufmann:
@osheroff fixed the naming and changed the option descriptions.

osheroff:
:+1:

PR #226, opened 2015-12-24
mysql-test, part 9

osheroff:
allow index definitions inside ADD COLUMN ( ... ) format.
this also fixes handling of that form

@zendesk/rules


src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -166,7 +166,7 @@ public void exitAdd_column_parens(mysqlParser.Add_column_parensContext ctx) {
 		while ( this.columnDefs.size() > 0 ) {
 			ColumnDef c = this.columnDefs.removeFirst();
 			// unable to choose a position in this form
-			alterStatement().columnMods.add(new AddColumnMod(c.getName(), c, null));
+			alterStatement().columnMods.add(new AddColumnMod(c.getName(), c, new ColumnPosition()));
```
dadah89: did `null` work before? seems to me it'd throw an exception when calling `index` on `null` (AddColumnMod apply method)

dadah89:
:+1:

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
osheroff: no, it didn't.  this code was all the way broken before.

Issue #223, opened 2015-12-23
Exception: Couldn't find table


jpeyper:
One of our more active testing environments is getting this error.

```
com.zendesk.maxwell.schema.ddl.SchemaSyncError: Couldn't find table: cgt.20151130_TSP_10291__CGPos_0_DF100961B3B9476FB02D495D50C37000
	at com.zendesk.maxwell.schema.ddl.TableAlter.apply(TableAlter.java:44)
	at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:296)
	at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:268)
	at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:89)
	at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:31)
	at com.zendesk.maxwell.Maxwell.run(Maxwell.java:91)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:97)
2015-12-23T13:50:13,450 INFO  com.zendesk.maxwell.schema.SchemaPosition: Storing final position: null
```

My theory was that this table is very short lived (should probably be a temporary table but that cant be helped).

I tried reproducing this by running these two commands quickly
```
CREATE TABLE `cgt`.`20151130_TSP_10291__CGPos_0_DF100961B3B9476FB02D495D50C37000` (`id` INT (11));

DROP TABLE `cgt`.`20151130_TSP_10291__CGPos_0_DF100961B3B9476FB02D495D50C37000`;
```

but this didn't cause any issues so I guess I'm wrong.

osheroff:
two commands quickly wouldn't do much; maxwell processes commands sequentially and shouldn't miss any DDL.  There's a possibility of a race condition when maxwell *first* captures schema; it doesn't take any locks, so if you create tables around the same time as the first boot of maxwell, that could maybe cause problems.  Or there could be something else.  Is there anything in the `maxwell.tables` table that looks like it might have screwed up the processing of the table creation?

jpeyper:
the maxwell.positions tables looks like this
```
"server_id"	"binlog_file"	"binlog_position"
"11543"	"sthx11543-bin.000007"	"669960990"
```
the query `SHOW BINLOG EVENTS IN "sthx11543-bin.000007" FROM 669957378 LIMIT 10;` (a few positions before 669960990) returns this
```
[...]

PR #230, opened 2015-12-28
Case sensitivity detection

osheroff:
here we read the value of the case sensitivity variable from the server and adapt maxwell accordingly, following the rules outlined in http://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html.

Addresses #223.

@zendesk/rules


vanchi-zendesk:
:+1: The code does look more complex now though.

jpeyper:
Oh hi,

Any change of getting this released soon?

PR #231, opened 2015-12-28
mysql-test, part 11

osheroff:
some further archaicism that means "use the ucs2 character set."

@zendesk/rules


dadah89:
:+1:

osheroff:
merged by hand

PR #232, opened 2015-12-28
remove org.json entirely

osheroff:
- teardown a couple of unused classes
- pull org.json dependency out.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #236, opened 2016-01-02
mysql-test, part 12

osheroff:
- handle alternative ordering of STORAGE x TABLESPACE y
@zendesk/rules


dadah89:
seems like TABLESPACE also supports [`ADD DATAFILE` and `FILE_BLOCK_SIZE`](http://dev.mysql.com/doc/refman/5.7/en/create-tablespace.html) (also `ENGINE`, but I suppose that works already). Is it something that you care about adding support now?

dadah89:
oh that is create table syntax... TABLESPACE seems fine. Surprised by the example with just STORAGE though, http://dev.mysql.com/doc/refman/5.6/en/create-table.html doesn't seem to define STORAGE followed by TABLESPACE or even just it by itself... if tests pass, does it mean mysql support it anyway? or am I looking the wrong doc?

osheroff:
yeah mysql supports quite a few undocumented SQL variants, and yeah, all the sql in `mysql-test-...` passed through mysql at some point in time.

I need to figure out why this branch fixed the cases where it's just `STORAGE DISK`, that doesn't seem right.

osheroff:
@dadah89 I made the PR a bit clearer.

dadah89:
:+1:

PR #237, opened 2016-01-02
mysql-test, part 14

osheroff:
MERGE MANUALLY.

handle the annoyingly large vocabulary of CHECK()

the problem:
mysql allows `CHECK(stmt)` constraints that have full sql statement capabilities inside the CHECK( .. ).  These statements do not a damn thing, but we need to parse them nonetheless, and the full language of what can go inside a `CHECK()` is huge and out-of-scope for maxwell.

the solution:

Our parser is good enough to recognize parenthesis and then a bunch of arbitrary tokens.  When we see that we've hit an unparsable section, we drop out of the parser and snip out matching parenthesis from the token stream, and then re-parse.  Note that we *could* technically do this with a lexer rule that changes grammar, but I'm more confidant about this at a parser level.

@zendesk/rules


osheroff:
@dadah89 did you miss this one?  I think it had a duplicate name

dadah89:
left for last b/c it had more than grammar changes =) will take a look later today

src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -56,4 +59,4 @@ fragment BACKTICK: '`';
 QUOTED_IDENT: BACKTICK (('``') | ~('`'))+ BACKTICK;
 
 fragment DIGIT: [0-9];
-WS  :   [ \t\n\r]+ -> skip ;
+WS  :   [ \t\n\r]+ -> channel(HIDDEN) ;
```
dadah89: what does `channel(HIDDEN)` mean?

osheroff: so when tokens (like our WS token here) are recognized, they are emitted to the parser in a stream.   Each token can also be attached to the notion of a "channel".  The parser is coded to only recognize the default channel, so if you want something ignored by the parser you can send the token to an alternative channel.  In this way the `TokenRewrite` stuff can keep all the whitespace in tact, but the parser can continue to ignore it.

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
```diff
@@ -55,36 +55,46 @@ private static boolean matchesBlacklist(String sql) {
...
-		} catch ( MaxwellSQLSyntaxError e) {
-			LOGGER.error("Error parsing SQL: '" + sql + "'");
-			throw (e);
 		}
 
+		while ( true ) {
+			try {
+				return parseSQL(currentDB, sql);
+			} catch ( ReparseSQLException e ) {
+				LOGGER.debug("rewrote SQL to " + sql);
```
dadah89: shouldn't the log line come after the stmt bellow?

src/main/antlr4/imports/mysql_idents.g4
dadah89: So TokenStream walks off channel tokens as well... but if everything is getting replaced with `___MAXWELL___` (including spaces), does it matter at all?

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
osheroff: yeah good point.

src/main/antlr4/imports/mysql_idents.g4
osheroff: yeah it does matter, because the rewriter is processing the entire stream, not just the part it's rewriting.  so if you look at how we rewrite, we add directives like "delete the tokens 5 through 12.  add __MAXWELL__ at position 5.  Now give me the entire sql document again."  And the entire SQL document isn't valid unless we preserve whitespace.

dadah89:
Read quite a bit about parsers/lexers and atlr4, enough for me to be comfortable understanding the changes here. Looked a bit at lexer solutions (at first I thought it'd use `mode`, then I thought it was wrong, but now I think it'd use it...), but the best ones seems to involve injecting some java code within the grammars, so the TokenStreamRewriter solution does something similar and seems cleaner (put the code in a separate place). Hope my understanding is good enough... changes look good to me, :+1:

osheroff:
yeah, using a lexer-mode is the other big solution here; I was just nervous that one might come across the CHECK token in a different context (tables can't be named `CHECK`, but they can be named the same as other tokens), and it felt saner on some level to let the parser handle it.  Sounds like you've got a handle on it.

osheroff:
merged.

src/main/antlr4/imports/mysql_idents.g4
dadah89: oh I see, was expecting it to ouput a new stream that you could re-instantiate the parser with, but the rewrite could've added new tokens and lexer rules, makes sense (kinda hard finding complete examples with TokenStreamRewrite and parser)

osheroff: that was actually what I was expecting too, but they let you replace a token with a string.  In theory one could build a more limited rewriter that only allows deletion of tokens or adding tokens (not strings) that would be able to re-output a TokenStream

PR #238, opened 2016-01-02
mysql-test, part 15

osheroff:
MERGE MANUALLY

- handle table names and ASC/DESC in `ALTER TABLE ORDER BY` ddl

@zendesk/rules

### DIFF
--- a/src/main/antlr4/imports/mysql_alter_table.g4
+++ b/src/main/antlr4/imports/mysql_alter_table.g4
@@ -40,7 +40,7 @@ ignored_alter_specifications:
     | DROP INDEX index_name
     | DISABLE KEYS
     | ENABLE KEYS
-    | ORDER BY index_columns
+    | ORDER BY alter_ordering (',' alter_ordering)*
     | FORCE
     /*
      I'm also leaving out the following from the alter table definition because who cares:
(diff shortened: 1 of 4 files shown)


src/main/antlr4/imports/mysql_alter_table.g4
```diff
@@ -65,3 +65,9 @@ ignored_alter_specifications:
     ;
   algorithm_type: DEFAULT | INPLACE | COPY;
   lock_type: DEFAULT | NONE | SHARED | EXCLUSIVE;
+
+alter_ordering: alter_ordering_column (ASC|DESC)?;
+alter_ordering_column:
+    name '.' name '.' name
```
dadah89: would `table_name '.' name | name;` work here? does it matter adding a test for that case?

dadah89:
:+1:

src/main/antlr4/imports/mysql_alter_table.g4
osheroff: not quite, there's three forms of column names that are *technically* acceptable:

```
`col`
`tbl`.`col`
`db`.`col`
```

In this case, it doesn't matter if I give them aliased names, as I'm parsing and throwing away all of this stuff.

dadah89: thought that table_name would take care of tbl and db.tbl, putting the logic in one place... think it's a small thing though, pr looks just fine to me

osheroff:
merged.

PR #239, opened 2016-01-02
mysql-test, part 16

osheroff:
MERGE MANUALLY

- add `ALTER USER` to ddl blacklist
- move GIS-extension tests into their own file

@zendesk/rules


dadah89:
```
ALTER USER was added in MySQL 5.6.6. However, in 5.6.6, ALTER USER also sets the Password column to the empty string, so do not use this statement until 5.6.7.
```
scary =)
:+1:

osheroff:
merged.

PR #240, opened 2016-01-02
mysql-test, part 17

osheroff:
MERGE MANUALLY

- somehow missed NULL as a default value.  scary. 
- allow `REFERENCES( .. )` as a column creation option

@zendesk/rules


src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -12,7 +12,7 @@ user_token: (IDENT | QUOTED_IDENT | string_literal);
 
 name: ( id | tokens_available_for_names | INTEGER_LITERAL);
 id: ( IDENT | QUOTED_IDENT );
-literal: (float_literal | integer_literal | string_literal);
+literal: (float_literal | integer_literal | string_literal | NULL);
```
dadah89: does SET DEFAULT support the same values as DEFAULT? https://github.com/zendesk/maxwell/blob/928451d808d7313df92a335dc1c1738e43db3ab0/src/main/antlr4/imports/column_definitions.g4#L90

src/main/antlr4/imports/mysql_indices.g4
```diff
@@ -47,9 +47,7 @@ index_column_asc_or_desc: ASC | DESC;
 reference_definition:
   REFERENCES table_name
   index_column_list
-  reference_definition_match?
-  reference_definition_on_delete?
-  reference_definition_on_update?
+  ( reference_definition_match | reference_definition_on_delete | reference_definition_on_update)*
```
dadah89: parenthesis spacing doesn't match =)

dadah89:
left few comments, besides that :+1:

src/main/antlr4/imports/mysql_idents.g4
osheroff: it doesn't.  it allows all literals (including TRUE/FALSE, which are not represented here), but not `NOW()` or `CURRENT_TIMESTAMP`

Issue #225, opened 2015-12-23
Change Maxwell's Schema


jpeyper:
This came out of the PR: https://github.com/zendesk/maxwell/pull/222

Currently the `maxwell` schema is hard-coded in a number of places. Being able to specify the database name would allow for many different setups, including (but not limited to)
- running multiple instances of Maxwell against the same MySQL instance, tracking different tables and posting them to different kafka topics
- with the linked PR, multiple Maxwell instances all storing their schemas on the same MySQL instance, but having different replication hosts

PR #242, opened 2016-01-06
Allow custom maxwell database name

kristiankaufmann:
resolves zendesk/maxwell#225

tests still pending.


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -59,6 +60,7 @@ private OptionParser getOptionParser() {
 		parser.accepts( "output_file", "output file for 'file' producer" ).withRequiredArg();
 		parser.accepts( "password", "mysql password" ).withRequiredArg();
 		parser.accepts( "port", "mysql port" ).withRequiredArg();
+		parser.accepts( "database", "database name for maxwell state (schema and binlog position)").withRequiredArg();
```
osheroff: maybe named `schema_database` just to be ultra-clear about intentions?

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
```diff
@@ -85,7 +87,7 @@ private void store(BinlogPosition newPosition) {
 		if ( newPosition == null )
 			return;
 
-		String sql = "INSERT INTO `maxwell`.`positions` set "
+		String sql = "INSERT INTO `" + this.schemaDatabaseName + "`.`positions` set "
```
osheroff: all of these look like potential SQL injection sites to me.  Maybe we instead of substituting in the dbName we strip off the explicit db name and connect the `ConnectionPool` to a particular db?  (if that's possible...)

src/main/resources/sql/maxwell_schema.sql
```diff
@@ -1,34 +1,58 @@
...
+SET @c = CONCAT('CREATE DATABASE IF NOT EXISTS ', @maxwell_db)
 
-CREATE TABLE IF NOT EXISTS `maxwell`.`schemas` (
+PREPARE maxwell_stmt from @c
+
+EXECUTE maxwell_stmt
+
+DEALLOCATE PREPARE maxwell_stmt
+
+SET @c = CONCAT('CREATE TABLE IF NOT EXISTS `', @maxwell_db, '`.`schemas` (
```
osheroff: I think the same solution should stand here:  have the calling code perform the `create database if not exists` and then execute the sql file under a "USE db" style connection.

src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
kristiankaufmann: looks like I could specify a default database in the jdbc connection URI.

the other option would be to use connection.setCatalog to specify the db.

osheroff: either way is fine.  `setCatalog` may be easier from a standpoint of not having to do two connections (one for the `CREATE DATABASE`, the other for the usage)

src/main/resources/sql/maxwell_schema.sql
kristiankaufmann: I am having trouble finding away to avoid the following
```
String sql = "CREATE DATABASE IF NOT EXISTS " + dbName;
conn.createStatement.execute(sql)
```

One option would be to put the onus on the user to ensure the database has been created.

osheroff: I think it's ok.  We can quote the database name properly ( in backticks ), and `CREATE DATABASE` shouldn't be open to sql-injection in the same way that `INSERT INTO` or `SELECT * FROM` is.

kristiankaufmann: I have setCatalog working, but I ran into an odd problem when running the tests.

When I run the tests some of the them would fail with a no database selected SQL Exception. However when I rerun just the failed test, it would pass. It turned out that CachedStatements didn't maintain the database from setCatalog call. If I disable caching on the connection pool then the tests pass except for the case sensitivity test.

I am going to merge master into this branch before proceeding since my previously merged PR will conflict with this branch.

osheroff: disabling caching is totally fine fine fine.

kristiankaufmann:
I have this working. It would benefit from a couple other people trying it out to make sure I haven't missed anything though. I also probably should add some tests.

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -74,7 +74,8 @@ private void updateInsertedRowsColumn(int insertedRows, RowMap startBootstrapRow
 		long now = System.currentTimeMillis();
 		if ( now - lastInsertedRowsUpdateTimeMillis > INSERTED_ROWS_UPDATE_PERIOD_MILLIS ) {
 			long rowId = ( long ) startBootstrapRow.getData("id");
-			String sql = "update maxwell.bootstrap set inserted_rows = ?, binlog_file = ?, binlog_position = ? where id = ?";
+			String sql = "update `bootstrap` set inserted_rows = ?, binlog_file = ?, binlog_position = ? where id = ?";
+			connection.setCatalog(this.context.getConfig().databaseName);
```
osheroff: hmmm, we really shouldn't have to profilerate `setCatalog` calls out everywhere -- better to just to do a `setCatalog` on the connection once, after we check it out of the pool... unless there's something here I'm missing?

kristiankaufmann:
@osheroff I've consolidated the setCatalog calls.

osheroff:
great, thanks kristian.  Looks good after a surface-read, I'll do a proper deep CR on Tuesday.

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtility.java
```diff
@@ -63,8 +63,8 @@ public void run() {
 		}
 	}
 
-	private int getInsertedRowsCount(Connection connection, long rowId) throws SQLException {
-		String sql = "select inserted_rows from maxwell.bootstrap where id = ?";
+	private int getInsertedRowsCount(Connection connection, long rowId, String dbName) throws SQLException {
```
osheroff: unused parameters here and throughout this file...

osheroff:
other than those minor comments, we're good to go here.  I may refactor a tiny bit more post-merge, but it generally looks good!

kristiankaufmann:
@osheroff Let me know if there is anything else I missed or that you would like me to modify.

kristiankaufmann:
@osheroff As I mentioned on the other PR. let me know if you want a new PR with a cleaner commit history

osheroff:
nah I don't care about clean history.   I would like the docs updated with the new command line flags.  Let me know if you want me to take that part.

kristiankaufmann:
I'll update the docs

mark800:
Looking forward to this new feature, it will be very helpful!

kristiankaufmann:
@osheroff this is ready

osheroff:
:+1:

PR #243, opened 2016-01-08
GIS extensions support for maxwell

osheroff:
open-replicator ingests from mysql's WKB format, and we re-output it in WKT (well-known text) format.  Why everything is so well known is a bit of a mystery.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #249, opened 2016-01-19
Configurable kafka message partitioning

kristiankaufmann:
adds ability to partition by table or primary key.
adds option of using murmurhash3 as hashing function. String.hashCode is not recommended as there is no guarantee that Java will continue to hash the same in a distributed system see https://martin.kleppmann.com/2012/06/18/java-hashcode-unsafe-for-distributed-systems.html


src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -74,11 +76,23 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
...
+				break;
+			case "murmur3|table": this.partitioner = new Murmur3KafkaTablePartitioner(this.context.getConfig().murmur3Seed);
+				break;
+			case "murmur3|primary_key": this.partitioner = new Murmur3KafkaPrimaryKeyPartitioner(this.context.getConfig().murmur3Seed);
+				break;
+			case "default|primary_key": this.partitioner = new MaxwellKafkaPrimaryKeyPartitioner();
+				break;
+			case "default|table": this.partitioner = new MaxwellKafkaTablePartitioner();
+				break;
+			case "default|database":
```
osheroff: It'd probably be simpler to have a single `KafkaPartitioner` class that was instantiated with some options:

```
  this.partitioner = new MaxwellKafkaPartitioner(PARTITION_BY_DATABASE, HASH_MURMER3, murmer3SeedValue)
```

Or you could separate the responsibility into 
1. a group of classes that choose a string for partitioning
2. a group of classes that hashes that string into an integer

either way, it doesn't make sense to have a dedicated class for each string/hash function combination.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -143,6 +149,15 @@ private void parse(String [] argv) {
...
 			this.kafkaTopic = (String) options.valueOf("kafka_topic");
 
+		if ( options.has("kafka_partition_by"))
+			this.kafkaPartitionKey = (String) options.valueOf("kafka_partition_by");
+
+		if ( options.has("kafka_partition_hash"))
+			this.kafkaPartitionHash = (String) options.valueOf("kafka_partition_hash");
+
+		if ( options.has("kafka_murmur3_seed"))
+			this.murmur3Seed = Integer.valueOf((String) options.valueOf("kafka_murmur3_seed"));
```
osheroff: is there a rationale for making this user-configurable?  I could definitely see documenting the seed value so you could reproduce it elsewhere, not totally sure I see the need to configure it though....

kristiankaufmann: Where should I document the seed? The only time I can for see wanting to change it would be if there was a unfavorable distribution for the partitioning. This unlikely to occur so I'll just hard code the seed.

kristiankaufmann:
@osheroff take another look. Let me know if you want me to create another branch so that there is a cleaner commit history

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: we should talk about the hash function and partition-by in the kafka documentation.  LMK if you want to do or if I should take it

osheroff:
yeah, don't care about clean commit history.  Some documentation would be good tho

kristiankaufmann:
I'll add documentation

kristiankaufmann:
@osheroff this is ready

osheroff:
:+1:

PR #253, opened 2016-01-25
add --blacklist_tables command-line option

nmaquet:
Add `--blacklist_tables PATTERN` command-line option.

The option `blacklist_tables` controls whether Maxwell will send updates for a table to its producer AND whether it captures schema changes for that table. Note that once Maxwell has been running with a table marked as blacklisted, you *must* continue to run Maxwell with that table blacklisted or else Maxwell will halt. If you want to stop blacklisting a table, you will have to drop the maxwell schema first. Also note that with table blacklisting enabled, Maxwell cannot distinguish between row updates for blacklisted tables and row updates for unknown tables; it will just assume that all row updates for unknown tables relate to blacklisted tables.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
@@ -132,14 +132,17 @@ public class MaxwellReplicator extends RunLoopProcess {
 		MaxwellAbstractRowsEvent ew;
 		Table table;
 
-		table = tableCache.getTable(e.getTableId());
+		long tableId = e.getTableId();
 
-		if ( table == null && this.filter.hasTableBlacklist() ) {
-			LOGGER.warn(String.format("couldn't find table in cache for table id: %d, assuming blacklisted table", e.getTableId()));
+		if ( tableCache.isTableBlacklisted(tableId) ) {
+			LOGGER.debug(String.format("ignoring row event for blacklisted table %s", tableCache.getBlacklistedTableName(tableId)));
 			return null;
 		}
-		else if ( table == null ) {
-			throw new SchemaSyncError("couldn't find table in cache for table id: " + e.getTableId());
+
+		table = tableCache.getTable(tableId);
+
+		if ( table == null ) {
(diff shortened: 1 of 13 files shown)


src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -134,7 +134,11 @@ private MaxwellAbstractRowsEvent processRowsEvent(AbstractRowEvent e) throws Sch
 
 		table = tableCache.getTable(e.getTableId());
 
-		if (table == null) {
+		if ( table == null && this.filter.hasTableBlacklist() ) {
+			LOGGER.warn(String.format("couldn't find table in cache for table id: %d, assuming blacklisted table", e.getTableId()));
+			return null;
```
osheroff: hmmm, so this is a tick unelegant.  how about this approach instead:

- in TableMapCache, maintain a separate `Set` of table ids that are blacklisted.  Then here you can definitely ask `tableCache.isBlacklisted(e.getTableId)` and leave the `SchemaSyncError` in place

nmaquet: Yeah, I thought you might say that :)
If you want to remember the mapping of `tableId` to `tableName`, then you'd have to persist it in the `maxwell` schema, correct? An in-memory `Set` wouldn't be enough I think.

osheroff: ah, no, you misunderstand a bit.  Each transaction is preceded by a TABLE_MAP_EVENT which will contain the database name + table name.

nmaquet: Cheers! I'll make the change shortly.

osheroff: (as well as the table-id) -- that's what it's mapping, id -> table_name (and some schema info)

nmaquet:
@osheroff added blacklisted table list to `MaxwellTableCache`

osheroff:
good deal.  :+1:

Issue #254, opened 2016-01-28
Error parsing SQL: '/*!50003 DROP FUNCTION IF EXISTS `DAY_NAME_FROM_NUMER` */'


nmaquet:
Maxwell crashed with the following error today (1.0-PRE1):

```
04:31:33,645 ERROR MysqlParserListener - (statement DROP FUNCTION IF EXISTS `DAY_NAME_FROM_NUMER`)
04:31:33,646 ERROR SchemaChange - Error parsing SQL: '/*!50003 DROP FUNCTION IF EXISTS `DAY_NAME_FROM_NUMER` */'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: DROP
        at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:96)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parseSQL(SchemaChange.java:73)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:85)
        at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:310)
        at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:290)
        at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:101)
        at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:31)
        at com.zendesk.maxwell.Maxwell.run(Maxwell.java:100)
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:106)
04:31:33,647 INFO  SchemaPosition - Storing final position: null
```

osheroff:
k.  I think the solution is to use a cheap regex to unwrap sql from comments before checking against the blacklist.  Can hopefully look today.

osheroff:
fixed in 1.0.0-PRE2

PR #256, opened 2016-01-28
blacklist should un-comment SQL before checking blacklist

osheroff:
addresses #254

@zendesk/rules @nmaquet


nmaquet:
:+1: Thanks for fixing this.

vanchi-zendesk:
:+1:

Issue #255, opened 2016-01-28
Error Parsing Collate in Table Create Statement


danfiza:
Received the following parse error today on the latest Maxwell Release(1.0.0)

```
14:01:43,455 DEBUG SchemaChange - SQL_PARSE <- "/* THIS SCRIPT CREATES A NEW TABLE TO  LOG UPDATES IN CASE WE NEED TO RUN IT AGAIN. */
CREATE TABLE IF NOT EXISTS `test`.`test_logs` (
	`id` 		INT(11) 	AUTO_INCREMENT PRIMARY KEY, 
	`zip`		VARCHAR(10)	CHARACTER SET `latin1` 		COLLATE `latin1_swedish_ci`	DEFAULT '' 	NOT NULL, 
	`date`		DATE						NOT NULL, 
	`data` 		BLOB						NULL, 
	`created_by`	INT(11) 	DEFAULT 0 			NOT NULL, 
	`created_date` 	DATETIME 	DEFAULT CURRENT_TIMESTAMP	NOT NULL 
) ENGINE = INNODB CHARACTER SET = `latin1` COLLATE = `latin1_swedish_ci`"
14:01:44,074 ERROR MysqlParserListener - (string `latin1_swedish_ci`)
14:01:44,075 ERROR SchemaChange - Error parsing SQL: '/* THIS SCRIPT CREATES A NEW TABLE TO  LOG UPDATES IN CASE WE NEED TO RUN IT AGAIN. */
CREATE TABLE IF NOT EXISTS `test`.`test_logs` (
	`id` 		INT(11) 	AUTO_INCREMENT PRIMARY KEY, 
	`zip`		VARCHAR(10)	CHARACTER SET `latin1` 		COLLATE `latin1_swedish_ci`	DEFAULT '' 	NOT NULL, 
	`date`		DATE						NOT NULL, 
	`data` 		BLOB						NULL, 
	`created_by`	INT(11) 	DEFAULT 0 			NOT NULL, 
	`created_date` 	DATETIME 	DEFAULT CURRENT_TIMESTAMP	NOT NULL 
) ENGINE = INNODB CHARACTER SET = `latin1` COLLATE = `latin1_swedish_ci`'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: `latin1_swedish_ci`
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:96)
[...]

PR #258, opened 2016-01-29
deal with backtick quoted identifiers in table creation

osheroff:
I had fixed this in ALTER TABLE landia, but needed to in regular landia too.

addresses https://github.com/zendesk/maxwell/issues/255

@zendesk/rules


dadah89:
I'm assuming collate is already being processed correctly, change seems fine to me, :+1:

Issue #262, opened 2016-02-03
Parse Error


jpeyper:
I'm still running on v0.17.0, but I got this parse error today:

```
2016-02-03T11:12:42,878 ERROR com.zendesk.maxwell.schema.ddl.SchemaChange: Error parsing SQL: '-- Refresh the mysql access changes
FLUSH PRIVILEGES'
```

Has this been fixed by one of the v1.0.0-PRE releases?

osheroff:
no, not yet.  it's similar to another one around SQL comments and the
blacklist.  will get a patch in for the next PRE release

jpeyper:
Cool.

osheroff:
fixed in 1.0.0-RC1

PR #263, opened 2016-02-04
handle blacklisted statements with comments

osheroff:
addresses #262

@zendesk/rules


dadah89:
The multiline matching looks a bit scary, but MySQL seems to accept all sorts of syntax...
lgtm, :+1:

PR #266, opened 2016-02-09
Bootstrap tweaks, part 3 of 3

osheroff:
ah, the sauce. 

Rework how bootstrap command line works

The main thrust here is that we no longer cancel a bootstrap on CTRL-C
-- in my tests it was too easy for the bootstrap utility to get a SIGHUP
because you didn't launch it under screen, and having that kill off the
bootstrap seems to violate principle-of-least-surprise.

Other changes that came along to support this:
- add --abort ID and --monitor ID to cancel/resume bootstrap operations
- count rows to be bootstrapped once, and store that number in the table
- the bootstrap utility now picks up config.properties

@zendesk/rules @nmaquet

### DIFF
--- a/src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtility.java
+++ b/src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtility.java
@@ -22,40 +22,28 @@ public class MaxwellBootstrapUtility {
 
 	private void run(String[] argv) throws Exception {
 		MaxwellBootstrapUtilityConfig config = new MaxwellBootstrapUtilityConfig(argv);
+
 		if ( config.log_level != null ) {
 			MaxwellLogging.setLevel(config.log_level);
 		}
+
 		ConnectionPool connectionPool = getConnectionPool(config);
 		try ( final Connection connection = connectionPool.getConnection() ) {
-			int rowCount = getRowCount(connection, config);
-			final long rowId = insertBootstrapStartRow(connection, config);
-			Runtime.getRuntime().addShutdownHook(new Thread() {
-				public void run() {
-					try {
-						if ( !isComplete ) {
-							displayLine("");
-							LOGGER.warn("bootstrapping cancelled");
-							removeBootstrapRow(connection, rowId);
-						}
-					} catch ( Exception e ) {
-						System.exit(1);
-					}
-				}
-			});
-			int insertedRowsCount = 0;
-			Long startedTimeMillis = null;
-			while ( !isComplete ) {
-				if ( insertedRowsCount < rowCount ) {
-					if ( startedTimeMillis == null && insertedRowsCount > 0 ) {
-						startedTimeMillis = System.currentTimeMillis();
-					}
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtilityConfig.java
```diff
@@ -33,16 +36,31 @@ public String getConnectionURI( ) {
 
 	protected OptionParser buildOptionParser() {
 		OptionParser parser = new OptionParser();
+		parser.accepts( "config", "location of config file" ).withRequiredArg();
+		parser.accepts( "__separator_1", "" );
+		parser.accepts( "database", "database that contains the table to bootstrap").withRequiredArg();
+		parser.accepts( "table", "table to bootstrap").withRequiredArg();
+		parser.accepts( "__separator_2", "" );
+		parser.accepts( "abort", "bootstrap_id to abort" ).withRequiredArg();
+		parser.accepts( "monitor", "bootstrap_id to abort" ).withRequiredArg();
```
dadah89: abort -> monitor

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtility.java
```diff
@@ -22,40 +22,28 @@
...
-				displayProgress(rowCount, insertedRowsCount, startedTimeMillis);
-				Thread.sleep(UPDATE_PERIOD_MILLIS);
+			if ( config.abortBootstrapID != null ) {
+				removeBootstrapRow(connection, config.abortBootstrapID);
+				return;
+			}
+
+			long rowId;
+			if ( config.monitorBootstrapID != null )
+				rowId = config.monitorBootstrapID;
```
dadah89: would be nice to show an error and return if you're trying to monitor a rowId that doesn't exist

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtility.java
```diff
@@ -63,13 +51,52 @@ public void run() {
...
+
+		insertedRowsCount = initialRowCount;
+		while ( !isComplete ) {
+			if ( insertedRowsCount < rowCount ) {
+				if ( startedTimeMillis == null && insertedRowsCount > 0 ) {
+					startedTimeMillis = System.currentTimeMillis();
+				}
+				insertedRowsCount = getInsertedRowsCount(connection, rowId);
+			}
+			isComplete = getIsComplete(connection, rowId);
```
dadah89: will it ever leave this loop when rowId is removed?

osheroff: yeah, previously it threw a weird `SQLException`, now it throws something quite a bit nicer.

osheroff:
@dadah89 all concerns addressed.

dadah89:
:+1:

Issue #272, opened 2016-02-11
ERROR SchemaChange - Error parsing SQL: 'ALTER TABLE .`users`


nmaquet:
Maxwell is crashing with the following error today. Restarting it doesn't solve the problem unfortunately.

```
01:55:22,485 ERROR MysqlParserListener - (table_name . `users`)
01:55:22,485 ERROR SchemaChange - Error parsing SQL: 'ALTER TABLE .`users`
CHANGE COLUMN `password` `password` VARCHAR(60) CHARACTER SET 'utf8' COLLATE 'utf8_bin' NULL DEFAULT NULL COMMENT 'Length 60 for Bcrypt' ,
ADD COLUMN `password_scheme` ENUM('md5','bcrypt') NULL DEFAULT 'md5' AFTER `password`'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: .
        at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:96)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parseSQL(SchemaChange.java:75)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:87)
        at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:310)
        at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:290)
        at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:101)
        at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:31)
        at com.zendesk.maxwell.Maxwell.run(Maxwell.java:103)
[...]

PR #273, opened 2016-02-11
fix parsing table names with a leading '.'

osheroff:
addresses https://github.com/zendesk/maxwell/issues/272

### DIFF
--- a/src/main/antlr4/imports/mysql_idents.g4
+++ b/src/main/antlr4/imports/mysql_idents.g4
@@ -4,6 +4,7 @@ import mysql_literal_tokens;
 
 db_name: name;
 table_name: (db_name '.' name)
+            | '.' name // *shakes head in sadness*
             | name
             ;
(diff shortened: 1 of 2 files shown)


osheroff:
@zendesk/rules @nmaquet

src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -4,6 +4,7 @@ import mysql_literal_tokens;
 
 db_name: name;
 table_name: (db_name '.' name)
+            | '.' name // *shakes head in sadness*
```
nmaquet: :joy:

osheroff: the sins of our father-engineers will snowball into massive technical debts one day.  If they haven't already.

vanchi-zendesk:
:+1:

PR #274, opened 2016-02-14
fix ALTER DATABASE character set processing

osheroff:
@zendesk/rules


src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
```diff
@@ -78,7 +78,7 @@ public void testAlter() throws SQLException, SchemaSyncError, IOException, Inter
 
 	@Test
 	public void testAlterDatabase() throws Exception {
-		String sql[] = {
+	String sql[] = {
```
vanchi-zendesk: indentation

vanchi-zendesk:
No tests for where the failure is?

osheroff:
@vanchi-zendesk adding the `diff` method caused some existing test to fail which I then  fixed.

vanchi-zendesk:
:+1:

PR #277, opened 2016-02-16
refactor ColumnDef classes

osheroff:
In preparation for serializing ColumnDef classes, we move them to a more "proper" hierarchy in which only string columns have a character set, only enum columns have enum values, and only integer columns have a signed flag.

There's still some type-checking hacks that it'd be good to remove by delegating the methods to the objects, but this is good for now.

@zendesk/rules 

NOTE: this is a PR off field_renames_2.  merge appropriately.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/columndef/DecimalColumnDef.java
+++ b/src/main/java/com/zendesk/maxwell/schema/columndef/DecimalColumnDef.java
@@ -5,8 +5,8 @@ import java.math.BigDecimal;
 import com.google.code.or.common.util.MySQLConstants;
 
 public class DecimalColumnDef extends ColumnDef {
-	public DecimalColumnDef(String tableName, String name, String type, int pos) {
-		super(tableName, name, type, pos);
+	public DecimalColumnDef(String name, String type, int pos) {
+		super(name, type, pos);
 	}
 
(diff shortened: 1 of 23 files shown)


src/main/java/com/zendesk/maxwell/schema/columndef/DecimalColumnDef.java
```diff
@@ -21,4 +21,9 @@ public String toSQL(Object value) {
 		return d.toEngineeringString();
 	}
 
+	@Override
+	public ColumnDef copy() {
+		return new DecimalColumnDef(name, type, pos);
+	}
```
vanchi-zendesk: ugh. This is weird. No obvious other ways of doing this?

vanchi-zendesk:
:+1:

src/main/java/com/zendesk/maxwell/schema/columndef/DecimalColumnDef.java
osheroff: you mean repeating this endlessly?  Yeah, I mean you can play horrible tricks with reflection, but as with most things in javaland it's better to be verbose and clear than terse and horrible

osheroff:
merged by hand

PR #279, opened 2016-02-16
fix CONVERT TO CHARSET

osheroff:
we were parsing this fine, but it turns out we were throwing it away.
boooooo.

Note that this is sitting on top of a refactor.

@zendesk/rules


vanchi-zendesk:
:+1:

PR #283, opened 2016-02-22
Add bootstrap progress on stdout every 100k inserts on non-TTY

nmaquet:
Add bootstrap progress output on stdout every 100k inserts even when the process has no TTY (e.g. when running inside [Rundeck](http://rundeck.org/) or [Jenkins](https://jenkins-ci.org/)).


osheroff:
:+1:

PR #288, opened 2016-02-26
change Table#database into a String holder

osheroff:
I'm prepping for being able to serialize the Table object, as part of a bigger, seemingly endless refactor.

@zendesk/rules


dadah89:
:+1:

Issue #260, opened 2016-02-01
blacklisted table question


jpeyper:
For a lack of somewhere better to ask this...

Does this setting allow be to blacklist entire schemas or just tables?

For example, can I do
`--blacklist_tables=/__temp.*/`
to blacklist everything in the `__temp` scheme?

edit:

I'm not actually in a position to test this right now so asking the question.

osheroff:
@jpeyper, no, it just deals with table names.  will have to add a corresponding option to filter out whole databases.  

cc @nmaquet

osheroff:
fixed in https://github.com/zendesk/maxwell/pull/289

PR #289, opened 2016-02-28
Add ability to blacklist entire databases

jpeyper:
Similar to the blacklist_tables option, this new setting will allow entire databases to be blacklisted.

This should address #260

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellTableCache.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellTableCache.java
@@ -1,7 +1,6 @@
 package com.zendesk.maxwell;
 
 import java.util.HashMap;
-import java.util.HashSet;
 
 import com.google.code.or.binlog.impl.event.TableMapEvent;
 import com.zendesk.maxwell.schema.Database;
--- a/src/main/java/com/zendesk/maxwell/MaxwellTableCache.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellTableCache.java
@@ -17,17 +17,19 @@ public class MaxwellTableCache {
 			String dbName = new String(event.getDatabaseName().getValue());
 			String tblName = new String(event.getTableName().getValue());
 			Database db = schema.findDatabase(dbName);
-			if ( db == null )
-				throw new RuntimeException("Couldn't find database " + dbName);
-
-			Table tbl = db.findTable(tblName);
 
 			if ( filter != null && filter.isTableBlacklisted(dbName, tblName) )
 				blacklistedTableCache.put(tableId, tblName);
-			else if ( tbl == null )
-				throw new RuntimeException("Couldn't find table " + tblName);
-			else
-				tableMapCache.put(tableId, tbl);
+			else if ( db == null )
+				throw new RuntimeException("Couldn't find database " + dbName);
+			else {
+				Table tbl = db.findTable(tblName);
+
+				if (tbl == null)
+					throw new RuntimeException("Couldn't find table " + tblName);
+				else
+					tableMapCache.put(tableId, tbl);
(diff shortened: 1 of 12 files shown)


jpeyper:
@osheroff My first PR for you. Hopefully I've followed the you're protocols / styles properly :)

jpeyper:
This should address #289

src/main/java/com/zendesk/maxwell/MaxwellTableCache.java
```diff
@@ -1,6 +1,7 @@
 package com.zendesk.maxwell;
 
 import java.util.HashMap;
+import java.util.HashSet;
```
nmaquet: Unused import?

nmaquet:
Cool ! This will be useful :+1:

osheroff:
lgtm, fix the unused import and I'll merge.  thanks!

jpeyper:
@osheroff done.

src/main/java/com/zendesk/maxwell/MaxwellTableCache.java
```diff
@@ -22,7 +23,7 @@ public void processEvent(Schema schema, MaxwellFilter filter, TableMapEvent even
 
 			Table tbl = db.findTable(tblName);
 
-			if ( filter != null && filter.isTableBlacklisted(tblName) )
+			if ( filter != null && filter.isTableBlacklisted(tbl.getDatabase().getName(), tblName) )
```
osheroff: @jpeyper my eyes sort of glossed over this line, but in local testing I found that it's subject to crashing -- if the database is blacklisted, it's quite possible that `db.findTable()` will return `null` (if we didn't process the `CREATE TABLE` ddl), or the call to `schema.findDatabase` may even return null (if we didn't process the `CREATE DATABASE` ddl).  

Can you do an integration test that sets up a database blacklist and then runs through a few scenarios?  notably:

  - create DB
  - capture schema
  - blacklist DB
  - create table DB.tbl
  - insert into DB.tbl

and also

  - capture schema
  - blacklist DB
  - create DB
  - create tbl
  - insert into DB.tbl

jpeyper: I'm developing in windows, and I get errors when running you're test suite... but i did just notice that just above that line there is already a `dbName` variable in scope (on line 17).

So I've switched to just using that.

osheroff: that will handle the case where the table has not been tracked inside the database, but not the case where the database has not been tracked

jpeyper: Ok, I think I see now.

Still not sure how to do the integration tests, but I think the change I've just made should cover the crash cases you've seen.

osheroff: ok.  I'll add an integration test for the cases before I merge.

jpeyper:
Was there anything else that needs addressing?

osheroff:
nope, all good.  merged into master.

osheroff:
(after adding a couple of tests)

wushujames:
Congrats to @jpeyper on his first PR to Maxwell!

jpeyper:
@osheroff Any idea when this will get into a release / RC?

osheroff:
@jpeyper I'm trying to stabilize maxwell right now for a big-ol 1.0 release (just picking up bugfixes), but I'm happy to do a 1.1-PRE release with whatever's in master if that'd help.

osheroff:
https://github.com/zendesk/maxwell/releases/tag/v1.1.0-PRE1

PR #290, opened 2016-03-01
DDL Serialization

osheroff:
I wanted to split this PR more before I sent it out, but I'm starting to reach the limits of what I can usefully isolate out.  Here's the premise of what's going on:

- I want to make processing and storing schema updates faster
- to that end, I want to capture a baseline schema and then just store delta-updates to it instead of re-storing massive schemas each time. 
- to that end, I want schema changes to serialize to JSON properly
- to that end, I've changed some stuff.  

Schema changes are now applied with a two-step process:
1. We "resolve" the DDL update.  This involves stuff like taking "default" character sets and making them explicit, changing from "create table like other_table" to "create table ( ... )", and totally removing schema changes that don't actually apply ("create table if not exists existing_table").  After a schema change is resolved, it should be able to be deserialized and make changes to a schema with no external state.
2. We apply the resolved DDL update.

Note that there will be two future PRs that depend on this one.  One will output the DDL-as-JSON to the kafka bus, the other will start storing schemas in the new format.

I could *possibly* split apart DDL-resolving and resolved-serialization into 2 PRs if code reviewers found that helpful... LMK.

@zendesk/rules


src/main/java/com/zendesk/maxwell/schema/ddl/TableCreate.java
```diff
@@ -28,44 +30,48 @@ public TableCreate (String database, String table, boolean ifNotExists) {
...
-		t.rename(this.table);
-		d.addTable(t);
+	private void resolveCharsets(String dbCharset, Table resolved) {
+		if ( this.charset != null )
+			resolved.charset = this.charset;
+		else
+			// inherit charset from database
+			resolved.charset = dbCharset;
+
+		for ( StringColumnDef c : resolved.getStringColumns() ) {
```
dadah89: can replace with `resolved.setDefaultColumnCharsets();`

src/main/java/com/zendesk/maxwell/schema/ddl/TableCreate.java
```diff
@@ -28,44 +30,48 @@ public TableCreate (String database, String table, boolean ifNotExists) {
...
-		if ( sourceTable == null )
-			throw new SchemaSyncError("Couldn't find table " + likeDB + "." + likeTable);
+		return copiedTable;
+	}
 
-		Table t = sourceTable.copy();
-		t.rename(this.table);
-		d.addTable(t);
+	private void resolveCharsets(String dbCharset, Table resolved) {
+		if ( this.charset != null )
```
dadah89: is this needed? `resolved.getCharset()` already defaults to `database.getCharset()`

src/main/java/com/zendesk/maxwell/schema/Database.java
```diff
@@ -48,6 +49,24 @@ public Table findTable(String name) {
 		return null;
 	}
 
+	public Table findTableOrThrow(String table) throws SchemaSyncError {
+		Table t = findTable(table);
+		if ( t == null )
+			throw new SchemaSyncError("Couldn't find table '" + table + "'" + " in database '" + this.name);
```
dadah89: This seems out of place... maybe make it part of SchemaChange? (it'd need to accept a database as param)

dadah89: or maybe a generic table missing exception?

src/main/java/com/zendesk/maxwell/schema/Schema.java
```diff
@@ -42,6 +44,17 @@ public Database findDatabase(String string) {
 		return null;
 	}
 
+	public Database findDatabaseOrThrow(String name) throws SchemaSyncError {
+		Database d = findDatabase(name);
+		if ( d == null )
+			throw new SchemaSyncError("Couldn't find database '" + name + "'");
```
dadah89: similar to previous comment, schema sync error seems out of place here

src/main/java/com/zendesk/maxwell/schema/Database.java
osheroff: if you look at the shape of the calling code, it's pretty common that we have a database in hand, and then need to check if the table exists, or what have you.  We could maybe just rename `SchemaSyncError` to `InvalidSchemaError`?

src/main/java/com/zendesk/maxwell/schema/ddl/TableAlter.java
```diff
@@ -34,29 +31,17 @@ public String toString() {
...
-			Database destDB = newSchema.findDatabase(this.newDatabase);
-			if ( destDB == null )
-				throw new SchemaSyncError("Couldn't find database " + this.database);
-
-			table.rename(newTableName);
+			schema.findDatabaseOrThrow(this.newDatabase);
 
-			database.getTableList().remove(table);
-			destDB.addTable(table);
+			table.name = newTableName;
```
dadah89: No need to check `schema.getCaseSensitivity()` here?

osheroff: ah, good call.  I will add some tests around renaming tables on a lower-case server.  I think I can do that...

src/main/java/com/zendesk/maxwell/schema/Database.java
dadah89: `InvalidSchemaError` sounds good to me

dadah89:
We're basically searching for the same databases/tables twice with those changes (e.g. once in `TableCreate` and another time in `ResolvedTableCreate`), but it'll make more sense once it starts serializing them.

Checked the code changes but just quickly checked tests, everything seems fine to me, :+1:

src/main/java/com/zendesk/maxwell/schema/ddl/TableCreate.java
osheroff: mmmaybe.  I dunno.  It depends on what jackson chooses for serialization, the getter or the attribute

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -8,33 +8,56 @@
...
-
-		renumberColumns();
 	}
 
+	@JsonProperty("columns")
 	public List<ColumnDef> getColumnList() {
 		return columnList;
 	}
 
+	@JsonProperty("columns")
```
dadah89: is this correct, JsonProperty on a setter? (and it already defines `getColumnList ` as 'columns' above)

osheroff: yeah this is correct.

src/main/java/com/zendesk/maxwell/schema/ddl/TableCreate.java
osheroff: err, this isn't true.  it's true before, when we called `database.buildTable()`, but not now.  I think there's probably another pass of coding in here to make the resolver-code smart and the data-code dumb -- ie the resolver should be telling the `Table` "you have this character set, based on the nature of the CREATE TABLE and what I know about the database", the `Table` itself shouldn't know about implicit charsets and the like

PR #292, opened 2016-03-07
consume OpenReplicator heartbeats

osheroff:
long-lived connections can, at some points, good dead.  It's better to be vigilant than assume the network is reliable.

Note that I've tested this locally with `kill -STOP [mysqld_pid]`, but I don't know of a reliable way to get the connection to die underneath me in a test suite. 

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
@@ -58,6 +58,8 @@ public class MaxwellReplicator extends RunLoopProcess {
 
 		this.replicator.setLevel2BufferSize(50 * 1024 * 1024);
 
+		this.replicator.setHeartbeatPeriod(0.5f);
+
 		this.producer = producer;
 		this.bootstrapper = bootstrapper;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -79,6 +81,13 @@ private void ensureReplicatorThread() throws Exception {
 			LOGGER.warn("open-replicator stopped at position " + replicator.getBinlogFileName() + ":" + replicator.getBinlogPosition() + " -- restarting");
 			replicator.start();
 		}
+
+		Long ms = replicator.millisSinceLastEvent();
+		if ( ms != null && ms > 2000 ) {
+			LOGGER.warn("no heartbeat heard from server in " + ms + "ms.  restarting replication.");
+			replicator.stop(5, TimeUnit.SECONDS);
```
dadah89: doesn't it need to stop the `binlogEventListener`? (similarly to `beforeStop`)

dadah89:
lgtm :+1:

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
osheroff: no, that only applies if we need to shut down the whole system.  bouncing the replicator underneath the event listener is fine.

PR #293, opened 2016-03-08
Column filter

visgarw:
Hi ,

I added column filtering feature. In Db table few columns has sensitive data that not needs to be passed to other stream. I have created a Filter Producer which helps other us to filter the column before passing the information to other producer.


osheroff:
Hi @visgarw,
Thanks for the PR.  It's a start, but needs quite a bit of work to be mergable:

- you've deleted all sorts of files you shouldn't have (especially .travis.yml)
- you've changed whitespace in a bunch of places that didn't need to be changed
- I think a better approach than creating a whole new producer would be doing something like `RowMap#toJSON(columnFilter)`, or even just passing the `ColumnFilter` into the `RowMap`'s initializer. 
- Generally configuration is all done with raw command line flags.  So you'd have something like: `--exclude-column=col_name`, `--exclude-column=table.col_name`, or even `--exclude-column=database.table.col_name`.

visgarw:
Hi ,
Thanks for having a look. Will add ColumnFilter to RowMap's initializer.  if i use `--exclude-column=table.col_name` then it will be big list in case of more column and too difficult to manage. I'm passing the configLocation through command prompt.

osheroff:
(can't find this comment)

> if i use `--exclude-column=table.col_name` then it will be big list in case of more column and too difficult to manage. I'm passing the configLocation through command prompt.

If you have lots of columns you can write them out to a `config.properties` file -- all command line options are also supported as properties in the config file.

lileeyao:
hey @visgarw , I'm planning to carry on this feature. If you think otherwise, please let me know. THX :)

PR #295, opened 2016-03-12
tweaks to DDL serialization / deserialization

osheroff:
- do custom serialize / deserialize in `ColumnDef`.  I couldn't get the builtin jackson stuff to work right (it kept duplicating the `type` column). 
- modify table creation to be the same as table alter (a `def` column). 

@zendesk/rules


src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefDeserializer.java
```diff
@@ -0,0 +1,36 @@
+package com.zendesk.maxwell.schema.columndef;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.IntNode;
```
dadah89: not used?

dadah89:
:+1:

PR #296, opened 2016-03-15
add google pubsub producer

mravey:
this is a first draft to get feedbacks on a pubsub producer

### DIFF
--- a/build/maven_fetcher
+++ b/build/maven_fetcher
@@ -88,6 +88,8 @@ module MavenFetcher
       @vars[k] ||= case k
         when "project.version"
           first_text(@doc.root, "version") || first_text(@doc.root, "parent/version")
+        when "project.groupId"
+          first_text(@doc.root, "groupId") || first_text(@doc.root, "parent/groupId")
         else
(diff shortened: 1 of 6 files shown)


osheroff:
cool.  Always happy to see more producers.

There's ordering issues with your PR: since you're using an async http client, there's no guarantee that your updates will be sent to google in order.  Now I'm not sure what kind of guarantees the PubSub service provides, but at the least you have to make sure that the binlog position only moves in one direction:

```
	@Override
 +	public void completed(HttpResponse httpResponse) {
 +		try {
 +			if (this.rowMap.isTXCommit()) {
 +				this.context.setPosition(this.rowMap.getPosition());
 +			}
 +		} catch (SQLException e) {
```

Here one needs to ensure that we're not "rewinding time".  
More sophisticated (guaranteeing at-least-once delivery) would be something like:

Have the producer drop rows into a bounded queue.  Buffer 1 seconds' worth of rows at a time and send them to google (on a separate thread) in a batch, using a synchronous http call.  Commit the offset of the last row that you sent.  (this is roughly what the kafka producer does).

liulikun:
Closing this as it's done in #754

PR #299, opened 2016-03-18
Column Filter

lileeyao:
This is took-over work from https://github.com/zendesk/maxwell/pull/293.

### Status
:construction: 

### Reference
https://github.com/zendesk/maxwell/pull/293


lileeyao:
the code is extremely scratch, not runnable.

lileeyao:
what is the whitespace convention here? my vim is not getting along.

osheroff:
Tabs.

osheroff:
Yeah I'm expecting you'll gut most of it.  But the feature idea is good.

lileeyao:
@osheroff I'm not familiar with the tests here, what do you need?

osheroff:
the integration test framework here:

https://github.com/zendesk/maxwell/blob/a4638160bcefa734504e881f46b3e81df48ff44f/src/test/java/com/zendesk/maxwell/MaxwellIntegrationTest.java#L56

Is probably what you want.  you put in sql (and different settings), you get back RowMap stuff to assert on.

lileeyao:
@osheroff I need your help here, the test is driving me nuts. Apparently `getRowsForSQL` is mocking the producer by 
```
		AbstractProducer producer = new AbstractProducer(context) {
			@Override
			public void push(RowMap r) {
				list.add(r);
			}
		};
```

which does not touch `toJSON`, I have no idea how to refactor this (just seems a really huge refactor to me).

osheroff:
1.  try instantiating the `RowMap` with the list of excluded columns rather than passing it into `toJSON`.  much much cleaner.

osheroff:
2.  There is no 2.  Do that instead and get back to me.

lileeyao:
ping

lileeyao:
@osheroff this code has a minor situation: if all `data` / `old_data` has been filtered, should we keep the message or just remove it entirely?

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -81,6 +88,10 @@ public void blacklistDatabases(String name) throws MaxwellInvalidFilterException
 		blacklistDatabases.add(compile(name));
 	}
 
+	public void excludeColumns(String name) throws MaxwellInvalidFilterException {
+		excludeColumns.add(name);
```
osheroff: in all the other patterns we do `compile(name)` which lets us do regexp-based exclusion.  should support the same pattern here.

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -153,10 +161,22 @@ public String toJSON() throws IOException {
...
+		else {
+			for (String column : this.excludeColumns) {
+				this.data.remove(column);
+				this.oldData.remove(column);
+			}
+			
+			if ( !this.data.isEmpty() )
+				writeMapToJSON("data", this.data, true);
+			if ( !this.oldData.isEmpty() )
+				writeMapToJSON("old", this.oldData, true);
```
osheroff: this code can be simplified; first iterate and remove the excluded columns, then execute the same code path -- there's no need to do different things if you have excluded columns or not.

osheroff:
if all of `old_data` has been filtered I'd say we leave it out.  If all of `data` has been filtered, I suppose we should leave it in, just so there's a record that the transaction ever happened.

lileeyao:
ping

lileeyao:
so now, if the data has been removed, I will do `this.data.put(key, "")`, leaving an empty-value record.

lileeyao:
:question:

lileeyao:
ping

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -105,6 +107,10 @@ protected OptionParser buildOptionParser() {
 
 		parser.accepts( "help", "display help").forHelp();
 
+		parser.accepts( "__separator_8" );
+
+		parser.accepts( "exclude_columns", "columns which will be excluded, use format as --exclude_columns=id,name" ).withOptionalArg();
+
```
osheroff: put this in the same section as `exclude_tables`, `include_tables`, etc etc

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -153,9 +163,26 @@ public String toJSON() throws IOException {
...
+			// NOTE: to avoid concurrent modification.
+			Set<String> keys = new HashSet<String>();
+			keys.addAll(this.data.keySet());
+			keys.addAll(this.oldData.keySet());
+
+			for ( Pattern p : this.excludeColumns ) {
+				for ( String key : keys ) {
+					if ( p.matcher(key).matches() ) {
+						// NOTE: let the record still be in the data.
+						this.data.put(key, "");
```
osheroff: I think just remove the key from `data` too.

lileeyao:
ping

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -98,6 +98,7 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "exclude_dbs", "exclude these databases, formatted as exclude_dbs=db1,db2").withOptionalArg();
 		parser.accepts( "include_tables", "include these tables, formatted as include_tables=db1,db2").withOptionalArg();
 		parser.accepts( "exclude_tables", "exclude these tables, formatted as exclude_tables=tb1,tb2").withOptionalArg();
+		parser.accepts( "exclude_columns", "exclude these columns, , formatted as exclude_columns=col1,col2" ).withOptionalArg();
```
osheroff: double comma

osheroff:
:+1:

Issue #224, opened 2015-12-23
Schema changes slow on large databases


jpeyper:
Continuing from the discussion here: https://github.com/zendesk/maxwell/issues/223

The performance issue we have is that we have a database with a lot of databases, tables and columns.

Once Maxwell has done its initial load (which takes about 30-45 minutes) it looks like this.

```
SELECT COUNT(*) FROM maxwell.databases;
count(*)  
----------
       282
```
```
SELECT COUNT(*) FROM maxwell.tables;
COUNT(*)  
----------
     22516
```
```
SELECT COUNT(*) FROM maxwell.columns;
COUNT(*)  
----------
    947392
```

in my config i have this
```
...
include_dbs=asp,bbs,bip,bss,global,globip,jrb,map,tspa
include_tables=contact,investor
...
```

The tables I'm wanting to get events for are actually very static, almost never changing in structure (maybe once a year).

There are lots of tables and databases and tables I don't care about, and likely never will.

But every time a table gets created or altered in another schema Maxwell stops for many minutes while it deals with the structure change, before continuing through the log.

As @osheroff rightfully pointed out, the issue isn't that Maxwell tracks the entire server's structure changes, its that its slow at this scale.

As I understand, the reason it tracks all structure changes is so if you change your config to monitor different tables it is ready to go.

I guess I'd like to discuss what options there are to solve this.

A few ideas I've had are:

- Improve the performance to a point where I don't care that its doing it (not sure if that's even possible).
- Maxwell only tracks structure changes on included schemas / tables, and on startup checks whether it needs to load any new information into its tables.
- Maxwell has a list of schemas / tables that it ignores for schema changes, and I accept the risk that if I ever do want to track them then I'll have to deal with resetting Maxwell.
- Deal with it
- something amazing that I'm not nearly clever enough to think of...

osheroff:
> Improve the performance to a point where I don't care that its doing it (not sure if that's even possible).

I've begun some exploratory work on this, in which I keep only one full snapshot, and then maintain deltas to the snapshot.  It should make the cost of a schema update a single row insert, which is probably good enough.  

> Maxwell has a list of schemas / tables that it ignores for schema changes, and I accept the risk that if I ever do want to track them then I'll have to deal with resetting Maxwell.

sounds like @nmaquet will take a whack at this, see https://github.com/zendesk/maxwell/issues/252

osheroff:
[...]

PR #300, opened 2016-03-19
[WIP] Compact schema store

osheroff:
this is a big WIP that I'll probably try to split up if possible, but it's close enough to open up a WIP.

The premise is that instead of dumping huge gobs of rows into the database each time the schema changes, instead we will store a base schema along with serialized delta-changes to it.  This *dramatically* speeds up maxwell on a database that undergoes very frequent schema changes.  

what's maybe missing?

- In *theory* a lot of the existing test suite should cover this behavior, but I've got nothing in the way of unit tests.
- at some point letting the schema-deltas grow unbounded may be bad, ie we might want to re-checkpoint the full copy of the schema back to the database after N alters are processed (maybe 10k? something?)

cc especially @dadah89 who's been tracking this issue, but of course also @zendesk/rules.

Addresses https://github.com/zendesk/maxwell/issues/224


jpeyper:
How's this work going?

osheroff:
@jpeyper I've been doing a work/travel thing for the last few weeks and so haven't had time for this guy.  I'll be back in the bay next week and will pick up what's needed to get this tested and merged.  This, fyi is what I'll mark as a 1.1 release.

osheroff:
merged elsewhere.

PR #301, opened 2016-03-19
shutffle around defaults logic

osheroff:
rely on the default value of `.getProperty` to supply defaults.

@zendesk/rules


dadah89:
:+1:

PR #302, opened 2016-03-19
move towards changing the kafka key format

osheroff:
The old keyformat was nicely descriptive, but unfortunately it was quite hard to reproduce in any library that didn't do direct JSON serialization like maxwell does; so re-processing the stream in KStreams sucked, so I thought I'd convert it to something easier to make consistent. 

@zendesk/rules 

NOTE: merge into master by hand.  this is a chained pull request.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -26,6 +26,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 	public final Properties kafkaProperties;
 	public String kafkaTopic;
+	public String kafkaKeyFormat;
 	public String producerType;
 	public String kafkaPartitionHash;
 	public String kafkaPartitionKey;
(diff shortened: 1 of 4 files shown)


wushujames:
What about this?

```
{ "database" : "dbname",
  "table" : "tblname",
  "key" : {
            "col1" : "val1",
            "col2" : "val2" 
  }
}
```

Too verbose? You would have to ensure that the keys are emitted in sorted order, to ensure that compaction works.

osheroff:
> You would have to ensure that the keys are emitted in sorted order, to ensure that compaction works.

Yeah @wushujames that's actually the thing I'm avoiding.  It easy enough from Maxwell's perspective to ensure that the key gets output in that order, but from anywhere outside (think KStreams, or another application that might want to reproduce the key) it's actually a bit of a pain to ensure that the hash is ordered.

wushujames:
Ah, I see.

Did you know that Jackson has a way to automatically sort the keys when serializing to a string? But that's a Jackson-specific thing, so again, it imposes requirements that other people use a specific library.

Why do you need to ensure that other people produce the same key? Are they publishing to a Maxwell stream? Or are they doing joins against the stream that require exact matches against the key (as opposed to parsed versions of the key)?

osheroff:
It's me, I'm the one :)

https://github.com/zendesk/maxwell-join is me screwing around with KStreams to see if I can't produce a maxwell denormalization/join engine, and using KStreams involves a lot of serializing/deserialization of keys

dadah89:
so the key is only used by the kafka producer if no partition is provided (and for logging)?
if wanted to keep more descriptive, could extend @wushujames's suggestion:
`{ "database" : "dbname", "table" : "tblname", "keys": ["col1" : "val1",...]}`

anyway the change seems fine to me, :+1:

osheroff:
@dadah89 no, the key's primary function is to do log compaction.

dadah89:
got it. (the [ProducerRecord docs](https://github.com/apache/kafka/blob/45c8195fa14c766b200c720f316836dbb84e9d8b/clients/src/main/java/org/apache/kafka/clients/producer/ProducerRecord.java#L15-L38) don't mention it at all - -)

osheroff:
merged by hand

Issue #286, opened 2016-02-25
Support multiple topics


criccomini:
It would be nice to have the ability to configure one topic per table, rather than a single topic. An optional prefix config would also be useful.

nmaquet:
:+1: This would be useful for us as well. Right now we're running a Samza job to demux the Maxwell topic into one topic per table; would be even better to have Maxwell do that directly.

rtorrero:
@nmaquet could you by any chance provide more insight on how you do that with Samza? We are looking for a similar solution.

nmaquet:
@rtorrero Sure. Basically, we're running a simple stateless Samza job that reads each Maxwell message on the Maxwell topic and parses the JSON. Then, depending on the values of "database" and "table" in the message, the job outputs the message to the corresponding topic.

nmaquet:
@rtorrero Here's a simple Scala example of what I mean:

```
import java.util.{Map => JavaMap}
import scala.collection.mutable

import org.apache.samza.system.{SystemStream, IncomingMessageEnvelope, OutgoingMessageEnvelope}
import org.apache.samza.task.MessageCollector
import org.apache.samza.task.StreamTask
import org.apache.samza.task.TaskCoordinator
import org.apache.samza.util.Util


object MaxwellDemuxerTask {
  val OutputStreamPrefix = "maxwell"

}

class MaxwellDemuxerTask extends StreamTask {
  private val systemStreamCache = mutable.Map.empty[String, SystemStream]

  override def process(envelope: IncomingMessageEnvelope, collector: MessageCollector, coordinator: TaskCoordinator): Unit = {
    val map = envelope.getMessage.asInstanceOf[JavaMap[String, Object]]
    for {
      database <- Option(map.get("database").asInstanceOf[String])
      table <- Option(map.get("table").asInstanceOf[String])
    } {
      val outputSystemStream = getSystemStreamFromNames(s"kafka.${MaxwellDemuxerTask.OutputStreamPrefix}-$table-$database")
      val out = new OutgoingMessageEnvelope(outputSystemStream, map)
      collector.send(out)
    }
  }

  def getSystemStreamFromNames(name: String): SystemStream = {
    systemStreamCache.getOrElseUpdate(name, Util.getSystemStreamFromNames(name))
  }
}
```

MaheshMadushnaka:
Added changes related to this and created a full request.  
 #372

When we start maxwell if we pass

- --kafka_topic='database.table' create separate topic for each database table
[...]

PR #305, opened 2016-04-01
Allow topics by table name

lileeyao:
@osheroff 

Adding tests now. Just to give a headstart so you know what I'm doing.
### Reference

https://github.com/zendesk/maxwell/issues/286

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -38,12 +38,14 @@ public class MaxwellConfig extends AbstractConfig {
 	public Integer maxSchemas;
 	public BinlogPosition initPosition;
 	public boolean replayMode;
+	public boolean useTableTopic;
 
 	public MaxwellConfig() { // argv is only null in tests
 		this.kafkaProperties = new Properties();
 		this.replayMode = false;
 		this.replicationMysql = new MaxwellMysqlConfig();
 		this.maxwellMysql = new MaxwellMysqlConfig();
+		this.useTableTopic = false;
 	}
 
 	public MaxwellConfig(String argv[]) {
@@ -105,6 +107,10 @@ public class MaxwellConfig extends AbstractConfig {
 		parser.accepts( "__separator_7" );
 
 		parser.accepts( "help", "display help").forHelp();
+
+		parser.accepts( "__separator_8" );
+
+		parser.accepts( "use_table_topic", "configure one topic per table instead of single maxwell topic" );
 
 		BuiltinHelpFormatter helpFormatter = new BuiltinHelpFormatter(200, 4) {
 			@Override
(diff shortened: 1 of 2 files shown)


benmanns:
You may want to include the database name in addition to the table name for the new topic.

lileeyao:
@osheroff I don't see Kafka related tests, but in local test, we now have (seems fine):
```
[2016-04-12 14:14:21,540] INFO [KafkaApi-0] Auto creation of topic maxwell-maxwell_test_database.test_table with 1 partitions and replication factor 1 is successful! (kafka.server.KafkaApis)

```
@benmanns Sorry for late reply. I've added table name too, but for `pk` thing I'll leave it to @osheroff to decide.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -108,6 +110,10 @@ protected OptionParser buildOptionParser() {
 
 		parser.accepts( "help", "display help").forHelp();
 
+		parser.accepts( "__separator_8" );
+
+		parser.accepts( "use_table_topic", "configure one topic per table instead of single maxwell topic" );
```
osheroff: put this option in with the kafka options, and name it `kafka_topic_per_table`

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -93,13 +97,20 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
...
 
 		ProducerRecord<String, String> record =
 				new ProducerRecord<>(topic, this.partitioner.kafkaPartition(r, this.numPartitions), key, value);
 
 		kafka.send(record, new KafkaCallback(r, this.context, key, value));
 	}
 
+	private void addTableTopic(boolean useTableTopic, RowMap r) {
+		if (useTableTopic) {
+			this.topic = this.originalTopic + "-" + r.getDatabase() + "." + r.getTable();
```
osheroff: you don't need to set the instance variable here; it's messy.  better to do something like:

```
   String value = r.toJSON();
   String messageTopic; 
   if ( useTableTopic ) { 
      messageTopic = getTableTopic(...)
  } else {
      messageTopic = this.topic
  }
```

osheroff:
this PR needs to take into consideration:

- the partitioning code needs to be aware of each different topic.
- the default partitioner (by database) probably makes less sense in this use case -- we might want to partition by primary key by default instead. 
- we *may* want to allow both forms of topic names, either with database or without.  

And like we discussed yesterday, it might be that we want to somehow rework all this stuff into a more generic transformation layer rather than doing this featuritis stuff.  But I dunno.

benmanns:
A couple other things I thought of related to this - [Kafka topic name validations](https://github.com/apache/kafka/blob/5236bf60debbb0c08010315a92dd3fbfa482e871/core/src/main/scala/kafka/common/Topic.scala#L24-L25) are different from [MySQL identifier validations](http://dev.mysql.com/doc/refman/5.7/en/identifiers.html) so we may want to strip or sanitize the potential values. E.g.
```sql
CREATE TABLE `foo bar @(*$&(*$ ⛄️` (id INT);
```
actually worked on my local install.

osheroff:
finished elsewhere

Issue #312, opened 2016-04-12
add --jdbc_options command line flag


mcleae:
I'm trying to bootstrap a large table and I keep exceeding my db's wait_timeout setting.  Instead of changing the setting on the server I'd like to tweak it for just the maxwell db connection by including the netTimeoutForStreamingResults property in the jdbc connection string.  Since maxwell constructs the jdbc connection string from the --host and --port properties I was wondering if there was a way to give it additional properties for the jdbc connection string so i end up with something like 
jdbc:mysql://localhost:3306?useCursorFetch=true&zeroDateTimeBehavior=convertToNull&netTimeoutForStreamingResults=XXX

thanks

osheroff:
@lileeyao can you take a look at adding an option that lets you pass through jdbc options?

lileeyao:
k, will do it soon.

PR #314, opened 2016-04-14
jdbc options

lileeyao:
### Reference
https://github.com/zendesk/maxwell/issues/312

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
@@ -1,6 +1,7 @@
 package com.zendesk.maxwell;
 
 import java.util.ArrayList;
+import org.apache.commons.lang.StringUtils;
 
 import joptsimple.OptionSet;
 
@@ -42,15 +43,16 @@ public class MaxwellMysqlConfig {
 		if ( options.has(prefix + "port"))
 			this.port = Integer.valueOf((String) options.valueOf(prefix + "port"));
 		if ( options.has(prefix + "jdbc_options") ) {
-			String[] opts = ((String) options.valueOf(prefix + "jdbc_options")).replaceAll("\\s+", "").split(",");
+			String[] opts = ((String) options.valueOf(prefix + "jdbc_options")).split("&");
 			for ( String opt : opts ) {
-				this.jdbc_options.add(opt);
+				this.jdbc_options.add(opt.trim());
 			}
 		}
 	}
 
 	public String getConnectionURI() {
-		return "jdbc:mysql://" + host + ":" + port + "?" + String.join("&", this.jdbc_options);
+		return "jdbc:mysql://" + host + ":" + port + "?" +				
+				StringUtils.join(this.jdbc_options.toArray(), "&");
 	}
 
 	@Override
--- a/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
@@ -1,5 +1,7 @@
 package com.zendesk.maxwell;
 
+import java.util.ArrayList;
+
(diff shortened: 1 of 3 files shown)


lileeyao:
ping @osheroff

src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
```diff
@@ -35,9 +42,18 @@ public void parseOptions( String prefix, OptionSet options) {
 			this.user = (String) options.valueOf(prefix + "user");
 		if ( options.has(prefix + "port"))
 			this.port = Integer.valueOf((String) options.valueOf(prefix + "port"));
+		if ( options.has(prefix + "jdbc_options") ) {
+			String[] opts = ((String) options.valueOf(prefix + "jdbc_options")).replaceAll("\\s+", "").split(",");
```
osheroff: we should only replace leading and trailing whitespace.   Also: probably we should leave the options ampersand separated

lileeyao: what do you mean by `leave the options ampersand separated`?

osheroff: meaning you're doing `split(",")`, I'm saying you should be doing `split("&")`

lileeyao:
fixed, in this case I prefer to split into an array of options, which may be better organized than just concating strings. But it's your call.

osheroff:
👍

mcleae:
so this seems to work fine when i pass it as a command line option but doesn't get recognized when i include it in a configuration file.

As a quick check i put in a bad hostname so i could see the jdbc string, here's what i get

so when just in mine config file
```
Could not connect to jdbc:mysql://XXXXXX:3306?useCursorFetch=true&zeroDateTimeBehavior=convertToNull
```

but when passed as a command line option
```
Could not connect to jdbc:mysql://localhost:3306?useCursorFetch=true&zeroDateTimeBehavior=convertToNull&netTimeoutForStreamingResults=3600
```

I was under the impression all command line options could be passed via a config file as well.  Please let me know if that's not the case.  Thanks

mcleae:
In case its of assistance below are the contents of my config file and how i'm invoking
```
cat conf/config.properties.bad
user=maxwell
password=XXXXXXX
host=XXXXXX
producer=stdout
kafka_partition_by=primary_key
bootstrapper=sync
jdbc_options=netTimeoutForStreamingResults=3600&sessionVariables=wait_timeout=3600
log_level=DEBUG
```

Invocation 1
bin/maxwell --config /opt/maxwell/conf/config.properties.bad
```
14:38:45,545 ERROR MaxwellContext - Could not connect to jdbc:mysql://XXXXXX:3306?useCursorFetch=true&zeroDateTimeBehavior=convertToNull: Communications link failure
```

Invocation 2
bin/maxwell --jdbc_options=netTimeoutForStreamingResults=3600\&sessionVariables=wait_timeout=3600 --config /opt/maxwell/conf/config.properties.bad
```
14:40:13,640 ERROR MaxwellContext - Could not connect to jdbc:mysql://XXXXXX:3306?useCursorFetch=true&zeroDateTimeBehavior=convertToNull&netTimeoutForStreamingResults=3600&sessionVariables=wait_timeout=3600: Communications link failure
```

As discussed only the second invocation with the jdbc_option as a command line arg results in a modified jdbc string.

osheroff:
yeah this is just an oversight.  @lileeyao can you fix?

mcleae:
one more thing, while debugging this today i noticed that the replication connection string is the one used for the bootstrap and as presently written in MaxwellMysqlConfig.parseOptions jdbc_options (no prefix) will only be passed to the maxwellConnectionPool not the replicationConnectionPool.  i added a dupe param as replication_jdbc_options as a workaround but not sure if that was the intent.

PR #316, opened 2016-04-21
update document

lileeyao:
the quickstart instruction of kafka has been updated, the old document has some misleading info.


lileeyao:
merging, since its doc update, nothing harmful.

PR #323, opened 2016-04-25
Stop keeping a Connection object with the schema

osheroff:
it was dangerous and silly to do that -- the connection is pulled from the connection pool, and so shouldn't be kept for longer than it takes to do work.

@zendesk/rules


dadah89:
lgtm 👍

PR #325, opened 2016-04-27
Delta schemas

osheroff:
This PR addresses https://github.com/zendesk/maxwell/issues/224

After a ton of refactoring and work, this PR finally lands!  Instead of doing schema "snapshots" with each table alter, we now store a snapshot and deltas.  What this means is that each time we parse some DDL, we: 

1. apply the DDL to the in memory schema
2. perform an insert into `maxwell.schemas`, with a JSON-serialized version of the DDL and a backpointer to the previous schema.

When we want to restore the schema, first we restore the snapshot schema (referred to in code as "base schema") as normal, and then apply the list of schema deltas.

The net result is that we can now process thousands of DDL statements per second on large schemas, instead of taking > 3 seconds to process a single DDL statement.  We make some minor tradeoffs in boot-time speed.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -36,7 +36,6 @@ public class MaxwellConfig extends AbstractConfig {
 	public String outputFile;
 	public String log_level;
 
-	public Integer maxSchemas;
 	public BinlogPosition initPosition;
 	public boolean replayMode;
 
@@ -91,7 +90,7 @@ public class MaxwellConfig extends AbstractConfig {
 		parser.accepts( "__separator_5" );
 
 		parser.accepts( "schema_database", "database name for maxwell state (schema and binlog position)").withRequiredArg();
-		parser.accepts( "max_schemas", "how many old schema definitions maxwell should keep around.  default: 5").withOptionalArg();
+		parser.accepts( "max_schemas", "deprecated.").withOptionalArg();
 		parser.accepts( "init_position", "initial binlog position, given as BINLOG_FILE:POSITION").withRequiredArg();
 		parser.accepts( "replay", "replay mode, don't store any information to the server");
 
@@ -178,9 +177,6 @@ public class MaxwellConfig extends AbstractConfig {
 		if ( options.has("output_file"))
 			this.outputFile = (String) options.valueOf("output_file");
 
-		if ( options.has("max_schemas"))
-			this.maxSchemas = Integer.valueOf((String)options.valueOf("max_schemas"));
-
 		if ( options.has("init_position")) {
 			String initPosition = (String) options.valueOf("init_position");
 			String[] initPositionSplit = initPosition.split(":");
@@ -259,10 +255,6 @@ public class MaxwellConfig extends AbstractConfig {
 		this.blacklistDatabases = p.getProperty("blacklist_dbs");
 		this.blacklistTables = p.getProperty("blacklist_tables");
 
-		String maxSchemaString = p.getProperty("max_schemas");
-		if (maxSchemaString != null)
-			this.maxSchemas      = Integer.valueOf(maxSchemaString);
-
 		if ( p.containsKey("log_level") )
 			this.log_level = parseLogLevel(p.getProperty("log_level"));
 
@@ -335,8 +327,9 @@ public class MaxwellConfig extends AbstractConfig {
 									this.maxwellMysql.password);
 		}
 
-		if ( this.maxSchemas != null )
-			SchemaStore.setMaxSchemas(this.maxSchemas);
+		if ( this.databaseName == null) {
+			this.databaseName = "maxwell";
+		}
 	}
 
 	public Properties getKafkaProperties() {
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
@@ -273,9 +273,7 @@ public class SchemaStore {
 		return mapper.readerFor(listOfResolvedSchemaChangeType).readValue(json.getBytes());
 	}
 
-	/* restore a chain of derived schemas by following the chain backwards
-	 * until we find the "base" schema. */
-	private void restoreDerivedSchema(Connection conn, Long base_schema_id, Long schema_id) throws SQLException, IOException, InvalidSchemaError {
+	private HashMap<Long, HashMap<String, Object>> buildSchemaMap(Connection conn) throws SQLException {
 		HashMap<Long, HashMap<String, Object>> schemas = new HashMap<>();
 
 		PreparedStatement p = conn.prepareStatement("SELECT * from `schemas` where server_id = ?");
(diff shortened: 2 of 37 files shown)


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -91,7 +90,7 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "__separator_5" );
 
 		parser.accepts( "schema_database", "database name for maxwell state (schema and binlog position)").withRequiredArg();
-		parser.accepts( "max_schemas", "how many old schema definitions maxwell should keep around.  default: 5").withOptionalArg();
+		parser.accepts( "max_schemas", "deprecated.").withOptionalArg();
```
lileeyao: maybe just remove it rather than mark it?

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -229,8 +266,79 @@ public static SchemaStore restore(Connection connection, MaxwellContext context)
...
+		return schemaChain;
+	}
+
+	private void restoreDerivedSchema(Connection conn, Long schema_id) throws SQLException, IOException, InvalidSchemaError {
+		/* build hashmap of schema_id -> schema properties (as hash) */
+		HashMap<Long, HashMap<String, Object>> schemas = buildSchemaMap(conn);
+
+		/* walk backwards to build linked list with base schema at the
+		 * head, and the rest of the delta schemas following */
+		LinkedList<Long> schemaChain = buildSchemaChain(schemas, schema_id);
```
lileeyao: if you prefer to traverse linkedlist, you can just `LinkedHashMap` in the first place, thus you won't bother building chains.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: no, I want to have a release or two with it deprecated so I don't just break everyone's installation.

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -276,11 +392,16 @@ private void restoreSchemaMetadata(Connection conn, Long schemaID) throws SQLExc
 		LOGGER.info("Restoring schema id " + schemaRS.getInt("id") + " (last modified at " + this.position + ")");
 
 		this.schema_id = schemaRS.getLong("id");
+		this.base_schema_id = schemaRS.getLong("base_schema_id");
+
+		if ( schemaRS.wasNull() )
```
dadah89: Without this check `this.base_schema_id` would be set to the string 'NULL'?

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -401,17 +522,6 @@ public BinlogPosition getBinlogPosition() {
 		return this.position;
 	}
 
-	private void deleteOldSchemas(Connection connection, Long currentSchemaId) throws SQLException {
-		if ( maxSchemas <= 0  )
-			return;
-
-		Long toDelete = currentSchemaId - maxSchemas; // start with the highest numbered ID to delete, work downwards until we run out
-		while ( toDelete > 0 && schemaExists(connection, toDelete) ) {
```
dadah89: is `schemaExists` still being used? safe to remove?

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -110,7 +110,7 @@ public Table copy() {
 		ArrayList<ColumnDef> list = new ArrayList<>();
 
 		for ( ColumnDef c : columnList ) {
-			list.add(c.copy());
```
dadah89: isn't this going to cause issue with alters? (that's my understanding of where those `copy` methods are used)

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
osheroff: It'd be `0`.  For some unbelievable reason.  http://stackoverflow.com/questions/2920364/checking-for-a-null-int-value-from-a-java-resultset

src/main/java/com/zendesk/maxwell/schema/Table.java
osheroff: no, because we never mutate `ColumnDef` objects.

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
```diff
@@ -229,8 +266,79 @@ public static SchemaStore restore(Connection connection, MaxwellContext context)
...
+		if ( json == null )
+			return null;
+
+		return mapper.readerFor(listOfResolvedSchemaChangeType).readValue(json.getBytes());
+	}
+
+	private HashMap<Long, HashMap<String, Object>> buildSchemaMap(Connection conn) throws SQLException {
+		HashMap<Long, HashMap<String, Object>> schemas = new HashMap<>();
+
+		PreparedStatement p = conn.prepareStatement("SELECT * from `schemas` where server_id = ?");
```
dadah89: isn't this going to get really expensive as more schemas get added with time?

osheroff: eventually.  I've tested up to like 30k deltas, and the whole thing stays pretty fast, I will see what happens around like 1mm deltas.

dadah89: what I'm thinking: store the first_schema_id (I'd call them first_schema_id and parent_schema_id, base_schema_id for me might be understood as first or parent... not an english speaker though =)) in `saveDerivedSchema` (get it from parent.first_schema_id), and in buildSchemaMap the query changes to `"SELECT * from `schemas` where server_id = ? and first_schema_id = ?"`.

That way it is easy to only select the schemas that are part of this schema's chain.

If it won't be expensive, ignore the comment =)

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
dadah89: oh because of getLong, weird

src/main/java/com/zendesk/maxwell/schema/SchemaStore.java
osheroff: oh.  I see.  Right now we only have one chain at once, so we won't end up getting data we don't need (generally).  if and when I write something that occasionally snapshots chains, I can do this there.

dadah89:
so the gist is: for a base schema (the first row in the schema db), store extra rows, each one containing a string/json with deltas/changes. To restore, get the most recent, walk the chain to the first, restore that schema, apply deltas, last delta row is the new schema. When a new delta is added, the previous delta row will be the parent. Like this approach.

maybe at some point might be worth to add a thread to truncate schema deltas? guess I'm a bit paranoid with performance hah.

lgtm, 👍

osheroff:
@dadah89 precisely.  I will measure how many deltas it's feasible to keep; my suspicion is that it's in the hundreds of thousands, and after that it will be worthwhile to checkpoint a new schema, but of course it really all depends on the size of the schema.

PR #327, opened 2016-04-28
lighten kafka memory buffer footprint a bit

osheroff:
When kafka backs up, the `KafkaCallback` objects end up holding references to objects.  This allows at least the RowMap objects to get GC'ed in that case.

@zendesk/rules


dadah89:
👍

PR #329, opened 2016-04-29
stop copying schemas

osheroff:
this is worth about 100x speed-up when processing large amounts of deltas, and should be safe; the copying was important in the world before ResovledSchemaChange, but not now.

@zendesk/rules


lileeyao:
what is the original `newSchema` ( and creating copies ) used for?

osheroff:
it used to be used to take diffs of two schema in test mode, and also it used to be that there was a weird api contract -- if a schema change was actually a NO-OP (like `create table if not exists "foo" when "foo" exists), it would leave the old object in place.

lileeyao:
👍

PR #330, opened 2016-05-02
Aurora capture performance fixes

osheroff:
stop asking mysql to order primary keys by ordinal_position.  For some odd reason this is a performance nightmare on AWS Aurora, leading to schema capture times of like 15 minutes (vs 10 seconds after
the fix).

@zendesk/rules


dadah89:
👍

PR #331, opened 2016-05-02
Fix unsigned int schema capture bug

osheroff:
a bug in the regex prevented us from properly capturing whether a column was unsigned or not.  buh.

@zendesk/rules


dadah89:
👍

osheroff:
@dadah89 I'm gonna rebase this on top of master so I can release it as a 1.0.X point

PR #336, opened 2016-05-03
Use a streaming connection/statement for bootstrapping tables

mcleae:
fetch sizes are ignored by the mysql connector, in order to achieve row by row streaming (useful for bootstrapping large tables) you need to configure the statement as follows
```
stmt = conn.createStatement(java.sql.ResultSet.TYPE_FORWARD_ONLY,
              java.sql.ResultSet.CONCUR_READ_ONLY);
stmt.setFetchSize(Integer.MIN_VALUE);
```
additional details in the connector documentation [here](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-implementation-notes.html)

### DIFF
--- a/src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
+++ b/src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
@@ -48,9 +48,10 @@ public class SynchronousBootstrapper extends AbstractBootstrapper {
 		producer.push(startBootstrapRow);
 		producer.push(bootstrapStartRowMap(table, position));
 		LOGGER.info(String.format("bootstrapping started for %s.%s, binlog position is %s", databaseName, tableName, position.toString()));
-		try ( Connection connection = getConnection() ) {
+		try ( Connection connection = getConnection();
+                Connection streamingConnection = getStreamingConnection()) {
 			setBootstrapRowToStarted(startBootstrapRow, connection);
-			ResultSet resultSet = getAllRows(databaseName, tableName, schema, connection);
+			ResultSet resultSet = getAllRows(databaseName, tableName, schema, streamingConnection);
 			int insertedRows = 0;
 			while ( resultSet.next() ) {
 				RowMap row = bootstrapEventRowMap("bootstrap-insert", table, position);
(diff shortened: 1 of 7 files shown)


src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -182,8 +189,8 @@ private ResultSet getAllRows(String databaseName, String tableName, Schema schem
 	}
 
 	private Statement createBatchStatement(Connection connection) throws SQLException, InterruptedException {
-		Statement statement = connection.createStatement();
-		statement.setFetchSize(context.getConfig().bootstrapperBatchFetchSize);
+		Statement statement = connection.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
+		statement.setFetchSize(Integer.MIN_VALUE);
```
osheroff: So I went and looked into this further; according to http://stackoverflow.com/questions/26046234/is-there-a-sql-jdbc-that-will-respect-fetchsize/36986532#36986532, with `useCursorFetch=true` we were supposed to not buffer the whole damn thing into memory.  You're definitely right about using making mistakes writing to the connection while we're still reading results, but lemme poke a bit more at this tomorrow.

But also just turning this into a streaming query is likely totally fine.

mcleae: sounds good, as an FYI before making the streaming change I also tried bumping up to the latest connector version (5.1.38) with no luck. That version itself as a drop in seemed to work fine but didn't fix the cursor fetching.

I also tried the following once the jdbc_options patch was in place with no luck

netTimeoutForStreamingResults=3600&sessionVariables=wait_timeout=3600&useUnbufferedInput=true&useReadAheadInput=false

osheroff: caught up with production bugs.  Got back to this, and yeah, I'm not sure how this code ever worked.  when I point it at my db it... doesn't.  But I know for sure that it used to, but hey.

Anyway, I'll merge this.  opened https://github.com/RentTheRunway/maxwell-1/pull/1 for your review (tears down some of the stuff around the old methods), if you wanna merge that I'll merge this.

mcleae: merged on my fork, thanks

osheroff:
hmmm, something went wrong with the merge.

👍  here, but I'll merge by hand.

osheroff:
I merged this by hand.

PR #337, opened 2016-05-04
Fix encoding stuff

osheroff:
change method for setting up default file encoding.  In rare cases, the locale `en_US.UTF-8` appears to not be available.  This then defaults all output buffers to a standard "C" locale, which then happily corrupts utf-8 data.

At least, I think that's what's happening.  adding `-Dfile.encoding=UTF-8` appears to fix the case that I could reproduce locally. 

@zendesk/rules


osheroff:
rebased off the 1.0 branch: https://github.com/zendesk/maxwell/pull/338

PR #338, opened 2016-05-04
change method for setting up default file encoding

osheroff:
In rare cases, the locale `en_US.UTF-8` appears to not be available.
this then defaults all output buffers to a standard "C" locale, which
then happily corrupts utf-8 data.

At least, I think that's the gist of it.  adding `-Dfile.encoding=UTF-8`
appears to fix the case that I could reproduce.

@zendesk/rules


lileeyao:
👍

PR #340, opened 2016-05-04
BitColumn support for java Boolean

mcleae:


src/main/java/com/zendesk/maxwell/schema/columndef/BitColumnDef.java
```diff
@@ -16,7 +16,12 @@ public boolean matchesMysqlType(int type) {
 
 	@Override
 	public Object asJSON(Object value) {
-		byte[] bytes = (byte[]) value;
+		byte[] bytes;
+		if(Boolean.class.isAssignableFrom(value.getClass())){
```
osheroff: what classes are we account for with `isAssignableFrom`?

osheroff: whitespace should look like `if ( Boolean... ) {`

src/main/java/com/zendesk/maxwell/schema/columndef/BitColumnDef.java
```diff
@@ -16,7 +16,12 @@ public boolean matchesMysqlType(int type) {
 
 	@Override
 	public Object asJSON(Object value) {
-		byte[] bytes = (byte[]) value;
+		byte[] bytes;
+		if( Boolean.class.isAssignableFrom(value.getClass()) ){
```
mcleae: fixed whitespace, isAssignableFrom covers the case where they are of the same type or Boolean is a superclass/superinterface of value.getClass.  its a habit i'm in because its null correct unlike say Boolean.class.isInstance but I think either would do in this case.  if you have a preference just let me know

mcleae: actually i take that back, null should probably not trickle through safely here since it could produce an npe.  i'll tighten it up with the instance check

osheroff: sounds good.

osheroff:
looks good, thanks for the patch.

👍

PR #341, opened 2016-05-06
relax schema diffs on unsigned integer check

osheroff:
- stop throwing if there's some differences in the schema
- if we do find that we can't diff against the current schema, it's no use just crashing -- just warn and muddle on.

@zendesk/rules


dadah89:
👍

Issue #317, opened 2016-04-21
some primary keys are coming as null  due to case sensitive comparison 


xmlking:
while testing with `Percona MySQL` , we noticed second primarykey is coming as `null `.
wonder if anyone else facing this issue? 

```sql
UPDATE test.person SET
  id = 5,
  firstName = 'aaa',
  address_id = 3
WHERE id = 1 AND firstName = 'qqqq';
```

kafka key:
```json
'{"database":"test","table":"person","pk.id":5,"pk.firstName":null}'
```

kafka value: 
```json
{
    "database": "test",
    "table": "person",
    "type": "update",
    "ts": 1458335796,
    "xid": 4168944,
    "commit": true,
    "data": {
        "firstname": "aaa",
        "address_id": 3,
        "id": 5,
        "lastname": "aaa"
    },
    "old": {
        "firstname": "qqqq",
        "address_id": 4,
        "id": 1
    }
}
```

osheroff:
looks like a case-sensitivity issue.  If you write a test, I can probably whack out a fix.  Or you can try :)

xmlking:
tested locally switching  ` LinkedHashMap  with `TreeMap<String, Object>(String.CASE_INSENSITIVE_ORDER);` in RowMap.java. this resolved the problem.

osheroff:
keep `LinkedHashMap`, but downcase the column names on insert and lookup?  `LinkedHashMap` is in there preserving the order that the columns serialize in json, which is a nice readability win

xmlking:
agree. I will test  `data.containsKey(pk.toLowerCase())`

xmlking:
Changing to `data.containsKey(pk.toLowerCase()) ` at three locations works for me.
Will you cosider releasing this fix?

osheroff:
sure thing.  just PR it

PR #342, opened 2016-05-15
Fix: some primary keys are coming as null due to case sensitive comparison #317

xmlking:
Fix: some primary keys are coming as null due to case sensitive comparison #317
 https://github.com/zendesk/maxwell/issues/317


osheroff:
took a slightly different tack (and wrote a test) here: https://github.com/zendesk/maxwell/pull/342

PR #343, opened 2016-05-16
primary keys, case sensitivity

osheroff:
@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/Table.java
+++ b/src/main/java/com/zendesk/maxwell/schema/Table.java
@@ -19,8 +19,7 @@ public class Table {
 	public String name;
 	private List<ColumnDef> columnList;
 	public String charset;
-	@JsonProperty("primary-key")
-	public List<String> pkColumnNames;
+	private List<String> pkColumnNames;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -259,7 +258,10 @@ public String getPKString() {
 			return null;
 	}
 
+	@JsonProperty("primary-key")
 	public void setPKList(List<String> pkColumnNames) {
-		this.pkColumnNames = pkColumnNames;
+		this.pkColumnNames = new ArrayList<>();
```
lileeyao: `this.pkColumnNames = new ArrayList<String>()`?

lileeyao:
👍

src/main/java/com/zendesk/maxwell/schema/Table.java
osheroff: Check out https://docs.oracle.com/javase/tutorial/java/generics/genTypeInference.html, especially "Type Inference and Instantiation of Generic Classes"

PR #344, opened 2016-05-16
[wip] output DDL as json

osheroff:


xmlking:
@osheroff any chance of merging output DDL to master?

xmlking:
Please add `timestamp` , Raw `SQL`  and filed `size` to DDL output. 

Example...

```json
{
    "type": "table-alter",
    "database": "test",
    "table": "book",
    "ts": 1449786341,
    "sql":"alter table book modify column title varchar(100) not null;"
    "def": {
        "database": "test",
        "charset": "utf8",
        "table": "book",
        "columns": [
            {
                "type": "varchar",
                "name": "title",
                "size": 100,
                "charset": "utf8"
            },
```

osheroff:
@xmlking, sure, sql and ts sound like reasonable suggestions.  regarding merging this PR:  I'm a little hesistant to shove this feature through, as I haven't been able to reconcile how to actually output the DDL events into the kafka topic.  The basic question is, how are they partitioned?

For table creation and modification, if you've partitioned by database, it makes perfect sense for the DDL statements to go directly inline.  If you've partitioned your output topic by anything else, it stops making much sense.  Database creation and drop statements really have no place in particular to go. 

I think the only real way that this makes sense is to separate the DDL statements out into a separate producer configuration, and to say that the DDL statements don't keep any strict order with the row events.  that way you'd have a `maxwell` topic for rows, and `maxwell_ddl` topic for ddl, and never the twain meet.  But all of that is gated behind a larger multi-consumer/multi-producer feature that I'm slowly working towards.

Can you describe your use case for the DDL?  Is it largely logging/auditing based or are you trying to do something programmatic with it?

xmlking:
My use case:
I have to aggregate CDC events into `/${database}/${table}/${type}/v1/` file structure for other downstream batch system to consume on regular interval.  I am using `Apache NiFi` to build this streaming dataflow.  Whenever a table's schema changes, I have to create a new schema file into `/${database}/${table}/${type}/v2/` directory and create  respective CDC events into this directory until next schema change occur for that table. 

I am interested only in table DDL events, not interested in database DDL events. I cam currently using `kafka_partition_by = database` setting, thinking of using `primary_key` in the future. In my case it doesn't matter whether  Maxwell produce all DDL events to same `maxwell` topic or  new `maxwell_ddl` topic. 

My plan is, to build custom NiFi flow to keep track of table's current schema versions  in NiFi's build-in store. This flow will update  NiFi's store whenever new table DDL event received. Other NiFi flows can consume table versions' state from this decentralized NiFi's store.  

For now, you can route DDL and DML to `maxwell` if `kafka_partition_by = database`, otherwise to `maxwell` and `maxwell_ddl`. Wish Kafka will support API to produce some events to all partitions :)

xmlking:
I got temporary workaround to output `SQL` and `Timestamp` here:
https://github.com/xmlking/maxwell/commit/a4425ca0c5bdad9a181009e66496b2e4c91d1a3b 
After exploring options to maintain Table Schema versions (schema version  per table) in NiFi, found it is difficult. Would be nice if Maxwell could also output  Schema version # e.g.,
```json
{
    "type": "table-alter",
    "database": "test",
    "table": "book",
    "ts": 1449786341,
    "sql":"alter table book modify column title varchar(100) not null;"
    "def": {
        "database": "test",
        "charset": "utf8",
        "table": "book",
        "version": "2",
        "columns": [
            {
                "type": "varchar",
                "name": "title",
                "size": 100,
                "charset": "utf8"
            },
```

xmlking:
would be nice if this PR also in sync with master. 
@osheroff if I want just table DDLs (table_name , database_name, sql_statement , timestamp) where in the code I can tap for this info?

PR #346, opened 2016-05-18
Temporary table ALTER, CREATE and DROP now blacklisted

GREX-USS:
Was finding that Maxwell would throw in procedures where DROP was used before CREATE, added to the blacklist to ensure basic DDL statements for temporary tables are ignored.


osheroff:
cool, thanks for the patch.  👍

PR #348, opened 2016-05-23
fix a race condition in the binlog position thread

osheroff:
I latched onto a maxwell in production that was totally stuck.  I found that the main thread was waiting in setSync().  The variables were as follows:

```
main[2] dump p
p = {
    offset: 95823288
    file: "db7-bin-log.001331"
}

main[2] dump storedPosition.value
storedPosition.value = {
    offset: 95822855
    file: "db7-bin-log.001331"
}

main[2] dump position.value
position.value = {
    offset: 95822855
    file: "db7-bin-log.001331"
}
```

As you can see, the "requested" position for the thread(`position.value`) is *behind* the position "p" in setSync().  My best guess of what happened is that a kafka row came back after setSync() called position.set(), and clobbered the requested position.

Fix here is to never allow a position to rewind, which also means protecting the call with "synchronized".

@zendesk/rules

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.min.js binary
--- a/src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
+++ b/src/main/java/com/zendesk/maxwell/schema/SchemaPosition.java
@@ -112,13 +112,14 @@ public class SchemaPosition extends RunLoopProcess implements Runnable {
 		}
 	}
 
-	public void set(BinlogPosition p) {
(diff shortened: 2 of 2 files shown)


osheroff:
btw, didn't know you could attach to a running jvm with `jdb`.  Pretty neat.

osheroff:
`sudo jdb -connect sun.jvm.hotspot.jdi.SAPIDAttachingConnector:pid=PID`

lileeyao:
👍

PR #352, opened 2016-05-25
better at-least-once semantics regarding commiting position

osheroff:
previously there was some possibility of losing data if one partition was master-less, while others were fine.  Now we keep a list of outstanding messages, and set our binlog position as the last continuous
position that's completed.  

@vanchi-zendesk @dadah89 , do you remember talking about binlog positions?  The old code always did `setSync` -- ie after processing DDL the replicator would force the binlog position to be written out to the database -- but I'm really unclear why I thought that was necessary.  Maybe it was that previously it was very expensive (or damaging?) to replay a schema update?  

@zendesk/rules /cc @dasch (some of this is prep-work for independent consumer positions)

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
+++ b/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
@@ -1,0 +1,64 @@
+package com.zendesk.maxwell.producer;
+/* respresents a list of inflight messages -- stuff being sent over the
+   network, that may complete in any order.  Allows for only bumping
+   the binlog position upon completion of the oldest outstanding item.
+
+   Assumes .addInflight(position) will be call monotonically.
+   */
+
+import com.zendesk.maxwell.BinlogPosition;
+
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.ListIterator;
+
+public class InflightMessageList {
+	class InflightMessage {
+		public final BinlogPosition position;
+		public boolean isComplete;
+		InflightMessage(BinlogPosition position) {
+			this.position = position;
+			this.isComplete = false;
+		}
+	}
+
+	private LinkedList<InflightMessage> list;
+	private HashMap<String, InflightMessage> hash;
+
+	public InflightMessageList() {
+		this.list = new LinkedList<>();
+		this.hash = new HashMap<>();
+	}
+
+	public void addMessage(BinlogPosition p) {
+		InflightMessage m = new InflightMessage(p);
+		this.list.add(m);
+		this.hash.put(p.toString(), m);
+	}
+
+	/* returns the position that stuff is complete up to, or null if there were no changes */
+	public BinlogPosition completeMessage(BinlogPosition p) {
+		InflightMessage m = hash.get(p.toString());
+		assert(m != null);
+
+		m.isComplete = true;
+
+		BinlogPosition completeUntil = null;
+		ListIterator<InflightMessage> iterator = this.list.listIterator();
+		while ( iterator.hasNext() ) {
+			InflightMessage msg = iterator.next();
+			if ( !msg.isComplete )
+				break;
+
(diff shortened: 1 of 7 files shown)


dasch:
Should the inflight message buffer be bounded?

Otherwise looks good :+1:

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -0,0 +1,64 @@
...
+			this.isComplete = false;
+		}
+	}
+
+	private LinkedList<InflightMessage> list;
+	private HashMap<String, InflightMessage> hash;
+
+	public InflightMessageList() {
+		this.list = new LinkedList<>();
+		this.hash = new HashMap<>();
```
vanchi-zendesk: LinkedHashMap?

osheroff: Yeah, ok.  The whole thing with modifying iterators and such is pretty confusing but it looks like that'd work.

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -0,0 +1,60 @@
...
+		BinlogPosition completeUntil = null;
+		Iterator<InflightMessage> iterator = this.linkedMap.values().iterator();
+
+		while ( iterator.hasNext() ) {
+			InflightMessage msg = iterator.next();
+			if ( !msg.isComplete )
+				break;
+
+			completeUntil = msg.position;
+			iterator.remove();
```
osheroff: this works, although IMO it's kinda oblique from the code; `values().iterator()` actually iterates in the map itself, so the `remove()` call removes the node safely.

> https://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html#values()

osheroff:
@dasch, there's an implicit bound on the message buffer, which is however big the outstanding kafka buffer is.  Probably best to just leave the bounding there rather than trying to impose another limit.

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
vanchi-zendesk: ok, looks like the standard way: http://stackoverflow.com/questions/223918/iterating-through-a-collection-avoiding-concurrentmodificationexception-when-re

But I am sure you are missing some foldLeft flatMap action here? :-D

vanchi-zendesk:
👍

dasch:
> @dasch, there's an implicit bound on the message buffer, which is however big the outstanding kafka buffer is.  Probably best to just leave the bounding there rather than trying to impose another limit.

:+1:

dadah89:
@osheroff  sorry I'm late to comment, but just a question: what happens if a message fails to send? does it keep retrying until sending it? or do we risk getting the binlog position stuck?

I remember talking about/reviewing smt related to replicator/binlog position, maybe the comments in this commit will help? https://github.com/zendesk/maxwell/commit/569ce244 (smt about replicator thread being restarted and picking up the right position?)

osheroff:
@dadah89 the binlog position can and should get stuck if we don't deliver the message.. but I should look into what actually happens.  It's not all that great to keep writing way ahead for one partition and then have to backtrack and send all the data over again.

dadah89:
worst case, you could get stuck until there is no more binlog file =) would be nice to have some limit to how much it can lag behind

osheroff:
well like I said to @dasch, eventually the kafka buffer for that partition is gonna fill up and the replicator will stop.  Hard to know the "right" amount to backpressure is really, but I should try to get a jepsen-y thing setup locally and see how maxwell reacts in various situations.

dadah89:
ah didn't realize it was a similar question. If it stops once the buffer fills up, seems fine to me

PR #353, opened 2016-05-31
further on at-least-once delivery

osheroff:
I was tinkering this weekend and figured out that if a message errored we sort of just continued on, not checkpointing binlog positions, until we maybe ran out of kafka buffer room or memory or whatever.

So I'm setting up some defaults -- if we try to send a message that's too big, we'll drop and log.  We'll do one retry by default.  Any other exception and we'll simply crash maxwell, as it's probably the safest thing to do. 

@dasch @zendesk/rules 

PS: I did discover a totally pathological case, where if you send `kill -STOP` to a broker, the producer gets stuck in a state in which it never times out and never errors.  I'm not sure how close to any real failure scenario this is.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -16,6 +16,7 @@ import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.errors.RecordTooLargeException;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.slf4j.Logger;
@@ -42,7 +43,14 @@ class KafkaCallback implements Callback {
 	@Override
 	public void onCompletion(RecordMetadata md, Exception e) {
 		if ( e != null ) {
-			e.printStackTrace();
+			if ( e instanceof RecordTooLargeException ) {
+				LOGGER.error("RecordTooLargeException @ " + position + " -- " + key);
+				LOGGER.error("Maxwell dropped a row because it was too large (" + e.getLocalizedMessage() + ")");
+				LOGGER.error("Considering raising max.request.size broker-side.");
+				inflightMessages.completeMessage(position);
+			} else {
+				throw new RuntimeException(e);
+			}
 		} else {
 			try {
 				if ( LOGGER.isDebugEnabled()) {
@@ -67,7 +75,8 @@ class KafkaCallback implements Callback {
 public class MaxwellKafkaProducer extends AbstractProducer {
 	static final Object KAFKA_DEFAULTS[] = {
 		"compression.type", "gzip",
-		"metadata.fetch.timeout.ms", 5000
+		"metadata.fetch.timeout.ms", 5000,
+		"retries", 1
 	};
 	private final InflightMessageList inflightMessages;
 	private final KafkaProducer<String, String> kafka;
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
     <dependency>
     	<groupId>org.apache.kafka</groupId>
     	<artifactId>kafka-clients</artifactId>
-    	<version>0.8.2.0</version>
+    	<version>0.8.2.2</version>
     </dependency>
     <dependency>
     	<groupId>org.apache.logging.log4j</groupId>


src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -42,7 +43,14 @@ public KafkaCallback(InflightMessageList inflightMessages, BinlogPosition positi
...
 	public void onCompletion(RecordMetadata md, Exception e) {
 		if ( e != null ) {
-			e.printStackTrace();
+			if ( e instanceof RecordTooLargeException ) {
+				LOGGER.error("RecordTooLargeException @ " + position + " -- " + key);
+				LOGGER.error("Maxwell dropped a row because it was too large (" + e.getLocalizedMessage() + ")");
+				LOGGER.error("Considering raising max.request.size broker-side.");
+				inflightMessages.completeMessage(position);
+			} else {
+				throw new RuntimeException(e);
```
osheroff: I'm not sure what else could (reasonably) come up here.  Maybe `NotLeaderForPartition` -- that happens with `retries=0`... maybe a network timeout error could come up here, but I'm not sure what all we'd want to rescue and drop here.

...

I'm also thinking about the implications of duplicated maxwell messages in these crash-and-rewind type of scenarios.  For idempotent consumers, no big deal, but otherwise they'd really want to maintain a buffer of the last-N transactions, de-duping.  Or maybe I should just do the damn work and have maxwell read the last N transactions off each partition and de-dupe (like we talked about @dasch).

osheroff:
ping

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
dasch: > I'm not sure what else could (reasonably) come up here.  Maybe `NotLeaderForPartition` -- that happens with `retries=0`... maybe a network timeout error could come up here, but I'm not sure what all we'd want to rescue and drop here.

I'm not sure how much the Java lib handles – in HC, there's a single exception that the sync producer will raise if messages cannot be delivered. I would err on the side of crashing, adding in failure handling as necessary. If you try it out on staging you'll get plenty chaos testing :-)

> I'm also thinking about the implications of duplicated maxwell messages in these crash-and-rewind type of scenarios.  For idempotent consumers, no big deal, but otherwise they'd really want to maintain a buffer of the last-N transactions, de-duping.  Or maybe I should just do the damn work and have maxwell read the last N transactions off each partition and de-dupe (like we talked about @dasch).

Yeah, you'd probably have to read _n_ messages, where _n_ is the batch size of the producer.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -42,7 +43,14 @@ public KafkaCallback(InflightMessageList inflightMessages, BinlogPosition positi
 	@Override
 	public void onCompletion(RecordMetadata md, Exception e) {
 		if ( e != null ) {
-			e.printStackTrace();
+			if ( e instanceof RecordTooLargeException ) {
+				LOGGER.error("RecordTooLargeException @ " + position + " -- " + key);
+				LOGGER.error("Maxwell dropped a row because it was too large (" + e.getLocalizedMessage() + ")");
+				LOGGER.error("Considering raising max.request.size broker-side.");
+				inflightMessages.completeMessage(position);
```
dasch: Would there be a point in logging the failed binlog positions to another Kafka topic in order to be able to go in and retry them later? Or would that be overkill?

dasch:
> PS: I did discover a totally pathological case, where if you send kill -STOP to a broker, the producer gets stuck in a state in which it never times out and never errors. I'm not sure how close to any real failure scenario this is.

I think that may have been what I ran into during the Kafka decommissioning snafu. I would file a bug on Kafka.

dasch:
:+1:

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -67,7 +75,8 @@ public void onCompletion(RecordMetadata md, Exception e) {
 public class MaxwellKafkaProducer extends AbstractProducer {
 	static final Object KAFKA_DEFAULTS[] = {
 		"compression.type", "gzip",
-		"metadata.fetch.timeout.ms", 5000
+		"metadata.fetch.timeout.ms", 5000,
+		"retries", 1
```
wushujames: Why retries = 1? If your delivery ever fails, then Maxwill drop the message and keep going on. Except it won't checkpoint binlog offsets. So how will you ever get back to the dropped message? By crashing/restarting?

osheroff: mostly because of `NotLeaderForPartition` problems.  Been seeing long GC pauses result in a node getting kicked out of leading partitions, which means that the first try to write the records results in a `NotLeaderForPartiton` error, which is not retried by default.  So the metadata is updated, but the row is senselessly lost. 

But I think you also misunderstand the semantics of `retries=1` -- if the second delivery attempt succeeds, it will call the completion handler as normal -- in which case we'll checkpoint the offset and continue.  If it fails, it should call the completion handler with an exception, at which point we'll crash and hopefully transmit the message when we restart.  

Maxwell should never drop the message and keep going without checkpointing -- see the change in the `onCompletion` handler to either willfully drop the message, or just `throw new RuntimeException`

PR #356, opened 2016-06-03
WIP: Schema store refactor

osheroff:
none of this is working, but it's my thinking along lines of moving some of the raw schema storage functionality out of `MaxwellReplicator`.  ideally the MaxwellReplicator has no idea about the nature of how its schemas are stored.  This will get us mileage towards:

- multiple maxwell replicators running inside the same jvm, operating at different positions (with different producers)
- alternative stores for schema and position (some folks would like to store that stuff inside kafka, others might even opt for local disk)

It's still not quite right, as `SchemaStore` still keeps state that's tied to the replicator (it needs access to the previous schema to create delta-snapshots).  Will keep working.  I think it's my reticence to tie a `BinlogPosition` directly to the `Schema` object that's killing all my metaphors.  But I dunno. 

cc @dasch @zendesk/rules


dasch:
Looks like the right direction.

wushujames:
I like the direction, too.

PR #357, opened 2016-06-05
Move filter building into config class

osheroff:
This gives error messages at the right point in the code, and stops `maxwellFilterException` from propogating out everywhere.

also move getReplicationConnectionPool to getReplicationConnection.

@zendesk/rules


lileeyao:
👍

PR #359, opened 2016-06-08
usage text was getting too long

osheroff:
it would dump all of --help, which would scroll off the screen and render the error message un-seeable.  Here we filter down to just the part of usage text that triggered the error message.

Gives you errors like:
```
alpha21:~/src/maxwell (friendlier_usage_errors) $ bin/maxwell --bootstrapper=hooo
please specify --bootstrapper=async|sync|none

Option                       Description
------                       -----------
--bootstrapper               bootstrapper type: async|sync|none. default: async
--bootstrapper_fetch_size    (deprecated)
--help                       display help
```


osheroff:
cc @zendesk/rules

dadah89:
👍

PR #364, opened 2016-06-15
Fix some bugs in the inflight-message code

osheroff:
Was trying some of this code out in a real way, trying to trace down some out-of-memory errors.  Found a couple of bugs that need fixing.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -47,7 +47,9 @@ class KafkaCallback implements Callback {
 				LOGGER.error("RecordTooLargeException @ " + position + " -- " + key);
 				LOGGER.error("Maxwell dropped a row because it was too large (" + e.getLocalizedMessage() + ")");
 				LOGGER.error("Considering raising max.request.size broker-side.");
-				inflightMessages.completeMessage(position);
+
+				if ( isTXCommit )
+					inflightMessages.completeMessage(position);
 			} else {
 				throw new RuntimeException(e);
 			}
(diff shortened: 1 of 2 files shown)


osheroff:
cc @dasch

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -45,9 +45,11 @@ public void onCompletion(RecordMetadata md, Exception e) {
 		if ( e != null ) {
 			if ( e instanceof RecordTooLargeException ) {
 				LOGGER.error("RecordTooLargeException @ " + position + " -- " + key);
-				LOGGER.error("Maxwell dropped a row because it was too large (" + e.getLocalizedMessage() + ")");
+				LOGGER.error(e.getLocalizedMessage());
 				LOGGER.error("Considering raising max.request.size broker-side.");
-				inflightMessages.completeMessage(position);
+
+				if ( isTXCommit )
+					inflightMessages.completeMessage(position);
```
dadah89: don't you need to set the new position in `context`? (similar to what is done below)

dasch:
Are multiple threads writing to the inflight message tracker concurrently? It would be interesting to see if you get congestion.

:+1:

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
osheroff: i do.  will fix.

osheroff:
@dasch the callback is executed on the kafka producer thread, so that can write concurrently with the maxwell producer

osheroff:
@dadah89 fixed comments

PR #367, opened 2016-06-20
Schema store refactor

osheroff:
This is the final resting place of https://github.com/zendesk/maxwell/pull/356.  It removes direct responsibility for Schema processing from MaxwellReplicator, which instead delegates the work of processing SQL changes and storing updates to the `SchemaStore` interface. 

This is done with an eye towards multiple consumers in the same JVM.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -12,28 +12,14 @@ import com.zendesk.maxwell.bootstrap.AbstractBootstrapper;
 import com.zendesk.maxwell.producer.AbstractProducer;
 import com.zendesk.maxwell.schema.Schema;
 import com.zendesk.maxwell.schema.SchemaCapturer;
-import com.zendesk.maxwell.schema.MysqlSavedSchema;
+import com.zendesk.maxwell.schema.MysqlSchemaStore;
 import com.zendesk.maxwell.schema.SchemaStoreSchema;
 import com.zendesk.maxwell.schema.ddl.InvalidSchemaError;
 
 public class Maxwell {
-	private MysqlSavedSchema savedSchema;
 	private MaxwellConfig config;
 	private MaxwellContext context;
 	static final Logger LOGGER = LoggerFactory.getLogger(Maxwell.class);
-
-	private void initFirstRun(Connection connection, Connection schemaConnection) throws SQLException, IOException, InvalidSchemaError {
-		LOGGER.info("Maxwell is capturing initial schema");
-		SchemaCapturer capturer = new SchemaCapturer(connection, this.context.getCaseSensitivity());
-		Schema schema = capturer.capture();
-
-		BinlogPosition pos = BinlogPosition.capture(connection);
-
-		this.savedSchema = new MysqlSavedSchema(this.context.getServerID(), this.context.getCaseSensitivity(), schema, pos);
-		this.savedSchema.save(schemaConnection);
-
-		this.context.setPosition(pos);
-	}
 
 	private void run(String[] argv) throws Exception {
 		this.config = new MaxwellConfig(argv);
@@ -55,15 +41,8 @@ public class Maxwell {
 
 			SchemaStoreSchema.handleMasterChange(schemaConnection, context.getServerID(), this.config.databaseName);
 
-			if ( this.context.getInitialPosition() != null ) {
-				String producerClass = this.context.getProducer().getClass().getSimpleName();
-
-				LOGGER.info("Maxwell is booting (" + producerClass + "), starting at " + this.context.getInitialPosition());
-
-				this.savedSchema = MysqlSavedSchema.restore(schemaConnection, this.context);
-			} else {
-				initFirstRun(connection, schemaConnection);
-			}
+			String producerClass = this.context.getProducer().getClass().getSimpleName();
+			LOGGER.info("Maxwell is booting (" + producerClass + "), starting at " + this.context.getInitialPosition());
 		} catch ( SQLException e ) {
 			LOGGER.error("SQLException: " + e.getLocalizedMessage());
 			LOGGER.error(e.getLocalizedMessage());
@@ -73,7 +52,8 @@ public class Maxwell {
 		AbstractProducer producer = this.context.getProducer();
 		AbstractBootstrapper bootstrapper = this.context.getBootstrapper();
 
-		final MaxwellReplicator p = new MaxwellReplicator(this.savedSchema, producer, bootstrapper, this.context, this.context.getInitialPosition());
+		MysqlSchemaStore mysqlSchemaStore = new MysqlSchemaStore(this.context);
+		final MaxwellReplicator p = new MaxwellReplicator(mysqlSchemaStore, producer, bootstrapper, this.context, this.context.getInitialPosition());
 
 		bootstrapper.resume(producer, p);
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
@@ -22,6 +22,7 @@ import com.zendesk.maxwell.producer.AbstractProducer;
 import com.zendesk.maxwell.schema.Schema;
 import com.zendesk.maxwell.schema.MysqlSavedSchema;
 import com.zendesk.maxwell.schema.Table;
+import com.zendesk.maxwell.schema.SchemaStoreException;
 import com.zendesk.maxwell.schema.ddl.SchemaChange;
 import com.zendesk.maxwell.schema.ddl.ResolvedSchemaChange;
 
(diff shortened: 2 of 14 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -226,19 +226,30 @@ private void executeColumnInsert(Connection conn, ArrayList<Object> columnData)
...
+	private List<ResolvedSchemaChange> parseDeltas(String json) {
 		if ( json == null )
 			return null;
 
-		return mapper.readerFor(listOfResolvedSchemaChangeType).readValue(json.getBytes());
+		try {
+			return mapper.readerFor(listOfResolvedSchemaChangeType).readValue(json.getBytes());
+		} catch ( IOException e ) {
+			throw new RuntimeException("couldn't parse json delta: " + json.getBytes(), e);
+		}
```
osheroff: Stopped this IOException from leaking up the stack, as it's not gonna be thrown 99.999% of the time.  Best to just crash.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -306,19 +317,7 @@ private void restoreDerivedSchema(Connection conn, Long schema_id) throws SQLExc
...
-	private void restoreFrom(Connection conn, BinlogPosition targetPosition) throws SQLException, IOException, InvalidSchemaError {
-		Long schemaID = findSchema(conn, targetPosition, this.serverID);
-		if ( schemaID == null ) {
-			throw new InvalidSchemaError("Could not find schema for "
-					+ targetPosition.getFile() + ":" + targetPosition.getOffset());
-		}
-
-		restoreFromSchemaID(conn, schemaID);
-
-		handleVersionUpgrades(conn, schemaID, this.schemaVersion);
```
osheroff: moved this into the `restore` static method

osheroff:
cc @dasch

dasch:
:+1: to the degree that I can follow the code.

osheroff:
ping @dadah89 btw

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -315,56 +315,19 @@ protected BinlogEventV4 pollV4EventFromQueue() throws InterruptedException {
...
-
-		if ( !this.context.getReplayMode() ) {
-			try (Connection c = this.context.getMaxwellConnection()) {
-				this.savedSchema = this.savedSchema.createDerivedSchema(updatedSchema, p, changes);
-				this.savedSchema.save(c);
-			}
-
-			this.producer.writePosition(p);
-		}
+		this.producer.writePosition(position);
```
dadah89: Not sure the difference it makes, is it ok to change it to write position even if in replay mode?

osheroff: should be ok.  in replay mode the position store is a `ReadOnlyMysqlPositionStore`

src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -55,15 +41,8 @@ private void run(String[] argv) throws Exception {
...
-
-				LOGGER.info("Maxwell is booting (" + producerClass + "), starting at " + this.context.getInitialPosition());
-
-				this.savedSchema = MysqlSavedSchema.restore(schemaConnection, this.context);
-			} else {
-				initFirstRun(connection, schemaConnection);
-			}
+			String producerClass = this.context.getProducer().getClass().getSimpleName();
+			LOGGER.info("Maxwell is booting (" + producerClass + "), starting at " + this.context.getInitialPosition());
 		} catch ( SQLException e ) {
```
dadah89: suppose this still make sense? maybe smt like `SchemaStoreSchema.upgradeSchemaStoreSchema` can still throw those?

src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -73,7 +52,8 @@ private void run(String[] argv) throws Exception {
 		AbstractProducer producer = this.context.getProducer();
 		AbstractBootstrapper bootstrapper = this.context.getBootstrapper();
 
-		final MaxwellReplicator p = new MaxwellReplicator(this.savedSchema, producer, bootstrapper, this.context, this.context.getInitialPosition());
+		MysqlSchemaStore mysqlSchemaStore = new MysqlSchemaStore(this.context);
```
dadah89: does `MysqlSavedSchema.restore` throw `SQLException`? (asking as it would then just log and return, wondering if we want the same behavior with the new code... although I guess it'd need to rescue the run loop, and now it gets triggered by `getSchema`, which can happen in many places I suppose)

dadah89: (guess in that case it'd be `SchemaStoreException`)

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -226,19 +226,30 @@ private void executeColumnInsert(Connection conn, ArrayList<Object> columnData)
 		columnData.clear();
 	}
 
-	public static MysqlSavedSchema restore(Connection connection, MaxwellContext context) throws SQLException, IOException, InvalidSchemaError {
-		MysqlSavedSchema s = new MysqlSavedSchema(context.getServerID(), context.getCaseSensitivity());
+	public static MysqlSavedSchema restore(MaxwellContext context, BinlogPosition targetPosition) throws SQLException, InvalidSchemaError {
+		try ( Connection conn = context.getMaxwellConnection() ) {
+			Long schemaID = findSchema(conn, targetPosition, context.getServerID());
+			if (schemaID == null)
+				return null;
```
dadah89: Before it'd raise... so I suppose MysqlSchemaStore will just capture a new schema for the same binlong initial position? does `null` here means maxwell is in a bad state (saved position but missing schema) or does it mean that potentially maxwell was down for too long, and updates were lost?

dadah89: (so where we'd raise before, I think it'd be nice now do logger.error or warn, explaining in what kind of invalid state we were and that we recovered)

dadah89: (hm maybe it isn't possible anymore b/c context now defaults to capturing a new binlog position, which case findSchema would return empty... if not possible to log error, then it's fine)

src/main/java/com/zendesk/maxwell/schema/AbstractSchemaStore.java
```diff
@@ -0,0 +1,56 @@
...
+	static final Logger LOGGER = LoggerFactory.getLogger(AbstractSchemaStore.class);
+	protected final MaxwellContext context;
+
+	protected AbstractSchemaStore(MaxwellContext context) {
+		this.context = context;
+	}
+
+	protected Schema captureSchema() throws SQLException {
+		try(Connection connection = context.getReplicationConnection()) {
+			LOGGER.info("Maxwell is capturing initial schema");
```
dadah89: that line should probably go in MysqlSchemaStore#restoreOrCaptureSchema

dadah89:
Checking where the old code moved to, looks to me it''ll work just fine. I think there one situation we used to crash and now we just recover, which seemed fine to me as well. 👍

src/main/java/com/zendesk/maxwell/Maxwell.java
osheroff: yeah, I think lots of methods in here still throw that.  `ensureMaxwellSchema` for one.

src/main/java/com/zendesk/maxwell/Maxwell.java
osheroff: I think it's the same.  The only reason that we catch & return `SQLException` (instead of just letting `main` log it before exiting) is that it's very common to misconfigure your SQL connection, so we generally don't throw a stack trace in there.

src/main/java/com/zendesk/maxwell/schema/AbstractSchemaStore.java
osheroff: you mean just the log line?  No, I think if we had a `KafkaSchemaStore` it'd want all this code too.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
osheroff: yeah, I went over the code; your analysis is correct, this section of the code doesn't know anymore whether it's restoring a binlog position that was stored in the db or whether it's a brand new position.  I think that's.. mostly ok.  hard to say what to do if we get a binlog position but the schema goes missing.  I suppose it's as good as anything to recapture it.

Issue #362, opened 2016-06-13
Lower casing field names


daledude:
Looks like lower casing was forced in #51. Is it possible to have a cli arg to disable this? I have a number of systems that require the case be the same as in the schema. I could remap it in my code but that would mean maintaining a mapping and conversion.

Thanks for consideration.

osheroff:
yeah, I think the ideal behavior is to try to preserve the case that the columns were created with, but always do case-insensitive comparisons on lookup and such.

PR #377, opened 2016-07-10
#362: preserve case of columns as they came in

osheroff:
Instead of downcasing all fields as they come in, preserve the original casing of the field, only doing case-insensitive compares on lookup.

primary keys stay downcased; it's probably better to keep kafka keys static than to preserve original casing there.

@zendesk/rules 

note: some sections best viewed with `?w=1`

### DIFF
--- a/src/main/java/com/zendesk/maxwell/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/RowMap.java
@@ -105,7 +105,7 @@ public class RowMap implements Serializable {
 				if ( data.containsKey(pk) )
 					pkValue = data.get(pk);
 
-				g.writeObjectField("pk." + pk, pkValue);
+				g.writeObjectField("pk." + pk.toLowerCase(), pkValue);
 			}
 		}
 
@@ -128,7 +128,7 @@ public class RowMap implements Serializable {
 				pkValue = data.get(pk);
 
 			g.writeStartObject();
-			g.writeObjectField(pk, pkValue);
+			g.writeObjectField(pk.toLowerCase(), pkValue);
 			g.writeEndObject();
 		}
 		g.writeEndArray();
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -9,7 +9,7 @@ import com.fasterxml.jackson.databind.JavaType;
 import com.zendesk.maxwell.CaseSensitivity;
 import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.schema.columndef.*;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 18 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -513,17 +507,43 @@ private void fixUnsignedColumns(Connection conn) throws SQLException, InvalidSch
...
+					Database db = sc.capture().findDatabase("mysql");
+					this.schema.addDatabase(db);
+					this.shouldSnapshotNextSchema = true;
+				}
+
+				fixUnsignedColumns(recaptured);
 			}
 
-			fixUnsignedColumns(conn);
+			if ( this.schemaVersion < 2 ) {
```
dadah89: Don't you need to bump `SchemaStoreVersion` for this to make sense?

osheroff: ah, yeah.  I do.

src/main/java/com/zendesk/maxwell/schema/columndef/StringColumnDef.java
```diff
@@ -81,7 +80,7 @@ public Object asJSON(Object value) {
 	}
 
 	private String quoteString(String s) {
-		String escaped = StringEscapeUtils.escapeSql(s);
+		String escaped = s.replaceAll("'", "''");
```
dadah89: hm from `StringEscapeUtils.escapeSql` docs those are equivalent

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -105,7 +105,7 @@ private String pkToJsonHash() throws IOException {
 				if ( data.containsKey(pk) )
```
dadah89: wasn't this check working before, or does `pk` need to be lower case here as well?

dadah89: (also applies to `pkAsConcatString`)

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -105,7 +105,7 @@ private String pkToJsonHash() throws IOException {
...
 					pkValue = data.get(pk);
 
-				g.writeObjectField("pk." + pk, pkValue);
+				g.writeObjectField("pk." + pk.toLowerCase(), pkValue);
```
dadah89: if it's lower casing `pk` here, why is `Table` `normalizedPKColumnNames` needed? (couldn't find other places where `Table#getPKList` is being used)

osheroff: so the change is:

- Columns now are CaSed
- Primary keys are still output as lower_cased

so in order to get the value for a PK row, the PK is stored inside the `Table` object as "CaSed" (by `normalizedPKColumns`), looked up in `RowMap#data` as "CaSed", but then output as "cased".

dadah89: hm so `data` contains `CaSed` keys, `normalizedPKColumns ` makes sense then

dadah89:
👍 after bumping `SchemaStoreVersion`

Issue #371, opened 2016-06-29
mysql datatype alias SERIAL aborts maxwell producer


joel-privalia:
A create table statement using the SERIAL datatype, which mysql (5.5) implicitly converts to BIGINT UNSIGNED breaks the maxwell producer:

```
---- customers (TABLE)
COLUMN_NAME | DATA_TYPE       | PK | NULLABLE | DEFAULT | AUTOINCREMENT | COMPUTED | REMARKS | POSITION
------------+-----------------+----+----------+---------+---------------+----------+---------+---------
id          | BIGINT UNSIGNED | NO | NO       |         | YES           | NO       |         |        1
first_name  | VARCHAR(255)    | NO | YES      |         | NO            | NO       |         |        2
last_name   | VARCHAR(255)    | NO | YES      |         | NO            | NO       |         |        3


13:50:13,355 INFO  Maxwell - Maxwell is booting (MaxwellKafkaProducer), starting at BinlogPosition[binlog.000003:30343]
13:50:13,356 INFO  ProducerConfig - ProducerConfig values: 
        compression.type = gzip
        metric.reporters = []
        metadata.max.age.ms = 300000
        metadata.fetch.timeout.ms = 5000
        acks = 1
        batch.size = 16384
        reconnect.backoff.ms = 10
        bootstrap.servers = [localhost:9092]
        receive.buffer.bytes = 32768
        retry.backoff.ms = 100
        buffer.memory = 33554432
        timeout.ms = 30000
        key.serializer = class org.apache.kafka.common.serialization.StringSerializer
        retries = 1
        max.request.size = 1048576
        block.on.buffer.full = true
        value.serializer = class org.apache.kafka.common.serialization.StringSerializer
        metrics.sample.window.ms = 30000
        send.buffer.bytes = 131072
        max.in.flight.requests.per.connection = 5
        metrics.num.samples = 2
        linger.ms = 0
        client.id = 

13:50:13,360 WARN  ProducerConfig - The configuration value.serializer = class org.apache.kafka.common.serialization.StringSerializer was supplied but isn't a known config.
13:50:13,360 WARN  ProducerConfig - The configuration key.serializer = class org.apache.kafka.common.serialization.StringSerializer was supplied but isn't a known config.
13:50:13,480 INFO  TransportImpl - connecting to host: 172.16.215.19, port: 3306
[...]

PR #379, opened 2016-07-10
#371 - handle SERIAL datatype properly

osheroff:
@zendesk/rules


dadah89:
```
SERIAL is an alias for BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE.

SERIAL DEFAULT VALUE in the definition of an integer column is an alias for NOT NULL AUTO_INCREMENT UNIQUE.
```
I see `NOT NULL` matches the `nullability` definition, but from `column_options` the only think we seem to care is `primary_key`?

osheroff:
that's correct.  but I think we roughly match the mysql parser by accepting all column options (and then ignoring the ones we don't care about).

Makefile
```diff
@@ -17,6 +17,9 @@ ANTLR_IMPORTS=src/main/antlr4/imports
 ANTLR_DIR=target/generated-sources/src/main/antlr4/com/zendesk/maxwell/schema/ddl
 ANTLR_OUTPUT=$(ANTLR_DIR)/mysqlBaseListener.java $(ANTLR_DIR)/mysqlLexer.java $(ANTLR_DIR)/mysqlListener.java $(ANTLR_DIR)/mysqlParser.java
 
+$(ANTLR_IMPORTS)/mysql_literal_tokens.g4: $(ANTLR_IMPORTS)/generate_tokens.rb
```
dadah89: this step was executed manually before this change?

dadah89:
lgtm 👍

Makefile
osheroff: yeah

PR #392, opened 2016-07-25
refactor configuration.

osheroff:
I'm prepping to have multiple maxwells running inside the same JVM.

This PR is best viewed side-by-side if you ask me; too many unrelated changes in the same place.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -171,7 +171,10 @@ public class MaxwellConfig extends AbstractConfig {
 		this.kafkaPartitionKey  = fetchOption("kafka_partition_by", options, properties, "database");
 		this.kafkaPartitionHash = fetchOption("kafka_partition_hash", options, properties, "default");
 
-		this.kafkaProperties.setProperty("bootstrap.servers", fetchOption("kafka.bootstrap.servers", options, properties, null));
+		String kafkaBootstrapServers = fetchOption("kafka.bootstrap.servers", options, properties, null);
+		if ( kafkaBootstrapServers != null )
+			this.kafkaProperties.setProperty("bootstrap.servers", kafkaBootstrapServers);
+
 		for ( Enumeration<Object> e = properties.keys(); e.hasMoreElements(); ) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -124,52 +123,71 @@ private String parseLogLevel(String level) {
...
+			return (String) options.valueOf(name);
+		else if ( (properties != null) && properties.containsKey(name) )
+			return (String) properties.getProperty(name);
+		else
+			return defaultVal;
+	}
+
+	private MaxwellMysqlConfig parseMysqlConfig(String prefix, OptionSet options, Properties properties) {
+		MaxwellMysqlConfig config = new MaxwellMysqlConfig();
+		config.host     = fetchOption(prefix + "host", options, properties, null);
```
osheroff: default *is* actually localhost, but we spit out a warning down in `validate`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -124,52 +123,74 @@ private String parseLogLevel(String level) {
...
 
 		if ( options.has("help") )
 			usage("Help for Maxwell:");
 
-		if ( options.has("log_level")) {
-			this.log_level = parseLogLevel((String) options.valueOf("log_level"));
-		}
-
-		this.maxwellMysql.parseOptions("", options);
+		this.log_level = fetchOption("log_level", options, properties, null);
```
dadah89: wondering if setting `properties` and `options` to instance variables (and unsetting them before leaving `parse`) so they won't be repeated with all the method calls wouldn't look better

dadah89:
lgtm 👍

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: could go down that road, but the right there would be to be like:

```
  new MaxwellConfigOptionParser(this, options, properties).parse()
```

or something.  Leaving instance vars around just to not pass them into other methods (ie they'll get thrown away at the end of the method) is IMO kinda bad juju.

PR #393, opened 2016-07-25
Give the position thread a client_id

osheroff:
Towards enabling multiple maxwell instances pointing at one server, we give the position store a `client_id` that identifies unique instances of maxwell. 

@zendesk/rules


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -330,6 +339,10 @@ private void validate() {
 			this.databaseName = "maxwell";
 		}
 
+		if ( this.clientID == null ) {
+			this.clientID = "maxwell";
```
dadah89: would be better to make this unique so we can identify from which maxwell instance this is coming from? or we don't care that much?
```
An id string to pass to the server when making requests. The purpose of this is to be able to track the source of requests beyond just ip/port by allowing a logical application name to be included in server-side request logging.
```

src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
```diff
@@ -115,30 +119,15 @@ public synchronized void set(BinlogPosition p) {
...
-	}
-
 	public BinlogPosition get() throws SQLException {
 		BinlogPosition p = position.get();
 		if ( p != null )
 			return p;
 
 		try ( Connection c = getConnection() ) {
-			PreparedStatement s = c.prepareStatement("SELECT * from `positions` where server_id = ?");
+			PreparedStatement s = c.prepareStatement("SELECT * from `positions` where server_id = ? and client_id = ?");
```
dadah89: hm there is one maxwell running per shard, if they all share the same client_id, won't it cause issues here?

osheroff: one maxwell per *cluster*.  And they'll have to be configured with separate client ids.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: this is different than the kafka client_id.

dadah89:
would be nice to comment somewhere in the help text/docs that if you change the client id, you loose the binlog position (and maybe emphasize that they *should* be unique)

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -119,21 +122,43 @@ public static void upgradeSchemaStoreSchema(Connection c, String schemaDatabaseN
...
 			performAlter(c, "alter table `schemas` add column deltas mediumtext charset 'utf8' NULL default NULL after base_schema_id");
 
-		if ( !getTableColumns("schemas", c).containsKey("version")) {
+		if ( !schemaColumns.containsKey("version")) {
 			performAlter(c, "alter table `schemas` add column `version` smallint unsigned not null default 0 after `charset`");
 		}
+
+		if ( !getTableColumns("positions", c).containsKey("client_id") ) {
+			performAlter(c, "alter table `positions` add column `client_id` varchar(255) charset 'latin1' not null default 'maxwell'");
+			performAlter(c, "alter table `positions` add column drop primary key, add primary key(`server_id`, `client_id`)");
```
dadah89: is this valid syntax, `add column drop primary key`?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -119,21 +122,43 @@ public static void upgradeSchemaStoreSchema(Connection c, String schemaDatabaseN
...
+		}
+
+		if ( !schemaColumns.containsKey("position_sha") ) {
+			performAlter(c, "alter table `schemas` add column `position_sha` char(40) charset 'latin1' null default null, add unique index(`position_sha`)");
```
dadah89: is it supposed to be `latin1`? asking because `maxwell_schema` has columns with `utf8` charset...

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
osheroff: no, it's not.  I'll have to add some tests around this code.

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
osheroff: it's a hex-digest of a SHA.  doing it as `latin` makes the size of the unique index managable, and there's no chance of high-bit utf8 here.

dadah89:
👍 after merging/rebasing master and build is green

Issue #380, opened 2016-07-12
unhandled character set


fly3q:
09:22:44,890 INFO  AbstractSchemaStore - Maxwell is capturing initial schema
09:22:55,150 WARN  StringColumnDef - warning: unhandled character set 'gbk'
java.lang.NullPointerException: charset
        at java.lang.String.<init>(String.java:451)
        at java.lang.String.<init>(String.java:505)
        at com.zendesk.maxwell.schema.columndef.StringColumnDef.asJSON(StringColumnDef.java:79)
        at com.zendesk.maxwell.ColumnWithDefinition.asJSON(ColumnWithDefinition.java:27)
        at com.zendesk.maxwell.MaxwellAbstractRowsEvent.jsonMaps(MaxwellAbstractRowsEvent.java:204)
        at com.zendesk.maxwell.MaxwellReplicator.getTransactionRows(MaxwellReplicator.java:220)
        at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:301)
        at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:114)
        at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:31)
        at com.zendesk.maxwell.Maxwell.run(Maxwell.java:76)
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:82)
09:22:55,153 INFO  MysqlPositionStore - Storing final position: BinlogPosition[bin.000006:501938]

PR #395, opened 2016-07-25
issue #380 -- support a bunch more character encodings

osheroff:
those mysql encodings that just naturally line up with java's, we support.

@zendesk/rules


dadah89:
could you add a test that executes the default case with a valid charset?

osheroff:
@dadah89 added a some tests.

dadah89:
👍

PR #398, opened 2016-08-04
stop trying to be clever about exceptions

osheroff:
the original intention of throwing RuntimeException was to crash
maxwell, but that's not what happened.  What happened was that the kafka
producer ended up eating the error, leaving us in a state where we were
keeping endless amounts of InflightMessage objects, as well as not
advancing the binlog poisition.

@zendesk/rules


dadah89:
the kafka producer was catching the `RuntimeException`?

osheroff:
yeah.  it's surprisingly hard to crash a java process unless you want to call `exit` (which I don't)

dadah89:
My concern was letting maxwell advance the binlog position without producing messages, 'losing' them. So we agree that kafka producer settings should be used for that, by settings `acks`, `retries`, `max.in.flight.requests.per.connection` it is possible to choose between availability and consistency, in/out of order message delivery, and messages retries (setting `retry` to max basically means maxwell stalls if it can produce messages).
👍

Issue #387, opened 2016-07-18
partitioned tables


smferguson:
Maxwell currently blows up when it encounters a partitioned table.
Does it need to know about the schema? I'm wondering if it would be possible/not too complex to add an option to skip ddl processing. It feels like a much more straightforward change than parsing partitioned table ddl.

osheroff:
Yeah, Maxwell needs to know about the schema.  The binlog itself only contains columns in arrays (no lovely column names for you!).  I think ultimately will have to bite the bullet and parse the partition statements; maxwell doesn't care about them, but they have to be parsed even so to get at the DDL that it needs.

smferguson:
ahh. thank you!

osheroff:
You're welcome :)
Going to leave this issue open until the PR merges and I release the fixes, otherwise I tend to forget stuff.

smferguson:
makes sense. sorry for closing!

osheroff:
fixed in v1.1.5

PR #399, opened 2016-08-06
Support for parsing up PARTITION table ddl

osheroff:
we use the same trickery we used in parsing the CHECK() function; we don't want to bother parsing a complex mysql expression, so we hint to the parser that we want to skip a matching set of parenthesis, and the parser does a token-stream rewrite and starts the parse over again.

Note that bringing in the partition sql (generated out of the mysql-server test suite) also exposed some issues in the string literal tokenizer.

@zendesk/rules


osheroff:
addresses issue #387

src/main/antlr4/imports/mysql_partition.g4
```diff
@@ -0,0 +1,68 @@
+grammar mysql_partition;
+
+import mysql_idents;
+
+partition_by:
+  PARTITION BY
+  partition_by_what
+  partition_count?
+  subpartition_by?
```
dadah89: missing `[SUBPARTITIONS num]`?

dadah89: oh nvm, part of `subpartition_by`

src/main/antlr4/imports/mysql_partition.g4
```diff
@@ -0,0 +1,68 @@
...
+  ;
+
+partition_by_what:
+    LINEAR? HASH skip_parens
+  | LINEAR? KEY (ALGORITHM '=' INTEGER_LITERAL)? skip_parens
+  | RANGE COLUMNS? skip_parens
+  | LIST COLUMNS? skip_parens;
+
+subpartition_by:
+  SUBPARTITION BY partition_by_what (SUBPARTITIONS INTEGER_LITERAL)?;
```
osheroff: @dadah89 the nubmer of subpartitions is here.

src/main/antlr4/imports/mysql_partition.g4
```diff
@@ -0,0 +1,68 @@
+grammar mysql_partition;
+
+import mysql_idents;
+
+partition_by:
+  PARTITION BY
+  partition_by_what
+  partition_count?
+  subpartition_by?
+  partition_definitions?
```
dadah89: `skip_parens` doesn't work here? guess it could mach something that it shouldn't?

osheroff: it.. might, if it was `skip_parens?`  generally I've tried to only use the skip-parens code when I'm *sure* there's parenthesis to be skipped (ie they're anchored to another expression).  The whole partition definition stuff here is optional, so yeah, could maybe be something else.

dadah89:
lgtm 👍

PR #400, opened 2016-08-06
fix parsing issue for CHARSET ASCII

osheroff:
"ASCII" here, if unquoted, was coming in as the ASCII token rather than becoming part of `charset_name`.

@zendesk/rules


dadah89:
👍

PR #402, opened 2016-08-09
hack together support for putting NULL fields in by default

osheroff:
@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -93,7 +93,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts( "output_binlog_position", "produced records include binlog position; [true|false]. default: false" ).withOptionalArg();
 		parser.accepts( "output_commit_info", "produced records include commit and xid; [true|false]. default: true" ).withOptionalArg();
-		parser.accepts( "output_omit_null", "produced records exclude fields with NULL values [true|false]. default: false" ).withOptionalArg();
+		parser.accepts( "output_nulls", "produced records include fields with NULL values [true|false]. default: true" ).withOptionalArg();
 
 		parser.accepts( "__separator_5" );
 
@@ -276,7 +276,7 @@ public class MaxwellConfig extends AbstractConfig {
 		this.outputConfig = new MaxwellOutputConfig();
 		outputConfig.includesBinlogPosition = fetchBooleanOption("output_binlog_position", options, properties, false);
 		outputConfig.includesCommitInfo = fetchBooleanOption("output_commit_info", options, properties, true);
-		outputConfig.omitNull = fetchBooleanOption("output_omit_null", options, properties, true);
+		outputConfig.includesNulls = fetchBooleanOption("output_nulls", options, properties, true);
 	}
 
 	private Properties parseFile(String filename, Boolean abortOnMissing) {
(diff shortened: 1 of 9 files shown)


dadah89:
failing tests

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -212,6 +214,12 @@ private void parse(String [] argv) {
 		if ( options.has("replay")) {
 			this.replayMode = true;
 		}
+
+		if ( properties.contains("omit_nulls") )
```
dadah89: fetchOption?

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -215,7 +217,7 @@ public String toJSON() throws IOException {
 			}
 		}
 
-		writeMapToJSON("data", this.data, false);
+		writeMapToJSON("data", this.data, !this.omitNulls);
 
 		if ( !this.oldData.isEmpty() ) {
 			writeMapToJSON("old", this.oldData, true);
```
dadah89: missed passing the flag here

src/main/java/com/zendesk/maxwell/schema/columndef/EnumColumnDef.java
```diff
@@ -29,7 +29,7 @@ private String asString(Object value) {
 		Integer i = (Integer) value;
 
 		if ( i == 0 )
-			return null;
+			return "";
```
dadah89: why was this necessary?

src/test/java/com/zendesk/maxwell/RowMapBufferTest.java
```diff
@@ -13,9 +13,9 @@ public void TestOverflowToDisk() throws Exception {
 		RowMapBuffer buffer = new RowMapBuffer(2, 250); // allow about 250 bytes of memory to be used
 
 		RowMap r;
-		buffer.add(new RowMap("insert", "foo", "bar", 1L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1")));
-		buffer.add(new RowMap("insert", "foo", "bar", 2L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1")));
-		buffer.add(new RowMap("insert", "foo", "bar", 3L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1")));
+		buffer.add(new RowMap("insert", "foo", "bar", 1L, new ArrayList<String>(), new BinlogPosition(3, "mysql.1"), true));
```
dadah89: suppose you need a fixture change (expected result) for this to pass?

src/main/java/com/zendesk/maxwell/schema/columndef/EnumColumnDef.java
osheroff: this emulates what mysql does when you pass it an invalid enum; it's stored internally as a "0" but when SELECTed it appears as the empty string.

src/main/java/com/zendesk/maxwell/RowMap.java
osheroff: that's intentional: `oldData` only contains data that's changed in the row, and we *always* put a null value in there if the row was null.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: would have to write a version of `fetchBooleanOption`, which is possible.

xmlking:
Any chance this PR will  be released in near time ?
Thanks

osheroff:
@xmlking yeah @dadah89 did a cleaner PR around the area of controlling the row format.  I re-coded for that work and pushed here.

osheroff:
@dadah89 code-review here at your leisure

xmlking:
I thing  `output_null=true` is more readable then  `output_omit_null=false`

dadah89:
changing the flag name makes sense to me, up to you
lgtm 👍

osheroff:
ok @xmlking good suggestion, updated

osheroff:
@dadah89 one last CR?

dadah89:
👍 !

PR #404, opened 2016-08-11
(master-recovery, part 1) Refactor position store into two components

osheroff:
split apart the SQL work from the thread-loop of the position store. 

Also, make sure that `MaxwellConfig` always gets setup with sane defaults.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
+++ b/src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
@@ -71,13 +71,11 @@ public class PositionStoreThread extends RunLoopProcess implements Runnable {
 			position = p;
 	}
 
-	public BinlogPosition getPosition() throws SQLException {
+	public synchronized BinlogPosition getPosition() throws SQLException {
 		if ( position != null )
 			return position;
 
-		synchronized(this) {
-			position = store.get();
-		}
+		position = store.get();
 
 		return position;
 	}
--- a/src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
+++ b/src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
@@ -1,0 +1,87 @@
+package com.zendesk.maxwell.schema;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.sql.SQLException;
+import java.util.concurrent.TimeoutException;
+
+import com.zendesk.maxwell.BinlogPosition;
+import com.zendesk.maxwell.RunLoopProcess;
+
+public class PositionStoreThread extends RunLoopProcess implements Runnable {
+	static final Logger LOGGER = LoggerFactory.getLogger(PositionStoreThread.class);
+	private BinlogPosition position; // in memory position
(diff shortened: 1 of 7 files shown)


src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
```diff
@@ -0,0 +1,86 @@
...
+			store.set(newPosition);
+			storedPosition = newPosition;
+		}
+
+		try { Thread.sleep(1000); } catch (InterruptedException e) { }
+	}
+
+	public synchronized void setPosition(BinlogPosition p) {
+		if ( position == null || p.newerThan(position) )
+			position = p;
```
dadah89: b/c this is synchronized we don't need `AtomicReference`?

osheroff: i think so.  this stuff is pretty hard to really reason about properly, the only problems I see are maybe that it could race against Line 78.  LMS about that.

src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
```diff
@@ -0,0 +1,86 @@
...
+				LOGGER.error("error storing final position: " + e);
+			}
+		}
+	}
+
+	public void work() throws Exception {
+		BinlogPosition newPosition = position;
+
+		if ( newPosition != null && newPosition.newerThan(storedPosition) ) {
+			store.set(newPosition);
```
dadah89: before it'd catch `SQLException` and stop the loop, now will it just keep going on? is that ok?

osheroff: the logic just moved to `run`

src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
osheroff: > Second, when a synchronized method exits, it automatically establishes a happens-before relationship with any subsequent invocation of a synchronized method for the same object. This guarantees that changes to the state of the object are visible to all threads.

https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html

so yeah, this is racy against line 78, but anyone calling `setPosition` will not clobber an old result due to it being synced.

osheroff: fixed line 78

src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
dadah89: should it call `stopLoop` then? otherwise it'll just start the loop again (from what I understood from RunLoopProcess code)

osheroff: the chain is this:
```
run
  runLoop
     work
```

runLoop has a `finally` block that sets its state to `STOPPED` at the end

dadah89: oh got confused with catch in the `run` method, I think you're right

dadah89:
lgtm 👍

PR #405, opened 2016-08-12
(master-recovery, part 2) Add a heartbeat and replica_server_id

osheroff:
This is part 2 of a many part PR series.

The plot thickened a bit with regards to running multiple copies of maxwell, and also with regard to doing master recovery.  

For multiple copies of Maxwell run on 5.5, each must have a unique server_id or mysql will unceremoniously cut the connection.

@zendesk/rules

PR #406, opened 2016-08-12
(master-recovery, part 3) add last_heartbeat_read into maxwell.positions

osheroff:
My first thought when it came to master recovery was:

- simply find the latest binlog entry on the slave where the `server_id` comes from the old master.  

this works, as long as Maxwell wasn't lagged at the time of the master switch.  

Second thought:  Ok, we output a heartbeat value whenever we write to `maxwell.positions`.  Find wherever we wrote that heartbeat value in the new master's binlog, and use that position.  

This also doesn't work with a lagged slave.  The important note is that we're writing two bits of data whenever we write to `maxwell.positions`.   The first write goes *immediately* to the table on disk.  The second write goes asynchronously to the (possibly lagged) Maxwell.  So maxwell may write a heartbeat to the *end* of the binlog saying "I'm at a binlog position that's days behind the current position".  So just matching heartbeats doesn't work.

The third idea, which should work and is represented here, is to read our own writes -- the "last_heartbeat_read" column in `maxwell.positions` represents the last heartbeat value that we've actually sent to the producer.  We should be able to scan backwards through a new master's log, find this heartbeat, and pick up replication again with no more than 1 second or so of overlap. 

@zendesk/rules


dadah89:
is it calling `heartbeat` from #405 anywhere?

osheroff:
ping, backported the call to `heartbeat`

dadah89:
lgtm 👍 
as a follow up, would be nice adding a test checking some value gets saved to positions. last_heartbeat_read

dadah89:
what will happen when storage and replica hosts aren't the same?

```
Schema storage host vs replica host
Maxwell needs two sets of mysql permissions to operate properly: a mysql database in which to store schema snapshots, and a mysql host to replicate from. The recommended configuration is that these two functions are provided by a single mysql host. In this case, just specify host, user, etc.

Some configurations, however, may need to write data to a different server than it replicates from. In this case, the host described by host, user, ..., will be used to write schema information to, and Maxwell will replicate events from the host described by replication_host, replication_user, ... Note that bootstrapping is not available in this configuration.
```

osheroff:
merged by hand.

PR #407, opened 2016-08-12
Added kafka topic per table feature

sschatts:
This is feature takes in the format for the topic names 'namespace_%{database}_%{table}'. Added the flags in the config file, in MaxwellKafkaProducer.java added the logic to compile the correct kafka topic name, if the topic doesn't exist it will go to the Kafka topic. I added Zookeeper flags to be able to check if the kafka topic exists and if it doesn't add it to 'maxwell_needs_topic_created' instead.


osheroff:
ok, first things first is to fix the whitespace -- tabs instead of spaces in this project.

osheroff:
then, I'm curious as to what errors you get if you try to produce to a topic doesn't yet exist -- if kafka topic auto-creation is turned off, Maxwell would ideally simply crash.  

Even so, I would think that we could determine what topics exist simply with the `producer.partitionsFor(topic)` call, no?  If needed we can maintain the results of that call as a cache, not sure that's necessary tho.

sschatts:
@osheroff These are the errors I get when Maxwell produces to a topic that doesn't exist yet:

> ERROR MaxwellKafkaProducer - TimeoutException @ BinlogPosition[<database>-bin.004545:287582] -- {"database":..}
> ERROR MaxwellKafkaProducer - Failed to update metadata after 69 ms.

It doesn't crash. It just displays the error and continues. I agree ideally Maxwell would just crash after it tries to produce to a topic that doesn't exist.

You are right we can check if a topic exists or not from the call `if(this.kafka.partitionsFor(this.topic).isEmpty())`

sschatts:
Great, I was able to get Maxwell to crash when the topic does not exist and log an error about the topic name that does not existing.

Also, what is your tab size set to? The formatting seems fine until I commit it.

osheroff:
re: tabs; the point is to configured your editor to indent with the *tab* character, not with a number of spaces.  So my tab could be 4 spaces, yours could be 8, but it doesn't matter because it's all just the tab character.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -113,8 +116,28 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
...
+			
+			//check if the topic exists
+			List<PartitionInfo> partitions_for_kafka;
+			try {
+				partitions_for_kafka = this.kafka.partitionsFor(this.topic);
+			} catch(TimeoutException e) {
+				LOGGER.error("This topic name does not exist: " + this.topic + ": " + e.getLocalizedMessage());
+				throw (e);
+			}
+        }
```
osheroff: hoist all of this into a `getTopic()` private function

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -113,8 +116,28 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
...
+			String kafkaTopicWithFormat = original_topic_per_table.replaceAll("%\\{database\\}", db_name);
+			
+			//replace the %{table} in kafkaTopicPerTable with the table name and set the topic name
+			String table_name = r.getTable();
+			this.topic = kafkaTopicWithFormat.replaceAll("%\\{table\\}", table_name);
+			
+			//check if the topic exists
+			List<PartitionInfo> partitions_for_kafka;
+			try {
+				partitions_for_kafka = this.kafka.partitionsFor(this.topic);
```
osheroff: how fast is this?  is the kafka producer caching the results of this well?

sschatts:
`partitions_for_kafka = this.kafka.partitionsFor(this.topic);`
Does cache the results. In partitionsFor(), the function waitOnMetadata is called and that only gets the metadata if the fetch is empty.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -112,9 +115,35 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
 		this.inflightMessages = new InflightMessageList();
 	}
 
+	//getTopic updates the topic based on the topic per table format (declare by the --kafka_topic_per_table flag)
+	private void getTopic(RowMap r){
```
osheroff: better to have this function return the topic as a String (rather than mutating `this.topic`).

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -112,9 +115,35 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
...
+			String kafkaTopicWithFormat = original_topic_per_table.replaceAll("%\\{database\\}", db_name);
+
+			//replace the %{table} in kafkaTopicPerTable with the table name and set the topic name
+			String table_name = r.getTable();
+			this.topic = kafkaTopicWithFormat.replaceAll("%\\{table\\}", table_name);
+
+			//check if the topic exists
+			List<PartitionInfo> partitions_for_kafka;
+			try {
+				partitions_for_kafka = this.kafka.partitionsFor(this.topic);
```
osheroff: you'll want to extract the logic here into a function -- you'll need to pass the return of `partitionsFor` into the line here: https://github.com/zendesk/maxwell/pull/407/files#diff-5c1924a1e21c60a874772298d057e39eR150

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -123,14 +152,13 @@ public void push(RowMap r) throws Exception {
...
 
-
 		/* if debug logging isn't enabled, release the reference to `value`, which can ease memory pressure somewhat */
 		if ( !KafkaCallback.LOGGER.isDebugEnabled() )
 			value = null;
 
 		KafkaCallback callback = new KafkaCallback(inflightMessages, r.getPosition(), r.isTXCommit(), this.context, key, value);
 
-		kafka.send(record, callback);
+        kafka.send(record, callback);
```
osheroff: still a couple of whitespace issues

src/main/resources/sql/maxwell_schema_bootstrap.sql
```diff
@@ -8,6 +8,6 @@ CREATE TABLE IF NOT EXISTS `bootstrap` (
   started_at      DATETIME default NULL,
   completed_at    DATETIME default NULL,
   binlog_file varchar(255) default NULL,
-  binlog_position int unsigned default 0
-);
-
+  binlog_position int unsigned default 0,
+  KEY `is_complete` (`is_complete`,`started_at`)
```
osheroff: don't slip stuff like this into your PR.  it's fine to add an index, but do it as part of another logical PR.

osheroff:
Ok, I think we're close now.  Two more things:

- are we sure the only exception to be thrown by kafka when the topic doesn't exist is a `TimeoutError`?  (I will do some investigation of my own)
- I still think it's unnecessary to have a separate command line flag for this: if the value `of --kafka_topic` includes `%{database}` or `%{table}` we'll interpolate the values there.  sound good?

powerLambda:
Vote for this feature. "topic per table feature" is useful when play with kafka log compaction feature. 
any update on this PR?

osheroff:
@sschatts any progress here?

sschatts:
@osheroff I am back at school and working on this on the side. Hope to make some progress in the next two weeks.

osheroff:
hey @powerLambda, if you're interested here you may want to contribute some code.  It's close to the finish line but it looks like Sarah may be deep in the muck of school... :)

powerLambda:
@osheroff sure, will look into this PR soon. :+1:

smferguson:
@osheroff fyi, looking at this now.

osheroff:
finished in another pr.  thanks for the work @sschatts

PR #409, opened 2016-08-17
(master-recovery part 4) two refactors

osheroff:
1.  In order to be able to create the `maxwell` database on first-run, we had made it so that the connection pool returned a connection that had no database selected.  This is necessary for initialization, but meant we were passing the database name around everywhere.  This fixes that so there's a special connection pool we use to create the maxwell database, but everyone else uses a database that's properly wired up with a database name.

2.  Remove `MaxwellContext` from some low-level interfaces.  `MaxwellContext` is a super-convenient bundle of objects, but creating one is a pain, and makes re-use a pain.   Here we make a couple of objects (`MaxwellReplicator`, `SchemaStore`) usable with the component objects instead of the bundled `Context`.  

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellContext.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellContext.java
@@ -27,6 +27,7 @@ public class MaxwellContext {
 
 	private final ConnectionPool replicationConnectionPool;
 	private final ConnectionPool maxwellConnectionPool;
+	private final ConnectionPool maxwellConnectionPoolWithoutDatabase;
 	private final MaxwellConfig config;
 	private PositionStoreThread positionStoreThread;
 	private Long serverID;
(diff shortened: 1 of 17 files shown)


src/main/java/com/zendesk/maxwell/MaxwellContext.java
```diff
@@ -42,6 +43,9 @@ public MaxwellContext(MaxwellConfig config) {
 		this.replicationConnectionPool = new ConnectionPool("ReplicationConnectionPool", 10, 0, 10,
 				config.replicationMysql.getConnectionURI(), config.replicationMysql.user, config.replicationMysql.password);
 
+		this.rawMaxwellConnectionPool = new ConnectionPool("RawMaxwellConenectionPool", 1, 2, 100,
```
dadah89: typo Conenection

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaStore.java
```diff
@@ -1,23 +1,51 @@
...
-		super(context);
+	public MysqlSchemaStore(ConnectionPool maxwellConnectionPool,
+							ConnectionPool replicationConnectionPool,
+							Long serverID,
+							BinlogPosition initialPosition,
+							CaseSensitivity caseSensitivity,
+							MaxwellFilter filter,
+							boolean replayMode) {
+		super(replicationConnectionPool, caseSensitivity, filter);
+		this.caseSensitivity = caseSensitivity;
```
dadah89: not needed?

dadah89:
lgtm 👍

PR #412, opened 2016-08-25
RowMap: toJSON includes binlog position

dadah89:
/cc @zendesk/rules

### Description

By consequence messages sent by MaxwellKafkaProducer will include the binlog position.

I refactored json tests to support pattern matching, instead of changing the json files expectations to include the specific binlog position (locally positions had numbers like 18645, 20004, 21497, etc). MaxwellIntegrationTest tests were 100ms slower (they take more than 16s). My worry was making the tests too brittle (seems like binlog_position will change frequently).

### Risks
* Low.

### DIFF
--- a/src/test/java/com/zendesk/maxwell/MaxwellTestJSON.java
+++ b/src/test/java/com/zendesk/maxwell/MaxwellTestJSON.java
@@ -2,6 +2,7 @@ package com.zendesk.maxwell;
 
 import java.util.*;
 import java.io.*;
+import java.util.regex.Pattern;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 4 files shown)


src/test/java/com/zendesk/maxwell/MaxwellTestJSON.java
```diff
@@ -23,11 +24,43 @@
...
+
 	public static void assertJSON(List<Map<String, Object>> jsonOutput, List<Map<String, Object>> jsonAsserts) {
 		ArrayList<Map<String, Object>> missing = new ArrayList<>();
+		Boolean contains;
+
+		for ( Map<String, Object> m : jsonAsserts ) {
+			contains = false;
+
+			for ( Map jm : jsonOutput ) {
+				contains = true;
```
dadah89: I'm not checking if the generated output has more keys than the expected json. Lmk if it is a good idea to do it.

osheroff:
1.  should be off by default.  control it with a command line flag.
2.  needs to include the filename.   suggest that there's just one field that looks like `position: "master.00001:649"`

osheroff:
3.  Generally it's better to do stuff as one-off tests rather than refactor the whole suite; ie there's a set of integration tests that aren't JSON-based that assert various things about the system.

dadah89:
@osheroff making it optional would've avoided spending time refactoring tests =) (it made a lot of them fail b/c of the new field, and I couldn't just change them b/c of the binlog position value itself). Gonna do the change. Should I remove the refactor?

osheroff:
yeah I'd remove the refactor.

dadah89:
Replaced by #421.

PR #413, opened 2016-08-26
Bump kafka clients to 0.9

dadah89:
/cc @zendesk/rules

### Description

Locally, it seems to work just fine.

I checked the producer configuration, all params supported by [0.8](https://github.com/apache/kafka/blob/0.8.2/clients/src/main/java/org/apache/kafka/clients/producer/ProducerConfig.java) are present in [0.9](https://github.com/apache/kafka/blob/0.9.0/clients/src/main/java/org/apache/kafka/clients/producer/ProducerConfig.java). In the config file for 0.9 it [mentions `metadata.fetch.timeout.ms` is deprecated](https://github.com/apache/kafka/blob/8b65ec9caca82f98e715a0acb2ebabee3ae4fef1/clients/src/main/java/org/apache/kafka/clients/producer/ProducerConfig.java#L49) (in favor of `max.block.ms`), but it doesn't say deprecated [in the docs](http://kafka.apache.org/documentation.html#producerconfigs), and it's still present for 0.10, so didn't change it for now.

[Upgrade guide](http://kafka.apache.org/090/documentation.html#upgrade_9) didn't mention anything that would be of concern here.

### Risks
* High. Might stop producing kafka messages.


osheroff:
something like this:

http://grokbase.com/t/kafka/users/161dmqk9gj/kafka-0-9-client-producer-compatibility-with-kafka-0-8-2-broker

makes me worry about unforseen stuff.  

I think the right thing to do here is to package maxwell with both jars, and switch to the 0.9 jar based on a command line flag, ie `--kafka-0.9`

dadah89:
@osheroff I'll do it, will make it a bit easier to switch back. But that link is about a 0.8.2 broker, and our is 0.9 no?

osheroff:
yes, but maxwell is an open-source project.

dadah89:
uh ofc.

dadah89:
@osheroff my Makefile fu isn't too good, added some repetition =\.

From my current understanding of packaging: for package (compile task), it compiles against kafka-clients-0.8.2.2. For tests, I added a target for 0.9 and multiple travis builds, to try compiling with 0.8 and 0.9: if we bump to a kafka 0.9 that isn't compatible with 0.8, or we use something 0.9 specific in our classes, a travis build should fail.

I did all that so I could also better understand how packaging/compiling works in the Makefile =) Open to any suggestions about simplifying or changing the approach.

osheroff:
when we compile this we're putting *both* 0.8.2 and 0.9.0 on the classpath.  should be only one or the other

dadah89:
@osheroff I was relying on the order they were added, afaik the first one declared is used. Is that risky, should I not rely on that?

osheroff:
I'd rather not.

dadah89:
k, will fix it

dadah89:
@osheroff updated

.travis.yml
```diff
@@ -1,6 +1,11 @@
 language: java
-env:
-  - MYSQL_VERSION=5.5
-  - MYSQL_VERSION=5.6
+branches:
+  only: master
```
osheroff: why this?  would prefer that other branches be built

dadah89: just get rid of duplicated builds, PRs still build automatically

dadah89: to be more clear, gets rid of the /push build:
screen shot 2016-09-07 at 11 22 10 am

build/maven_fetcher
```diff
@@ -281,7 +291,9 @@ parser = OptionParser.new do |opts|
 
   opts.on("-r", "--repo REPO", "maven repository base url, default is '#{DEFAULT_REPO}'") { |repo| options[:repo] = repo }
   opts.on("-f", "--fetch ARTIFACT", "fetch a given artifact, named as 'GROUP/ARTIFACT/VERSION'") { |artifact| options[:fetch] = artifact }
+  opts.on("--skip-dependencies", "do not fetch artifact dependencies (only applies if -f is present)") { options[:skip_depedencies] = true }
   opts.on("-p", "--pom [POM]", "fetch all dependencies for pom.xml") { |pom| options[:pom] = pom }
+  opts.on("--force-artifact ARTIFACT", "when fetching from pom.xml, replace artifact with argument, named as 'GROUP/ARTIFACT/VERSION'") { |artifact| options[:force_artifact] = artifact }
```
osheroff: more elegantly expressed as:

```
--lock-version ARTIFACT=version
```

I think

osheroff:
is it possible to check the kafka server's version and crash if we don't match?

build/maven_fetcher
dadah89: name is hard =) will make the change

dadah89:
@osheroff can't easily check the broker's version or crash (exception gets caught in the kafka producer sender thread and not bubbled up), but added docs like we talked about, let me know if you have suggestions

docs/docs/kafka.md
```diff
@@ -47,6 +47,25 @@ bin/kafka-topics.sh --zookeeper ZK_HOST:2181 --create \
 [http://kafka.apache.org/documentation.html#quickstart](http://kafka.apache.org/documentation.html#quickstart)
 
 
+### Kafka 0.9
+***
+By default, maxwell runs with kafka clients 0.9.0.1, which isn't compatible withbrokers running kafka 0.8. The exception below will show in logs when that is the case:
```
osheroff: `withbrokers`

docs/docs/kafka.md
```diff
@@ -47,6 +47,25 @@ bin/kafka-topics.sh --zookeeper ZK_HOST:2181 --create \
...
+org.apache.kafka.common.protocol.types.SchemaException: Error reading field 'throttle_time_ms': java.nio.BufferUnderflowException
+    at org.apache.kafka.common.protocol.types.Schema.read(Schema.java:71) ~[kafka-clients-0.9.0.1.jar:?]
+    at org.apache.kafka.clients.NetworkClient.handleCompletedReceives(NetworkClient.java:439) ~[kafka-clients-0.9.0.1.jar:?]
+    at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:265) ~[kafka-clients-0.9.0.1.jar:?]
+    at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:216) ~[kafka-clients-0.9.0.1.jar:?]
+    at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:128) [kafka-clients-0.9.0.1.jar:?]
+    at java.lang.Thread.run(Thread.java:745) [?:1.7.0_79]
+```
+
+There is a flag (--kafka0.8) that switches maxwell to run with kafka clients 0.8.2.2.
```
osheroff: need to put this flag in "config.md"

osheroff:
👍 after minor doc fixes

docs/docs/kafka.md
dadah89: oh thought I had done that

PR #417, opened 2016-08-30
Adds support for mysql datetime(6) and timestamp(6) and time(6)

Dagnan:
Now supports `datetime(6)` and `timestamp(6)` and `time(6)`.

Also related to Exodus PR (link to come).


src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
```diff
@@ -25,24 +25,34 @@ private static SimpleDateFormat getDateTimeFormatter() {
...
-			else if ( value instanceof Timestamp )
-				return getDateTimeFormatter().format(( Timestamp ) value);
-			else if ( value instanceof Date )
+			} else if ( value instanceof Timestamp ) {
+				int nanos = (( Timestamp ) value).getNanos();
+				if ( nanos > 0 ) {
+					StringBuilder result = new StringBuilder();
+					result.append(getDateTimeFormatter().format(( Date ) value));
+					result.append(".");
+					result.append(String.format("%06d", nanos/1000));
```
osheroff: is it possible to capture the resolution that the user specified at table-creation time?  I can help out with capturing it from a `CREATE TABLE` statement, but wondering if it's possible to capture it from INFORMATION_SCHEMA

osheroff: not a blocker for merging, but curious.

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
```diff
@@ -25,24 +25,34 @@ private static SimpleDateFormat getDateTimeFormatter() {
...
-			if ( value instanceof Long && getType().equals("datetime") )
+			if ( value instanceof Long && getType().equals("datetime") ) {
 				return formatLong(( Long ) value);
-			else if ( value instanceof Timestamp )
-				return getDateTimeFormatter().format(( Timestamp ) value);
-			else if ( value instanceof Date )
+			} else if ( value instanceof Timestamp ) {
+				int nanos = (( Timestamp ) value).getNanos();
+				if ( nanos > 0 ) {
+					StringBuilder result = new StringBuilder();
```
osheroff: this is actually some of the most performance-sensitive code in the system.  maybe re-use a thread-local StringBuilder?

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
Dagnan: I did something to get the `columnLength` here, and removed it. Let me try again.

Dagnan:
@osheroff Fixed tests/bugs + added table resolution + a thread-local StringBuilder (my Java knowledge is very basic and very, very old so please say so if you see things to improve).

Dagnan:
Looks like there is still one bug to fix.

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -367,12 +367,23 @@ public void exitColumn_definition(mysqlParser.Column_definitionContext ctx) {
 		}
 
 		colType = ColumnDef.unalias_type(colType.toLowerCase(), longStringFlag, columnLength, byteFlagToStringColumn);
-		ColumnDef c = ColumnDef.build(name,
-					                   colCharset,
-					                   colType.toLowerCase(),
-					                   -1,
-					                   signed,
-					                   enumValues);
+    ColumnDef c;
```
osheroff: tabs vs spaces issues here.

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -367,12 +367,23 @@ public void exitColumn_definition(mysqlParser.Column_definitionContext ctx) {
...
+    if (columnLength != null) {
+      c = ColumnDef.build(name,
+  														 colCharset,
```
osheroff: whitespace issues here.  since maxwell uses tabs, you should write this as:

```
c = ColumnDef.build(
<tab>name, 
<tab>...
)
```

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDef.java
```diff
@@ -11,12 +11,21 @@
...
 		this.type = type;
 		this.pos = pos;
+		this.columnLength = 0L;
+	}
+
+	public ColumnDef(String name, String type, int pos, Long columnLength) {
+		this.name = name;
+		this.type = type;
+		this.pos = pos;
+		this.columnLength = columnLength;
```
osheroff: `columnLength` should be a property of `DateTimeColumnDef`, not others

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDef.java
```diff
@@ -26,6 +35,16 @@ public Object asJSON(Object value) {
 		return value;
 	}
 
+	public static ColumnDef build(String name, String charset, String type, int pos, boolean signed, String enumValues[], Long columnLength) {
```
osheroff: no need for a polymorphic `build` function -- just pass `null` for `columnLength` in the cases where you don't have it.

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
```diff
@@ -20,29 +39,71 @@ private static SimpleDateFormat getDateTimeFormatter() {
...
+		}
+
+		nanos = Integer.parseInt(realStrNanos.toString());
+		int micros = nanos / 1000;
+		String strMicros = Integer.toString(micros);
+
+    // 6 is the max precision of datetime2 in MysQL
+		int divideBy = ((int) Math.pow(10, 6 - columnLength));
+
+		return micros / divideBy;
```
osheroff: I'm incredibly confused about this function.  Can you explain what it's trying to do and why just keeping things as numbers doesn't work?

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -16,7 +16,7 @@
 import org.slf4j.LoggerFactory;
 
 public class MysqlParserListener extends mysqlBaseListener {
-    final Logger LOGGER = LoggerFactory.getLogger(MysqlParserListener.class);
+		final Logger LOGGER = LoggerFactory.getLogger(MysqlParserListener.class);
```
osheroff: you fixed whitespace, but fixed it wrong :)

osheroff:
General comments:

1. I'm super excited about this PR. 
2. You need to persist the column length in a variety of places - how `DateColumnDef` serializes and deserializes to JSON, as well as in the `maxwell.columns` table.  If you write a `DDLIntegrationTest` you will see the errors. 
3. You're capturing field length from the DDL statements, which is great, but you need to also capture it in the initial schemas.
4. Unfortunately, when you add information that wasn't captured in the initial schema you have to also back-fill it.  I can help here, I've written a bunch of this crap.

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
osheroff: oh I see you were following my lame-ass whitespace.  sorry, the above is the new whitespace method

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
Dagnan: Yeah, a find/replace all was not the best idea. I switch to Intellij, was using Atom.

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
Dagnan: I renamed it to `truncateNanosToColumnLength`. For example if you have as input `12345` and your column length is `3` you'll have `123`. `1` for a column length of `3` will give `0`.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -155,15 +155,16 @@ public Long saveDerivedSchema(Connection conn) throws SQLException {
...
+		return executeInsert(
+				insert,
+				this.baseSchemaID,
+				deltaString,
+				position.getFile(),
+				position.getOffset(),
+				serverID,
+				schema.getCharset(),
+				SchemaStoreVersion,
+				getPositionSHA());
```
osheroff: i think just one tab here as long as you're fixing it

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
```diff
@@ -109,6 +109,14 @@ private void captureTable(Table t) throws SQLException {
 			String colEnc     = r.getString("CHARACTER_SET_NAME");
 			int colPos        = r.getInt("ORDINAL_POSITION") - 1;
 			boolean colSigned = !r.getString("COLUMN_TYPE").matches(".* unsigned$");
+			Long columnLength;
+
+			// Only available for datetime/timestamp
+			try {
+				columnLength = r.getLong("DATETIME_PRECISION");
+			} catch ( java.sql.SQLException e ) {
+				columnLength = null;
```
osheroff: not the right way to check for null.  try:

```
   columnLength = ...
   if ( r.wasNull() ) 
      columnLength = null;
```

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDef.java
```diff
@@ -67,7 +67,7 @@ public static ColumnDef build(String name, String charset, String type, int pos,
 			return new DateColumnDef(name, type, pos);
 		case "datetime":
 		case "timestamp":
-			return new DateTimeColumnDef(name, type, pos);
+			return new DateTimeColumnDef(name, type, pos, columnLength);
 		case "year":
 			return new YearColumnDef(name, type, pos);
 		case "time":
```
osheroff: TIME columns can have precision too.  shall we hit that in this PR?

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -367,12 +367,25 @@ public void exitColumn_definition(mysqlParser.Column_definitionContext ctx) {
...
+				columnLength);
+		} else {
+			c = ColumnDef.build(
+				name,
+				colCharset,
+				colType.toLowerCase(),
+				-1,
+				signed,
+				enumValues, null);
+		}
```
osheroff: a single function call works here now.

src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
```diff
@@ -190,6 +190,30 @@ public void testYearWithLength() throws Exception {
 	}
 
 	@Test
+	public void testDatetimeWithLength() throws Exception {
+		if ( !server.getVersion().equals("5.6") )
+			return;
+
+		String sql[] = {
+			"create TABLE `test_datetime` ( id datetime(3) )"
```
osheroff: can throw in an ALTER and a TIMESTAMP here for good measure

osheroff:
wander off @Dagnan ?

Dagnan:
@osheroff  Haven't worked on it since last week, I need to work on more urgent things – should be able to resume that maybe end of next week.

Dagnan:
(a new step is to update `open-replicator` to be able to instanciate `DateTime2`/`Time2` with a `long` instead of a `Date` (as `Timestamp2` already does I think), otherwise we just lose the millisec precision.

osheroff:
@Dagnan, well, the microseconds get lost but it keeps millisecond precision at least.  See around  https://github.com/zendesk/open-replicator/blob/master/src/test/java/com/google/code/or/OpenReplicatorTest.java#L180  -- using `.getTime()` will retrieve the millisecond precision, which isn't lost.

Dagnan:
@osheroff will continue working on that on Monday.

I think there is value in supporting microsec precision. Especially to create a timestamp for rows that have the same `created_at` in Classic (one instance I've seen is 7k rows for the same timestamp).

Anyway I'd love to hear your thoughts on that.

osheroff:
ok, if/when you resume work here I'm happy to implement a `getSeconds()` and `getMicros()` method on all of the column types that will return the time uniformly (without mucking about with all these complex/dumb java types)

Dagnan:
@osheroff here is what I did so far, the idea being able to manipulate `Timestamp`s  from `open-replicator`: https://github.com/zendesk/open-replicator/pull/25

Dagnan:
Also, I intend to use it in `ColumnWithDefinition` in `valueForJSON`: get the `Timestamp` and then it's displayed as a `String` in `DatetimeColumnDef` and `TimeColumnDef`.

src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
Dagnan: Done

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
Dagnan: I think it was on an old commit right?

Dagnan:
cc @osheroff I think it's ready.

You mentioned

> Unfortunately, when you add information that wasn't captured in the initial schema you have to also back-fill it. I can help here, I've written a bunch of this crap.

How do we do that? Anything else to do?

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
```diff
@@ -0,0 +1,73 @@
...
+		int divideBy = ((int) Math.pow(10, 6 - columnLength));
+		return micros / divideBy;
+	}
+
+	protected static String objectWithPrecisionToString(String value, Timestamp t, Long columnLength) {
+		int fraction = truncateNanosToColumnLength(t.getNanos(), columnLength);
+		if (fraction == 0)
+			return value;
+
+		String strFormat = "%0" + columnLength + "d";
```
osheroff: not totally sure, but can't you just do: 

```
String strFormat = "%0" + columnLength + "d";
String fractional = String.format(strFormat, nanos).substring(0, columnLength)
```

Dagnan: You're right, but actually I may have to keep the division and round it to the upper integer. MySQL does that: http://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html

`.777` for `time(2)` gives `.78`

osheroff:
re: backfill, there's two things.

1.  you need to perform maintenance on the `maxwell`.`columns` table to store the width.  see `upgradeSchemaStoreSchema` for how to do this.
2.  any existing schemas will not have any length info, so you need to do a backfill on them.  see `handleVersionUpgrades` and `fixUnsignedColumns` for pointers on the right way to do this.  Ask me if that crap seems crazy.

Dagnan:
I think I got it. See last commit.

However:
* Not sure how the versions work in `handleVersionUpgrades`, I used `3`
* In my test `testUpgradeAddColumnLengthForExistingSchemas`, I'd like to make sure that `getColumnLength` returns `null`, after having sent the request `update maxwell.columns set column_length = NULL where name = 'badcol'`. Couldn't do that.
* Any idea how to fix the failing tests? :/

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
```diff
@@ -109,6 +109,14 @@ private void captureTable(Table t) throws SQLException {
 			String colEnc     = r.getString("CHARACTER_SET_NAME");
 			int colPos        = r.getInt("ORDINAL_POSITION") - 1;
 			boolean colSigned = !r.getString("COLUMN_TYPE").matches(".* unsigned$");
+			Long columnLength;
+
+			// Only available for datetime/timestamp/time
+			try {
+				columnLength = r.getLong("DATETIME_PRECISION");
+			} catch ( java.sql.SQLException e ) {
+				columnLength = null;
```
osheroff: oh btw, this isn't the right way to test for null. you need r.wasNull()

Dagnan:
Re: my question above, I mean something like:

```
		c.createStatement().executeUpdate("update maxwell.schemas set version = 2 where id = " + this.savedSchema.getSchemaID());
		c.createStatement().executeUpdate("update maxwell.columns set column_length = NULL where name = 'badcol'");

		SchemaStoreSchema.upgradeSchemaStoreSchema(c);
		DateTimeColumnDef cd1 = (DateTimeColumnDef) this.schema.findDatabase("shard_1").findTable("without_col_length").findColumn("badcol");
		assertEquals(0L, cd1.getColumnLength());

		MysqlSavedSchema restored = MysqlSavedSchema.restore(context, context.getInitialPosition());
		DateTimeColumnDef cd = (DateTimeColumnDef) restored.getSchema().findDatabase("shard_1").findTable("without_col_length").findColumn("badcol");

		assertEquals(3L, cd.getColumnLength());
```

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -569,8 +579,36 @@ private void fixColumnCases(Schema recaptured) throws SQLException {
...
+			}
+
+			if ( colLengthDiffs > 0 )
+				this.shouldSnapshotNextSchema = true;
+		}
+	}
+
 	protected void handleVersionUpgrades(Connection conn) throws SQLException, InvalidSchemaError {
-		if ( this.schemaVersion < 2 ) {
+		if ( this.schemaVersion < 3 ) {
```
osheroff: need to bump the SchemaStoreVersion constant to 3, so that newly created schemas are saved properly.

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
Dagnan: Ah I thought it was for an old version of my code, I'll fix.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
Dagnan: Ok that makes sense, I didn't push too far on that.

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
Dagnan: Actually `r` is not null, we have the row but without the column. That's why it does:
`java.sql.SQLException: Column 'DATETIME_PRECISION' not found.`

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
```diff
@@ -0,0 +1,84 @@
...
+	// - col length 3 -> 0 (milli secs)
+	// - col length 6 -> 123 (micro secs).
+	protected static int roundNanosToColumnLength(int nanos, Long columnLength) {
+		// 6 is the max precision of datetime2/time6/timestamp2 in MysQL
+		// 3: we go from nano (10^9) to micro (10^6)
+		int divideBy = (int) Math.pow(10, 6 + 3 - columnLength);
+		BigDecimal bigDivideBy = new BigDecimal(divideBy);
+		java.math.RoundingMode rounding = java.math.RoundingMode.HALF_UP;
+		// Scale of result is 0
+		BigDecimal result = new BigDecimal(nanos).divide(bigDivideBy, 0, rounding);
```
osheroff: sorry to keep harping on this code, but we don't actually have to worry about rounding; mysql is doing that for us:

```
mysql> create table dt ( d1 timestamp(1), d2 timestamp(2), d3 timestamp(3) );
mysql> insert into dt set d1 = '2016-09-29 01:50:10.494', d2 = '2016-09-29 01:50:10.494', d3 = '2016-09-29 01:50:10.494';
```

yields us values for `nanos` here pre-rounded:

500000000, 490000000, 494000000

so really, I think we can still do:
`String.format("%0" + columnLength + "d", t.getNanos()).substring(...)`

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
osheroff: Oh, is this if you're running under 5.5?

Dagnan: (fixed it)

osheroff: this has gotta be about running under 5.5, but it makes sense now.

osheroff:
can you hold off on the rebases for a bit?  I'm trying to run your code as it changes, hard to do with the constant force-push + rebase

Dagnan:
Sure.

osheroff:
re: `recoveryTest`; not sure.  I'll have to look at it, it's been somewhat flaky in the past but I think at the moment it's truly broken.

osheroff:
re: 

>In my test testUpgradeAddColumnLengthForExistingSchemas, I'd like to make sure that getColumnLength returns null, after having sent the request update maxwell.columns set column_length = NULL where name = 'badcol'. Couldn't do that.

Ok, I think I see what you're trying to do.  You could copy and deconstruct the code here:

https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java#L275

and inject some tests in between the `restoreFromSchemaID` and `handleVersionUpgrades` code.

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
Dagnan: It does not cover cases like this, but maybe we don't need because we can assume that we'll always serialize/deserialize from the same precision?

```
t.setNanos(123456789);
assertThat(d.toSQL(t), is("'19:19:19.123457'"));
```

osheroff: yup.  there should only be one precision for a row, and we should always be in sync with that.  If for some reason we accepted non mysql inputs, your code would be more needed, but I think we can assume that mysql will do our rounding for us, and the only thing we need to worry about is padding/truncating the string

Dagnan:
@osheroff this should be good for review. You should be able to see just the last commit. I added a public static method in `MysqlSavedSchema` to get a specific schema without upgrading it.
And the Recovery tests are still failing.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -437,7 +457,8 @@ private void restoreTable(Connection connection, Database d, String name, int id
 					cRS.getString("name"), cRS.getString("charset"),
 					cRS.getString("coltype"), i++,
 					cRS.getInt("is_signed") == 1,
-					enumValues);
+					enumValues,
+					Long.valueOf(cRS.getInt("column_length")));
```
osheroff: this will always make `column_length` `0`.. not sure if you  were trying to allow NULL here or if you intended that "zero is a good enough null"

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -136,6 +136,10 @@ public static void upgradeSchemaStoreSchema(Connection c) throws SQLException, I
 			performAlter(c, "alter table `positions` add column `last_heartbeat_read` bigint null default null");
 		}
 
+		if ( !getTableColumns("columns", c).containsKey("column_length") ) {
+			performAlter(c, "alter table `columns` add column `column_length` bigint unsigned");
```
osheroff: `tinyint` will be fine here; no matter that we'll be casting it to long in code

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	protected static String objectWithPrecisionToString(String value, Timestamp t, Long columnLength) {
+		// 6 is the max precision of datetime2/time6/timestamp2 in MysQL
+		// 3: we go from nano (10^9) to micro (10^6)
+		int divideBy = (int) Math.pow(10, 6 + 3 - columnLength);
+		int fractional = ((int) t.getNanos()) / divideBy;
+
+		if (fractional == 0)
+			return value;
```
osheroff: is that true?  I would think we'd actually want to zero-pad this.  so if some one does:

```
create table ( d timestamp(6) );  insert into .. set d = '2013-03-03.000000';
```

shouldn't we be producing `2013-03-03.000000`, not `'2013-03-03`?

src/main/java/com/zendesk/maxwell/schema/columndef/TimeColumnDef.java
```diff
@@ -1,27 +1,29 @@
...
 	}
 
-	@Override
-	public String toSQL(Object value) {
-		Time t = (Time) value;
-		return "'" + String.valueOf(t) + "'";
-	}
+	protected String formatValue(Object value) {
+		if ( value instanceof Timestamp ) {
+			Time time = new java.sql.Time(((Timestamp) value).getTime());
```
osheroff: no need to say `new java.sql.Time` -- just `new Time` will work since you've imported it

src/main/resources/sql/maxwell_schema.sql
```diff
@@ -40,6 +40,7 @@ CREATE TABLE IF NOT EXISTS `columns` (
   coltype     varchar(255),
   is_signed   tinyint(1) unsigned,
   enum_values text,
+  column_length bigint unsigned,
```
osheroff: `tinyint`

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
Dagnan: Ok, I was gonna ask. It's weird that in the code we deal with a `long` though.

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
Dagnan: I assumed it wouldn't change anything, but you're right, I'll make it consistent.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
Dagnan: The constructor does something like this, but I updated this part to send null it the value read was null, so then the logic is only inside the constructor.

osheroff:
@Dagnan I fixed the broken tests with https://github.com/zendesk/maxwell/pull/417/commits/abc3fa35691c15292d4d00211a0c1dfa14b7e17c

I think we're good here, will do one last pass in the AM and merge

Dagnan:
Ah, thanks, good catch!

osheroff:
👍

Issue #418, opened 2016-08-31
DDL statement parsing error "MaxwellSQLSyntaxError: InnoDB"


powerLambda:
Full error stack as below:

3:50:50,710 INFO AbstractSchemaStore - storing schema @BinlogPosition[mysql-bin.003981:29180028] after applying "DROP TABLE IF EXISTS xxx_asset_access_audit /* generated by server */"
23:50:50,777 ERROR MysqlParserListener - (creation_engine ENGINE = InnoDB)
23:50:50,777 ERROR SchemaChange - Error parsing SQL: 'CREATE TABLE xxx_asset_access_audit (
id int(11) NOT NULL,
partner_id varchar(31) NOT NULL ,
partner_product_id varchar(31) NOT NULL ,
log_date datetime NOT NULL ,
batch_id varchar(64) NOT NULL ,
order_id varchar(64) NOT NULL ,
audit_status tinyint(4) NOT NULL ,
audit_reason varchar(255) DEFAULT NULL ,
remark varchar(255) DEFAULT ',
is_deleted char(1) NOT NULL DEFAULT 'N' ,
gmt_created datetime NOT NULL ,
creator varchar(32) NOT NULL ,
gmt_modified datetime NOT NULL ,
modifier varchar(32) NOT NULL ,
PRIMARY KEY (id),
UNIQUE UN_ORDER (partner_product_id, batch_id, order_id) comment ''
) ENGINE=InnoDB DEFAULT CHARACTER SET utf8 COLLATE utf8_general_ci ROW_FORMAT=COMPACT CHECKSUM=0 DELAY_KEY_WRITE=0'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: InnoDB
at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:84)
at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
[...]

PR #419, opened 2016-08-31
fix for back-quoted ENGINE=

osheroff:
addresses #418 

@zendesk/rules


dadah89:
👍 
how did you figure out it was back-quotes from #418's output?

osheroff:
there was a PR he opened up that had the SQL properly pasted.

PR #420, opened 2016-08-31
Master recovery

osheroff:
There's a couple of rough edges still, but it's certainly close enough for a CR (and maybe a deploy to an integration env).  

Apologies for the length.

@zendesk/rules


src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
```diff
@@ -352,6 +385,14 @@ public RowMap getRow() throws Exception {
 						setReplicatorPosition((AbstractBinlogEventV4) v4Event);
 					}
 					break;
+				case MySQLConstants.ROTATE_EVENT:
```
dadah89: is there a use case to stop on log file rotation? my understanding was that mysql would send eof when catching up to master (no more events to process) https://github.com/zendesk/open-replicator/pull/22, not necessarily on log rotation

src/main/java/com/zendesk/maxwell/RunLoopProcess.java
```diff
@@ -31,8 +34,8 @@ public boolean runLoop() throws Exception {
 			while (this.runState == RunState.RUNNING)
 				work();
 
-			this.beforeStop();
 		} finally {
+			this.beforeStop();
```
dadah89: just curiosity, not too sure on finally behavior =), if `beforeStop` throws `runState` won't be changed? (was going to suggest inverting the statements order, but afaik won't matter much, and the program will just finish main execution)

src/main/java/com/zendesk/maxwell/MaxwellReplicator.java
osheroff: yeah, I'm mixing terms a *little* bit.  open-replicator's eof-mode stops at the end of all the data, whereas maxwell's stops after each file. 

I'm stopping after each file so that I can scan backwards file by file to find the heartbeat.  I start with the most recent binlog and walk backwards from there.

osheroff:
TODO: write some tests when you *don't* find the target heartbeat

src/main/java/com/zendesk/maxwell/RunLoopProcess.java
osheroff: yeah that's totally true, but it's less harmful than current behavior. 

 I did this to make sure that if `work()` throws we still call `beforeStop` -- we weren't shutting down open-replicator and were leaking threads.

If ultra-paranoid could do:

```
try {
   ...
} finally { 
   try { 
   } finally { 
      this.runState = RunState.STOPPED
```

src/main/java/com/zendesk/maxwell/recovery/Recovery.java
```diff
@@ -0,0 +1,111 @@
...
+		return null;
+	}
+
+	/**
+	 * fetch a list of binlog postiions representing the start of each binlog file
+	 *
+	 * @return a list of binlog positions to attempt recovery at
+	 * */
+
+	private List<BinlogPosition> getBinlogInfo() throws SQLException {
```
dadah89: would it work rejecting everything that is older than the last known position file?

dadah89: hm I guess you'd need to find the position associated to the last heartbeat read, and not the one we're using here (associated to last heartbeat sent) from what I understood

osheroff: yeah.  just remember.  new master's position has no correlation to old master's position.  we pick up the old master's position so we can borrow its schema, but that's about it...

dadah89:
gave it a first pass, approach seems good to me

src/main/java/com/zendesk/maxwell/recovery/Recovery.java
dadah89: so could we break the `findHeartbeat` loop once we find a heartbeat older than last_heartbeat_read? just worried that I have more than 10 files locally, but maybe it doesn't take that long to parse binlog files and it isn't important?

osheroff: we can't break the `findHeartbeat` loop: we can't actually scan *backwards* in the logs, so we scan forwards, one file at a time -- meaning if we found an old heartbeat, ok, the one we're looking for might be forward in the file.  but we *can* break the `recover()` loop if `findHeartbeat` scans the whole file and finds only an old binlog position.

osheroff:
gonna close this one and re-open a PR against master

PR #424, opened 2016-09-08
master recovery tweaks

osheroff:
- document command line flag, add secontion to docs
- only send heartbeat if position advanced substantially since last
  heartbeat.  Prevents us from endlessly moving the binlog position
  forward.

@zendesk/rules


dadah89:
lgtm 👍

PR #425, opened 2016-09-08
Master recovery

osheroff:
@zendesk/rules


src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -25,7 +26,8 @@
 	private final String database;
 	private final String table;
 	private final Long timestamp;
-	private final BinlogPosition nextPosition;
+	private BinlogPosition nextPosition;
+	private boolean isHeartbeat;
```
dadah89: you don't need this, do you?

dadah89:
lgtm 👍

src/main/java/com/zendesk/maxwell/RowMap.java
osheroff: will remove

Issue #429, opened 2016-09-14
Alter with 'engine' + 'character set' crushes maxwell


imdark:
`"ALTER TABLE `test` ENGINE=`InnoDB` CHARACTER SET latin1"`

this alter throws a
`java.lang.ClassCastException: com.zendesk.maxwell.schema.ddl.TableAlter cannot be cast to com.zendesk.maxwell.schema.ddl.TableCreate 
`

the source of the issue is MysqlParserListener.java: the first line of this function
```
public void exitCreation_character_set(Creation_character_setContext ctx) {
                TableCreate tableCreate = (TableCreate) schemaChanges.get(0);
                tableCreate.charset = unquote_literal(ctx.charset_name().getText());
}

```
character set should work for both TableCreate and TableUpdate

osheroff:
fixed in 1.2.1

PR #430, opened 2016-09-14
Fix odd ALTER TABLE character set issue

osheroff:
this also fixes an issue in which we weren't properly setting up the default character set on a table on an ongoing basis (oops.)

@zendesk/rules 

addresses #429

### DIFF
--- a/src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
+++ b/src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
@@ -41,7 +41,7 @@ public class DDLIntegrationTest extends MaxwellTestWithIsolatedServer {
 
 			"create table shard_1.testDrop ( id int(11) )",
 			"drop table shard_1.testDrop",
-			"create table test.c ( v varchar(255) charset ascii )",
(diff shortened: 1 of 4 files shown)


src/test/java/com/zendesk/maxwell/schema/ddl/DDLIntegrationTest.java
```diff
@@ -40,7 +41,7 @@ public void testAlter() throws Exception {
 
 			"create table shard_1.testDrop ( id int(11) )",
 			"drop table shard_1.testDrop",
-			"create table test.c ( v varchar(255) charset ascii )"
+			"create table test.c ( v varchar(255) charset ascii )",
```
dadah89: extra comma

osheroff: i should put a source-formatter in maxwell like we do with hanlon.

osheroff: well, that search was a terrible waste of 15 minutes.

dadah89:
lgtm 👍

PR #435, opened 2016-09-16
separate out heartbeats from different client ids

osheroff:
this was messing up both the heartbeating code (two client_ids would ping-pong each other), and also maxwell would "recover" the position from its sister client_id

@zendesk/rules


dadah89:
lgtm 👍  on green

PR #438, opened 2016-09-21
output DDLs New

xmlking:
This is based on Maxwell 1.3.x with minimal code changes to keep in sync with master branch. 
Includes 
1. Output DDL


osheroff:
not bad, sumanth.  we'd definitely need an enable/disable flag here, but other than that it could be mergable.  I still am squinting at how the DDL is partitioned differently from everything else; I would if it would make sense to have a tri-state option, `output_ddl=[false|true|only]` -- that way folks who wanted to do something a bit saner could put the ddl changes into a separate (presumably 1-topic partition).

xmlking:
so far I am using single partition topic, so didn't realise partition issue. we can push to any random  partition if others are using multi-partition topic. logically they might have same consumer code for all partitions to process DDLs. in my case, DDL order is not that important. would you like the optional kafka topic flag name as `ddl_kafka_topic` . if user didn't supply `ddl_kafka_topic`  and if `output_ddl=true`  we can produce DDL output to `kafka_topic` topic. 
let me know if this make sense. 

what do you mean `only` in `output_ddl=[false|true|only]` ?  what default value you prefer for `output_ddl`?

osheroff:
maybe we could partition by database instead of random?  That would provide some kind of logical consistency, although 1 partition should be the encouraged setup. 

default for `output_ddl` should be `false`.

osheroff:
re: `only` -- I guess my original concept was that in a very proper setup, you'd run two copies of maxwell -- one would output ddl, the other would output data, and each would have its own partitioning settings and such, rather than the DDL being sort of a special-case. 

But that's probably overblown, and solved by:
- allowing for a separate DDL topic, and
- partitioning DDL by database.

xmlking:
@osheroff please validate and see if this can be merged.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -226,6 +231,8 @@ private void setup(OptionSet options, Properties properties) {
 		this.kafkaKeyFormat     = fetchOption("kafka_key_format", options, properties, "hash");
 		this.kafkaPartitionKey  = fetchOption("kafka_partition_by", options, properties, "database");
 		this.kafkaPartitionHash = fetchOption("kafka_partition_hash", options, properties, "default");
+		this.outputDDL 			= fetchBooleanOption("output_ddl", options, properties, false);
```
osheroff: should be part of `outputConfig`

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -105,6 +112,10 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
 
 	@Override
 	public void push(RowMap r) throws Exception {
+		if (r instanceof DDLMap && !outputDDL) {
+			return;
+		}
```
osheroff: the loose contract we put in place for HeartbeatRowMap should be used here:  if a row shouldn't be output, it should return `null` from toJSON; this will make all the producers honor the `outputDDL` flag rather (implicitly) rather than having to modify each producer.

more succintly: should put `outputDDL` into `MaxwellOutputConfig`, and have the `DDLRowMap` return `null` from `toJSON` if we shouldn't output it.

(that should fix your failing tests, too)

osheroff:
would also like to see an integration test in which we turn this feature on and test its output.

xmlking:
@osheroff please review the changes I applied based on your feedback.

osheroff:
code looks pretty good.  I'll test this out sometime this week and plan on merging.  Can I convince you to document your feature?

osheroff:
👍

Issue #437, opened 2016-09-20
UnsupportedOperationException error for ALTER TABLE


xmlking:
maxwell  1.2.0 stopped with following error:

```
05:29:21,051 INFO  AbstractSchemaStore - storing schema @BinlogPosition[binarylog.051041:409877] after applying "ALTER TABLE mytble  CHANGE abcSeqNum abcNum  INT (5)"
05:29:21,072 INFO  AbstractSchemaStore - storing schema @BinlogPosition[binarylog.051041:412553] after applying "ALTER TABLE mytble  CHANGE abcNum abcAdjNum  INT (5)"
java.lang.UnsupportedOperationException
  at java.util.AbstractList.remove(AbstractList.java:161)
  at com.zendesk.maxwell.schema.Table.removePKColumn(Table.java:299)
  at com.zendesk.maxwell.schema.Table.removeColumn(Table.java:261)
  at com.zendesk.maxwell.schema.ddl.ChangeColumnMod.apply(ChangeColumnMod.java:19)
  at com.zendesk.maxwell.schema.ddl.TableAlter.resolve(TableAlter.java:52)
  at com.zendesk.maxwell.schema.ddl.TableAlter.resolve(TableAlter.java:12)
  at com.zendesk.maxwell.schema.AbstractSchemaStore.resolveSQL(AbstractSchemaStore.java:57)
  at com.zendesk.maxwell.schema.MysqlSchemaStore.processSQL(MysqlSchemaStore.java:82)
  at com.zendesk.maxwell.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:415)
  at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:384)
  at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:146)
  at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:35)
  at com.zendesk.maxwell.Maxwell.start(Maxwell.java:163)
  at com.zendesk.maxwell.Maxwell.main(Maxwell.java:180)
```

osheroff:
ok, finally got to where i can repro this.  pk + change-columns is all fu'ed

osheroff:
fixed in 1.2.2

PR #440, opened 2016-09-22
fixes for issue #437

osheroff:
1 - copy pkList when cloning table; otherwise we end up mutating the old
one.

2 - deal with the fact that CHANGE COLUMN will keep something in the PK

@zendesk/rules #437


dadah89:
lgtm 👍

PR #446, opened 2016-09-28
Add an option to include the server and thread ID in the output

sagiba:
Added an option (--output_thread_info) to include the MySQL server_id and thread_id in the output. We use this for joining maxwell event data against thread data stored elsewhere to add application context to every binlog event.


osheroff:
great, thanks for the PR.  Code looks good, couple of mods desired:

1 - can you separate out server_id and thread_id into two flags?  Seems like they'd be independently useful.

2 - plz write a little integration test exercising your code paths.

sagiba:
Done, separated the config into two flags (output_server_id and output_thread_id), and added integration tests.

osheroff:
very nice.  👍

Issue #427, opened 2016-09-13
partition by column?


smferguson:
We're going to use maxwell as CDC in a (mysql backed) event sourced system. In this case the primary key has no meaning (tables are append-only). We would like to partition by a non-distinct 'event_id' column instead of the primary key. Is there any ability to partition by a random column? (or are we crazy for a number of reasons?)

osheroff:
i don't think it's crazy.  I actually kinda have a use case for partitioning by a column, but I could never quite figure out what to do if/when the column didn't exist in a table; fall back to PK/random?  Would take a patch around this if you wrote one

xmlking:
partition by a **callback function** that a developer can supply as part of deployment configuration would be nice to accommodate specific use cases:
```java
String getPartitionKey( String database,  String table, Sting pk, Schema schema) {
  // your logic   
 return  composedPartitionKey
} 
```

smferguson:
@xmlking neat idea, but beyond the scope of work for me at the moment (and @osheroff hasn't commented on it). please start a new discussion if you want to pursue that.

PR #450, opened 2016-10-07
Issue #427: add partitioning by column

smferguson:
Add partitioning by one or multiple columns. If the column doesn't exist fall back to partitioning by table. I chose table as the fallback because falling back to primary key seemed like it could be dangerous in certain cases (e.g. using mysql as an immutable store - pk has no meaning). I'm not completely sold on this fallback.
The partitioner holds the columns we will try to partition by. I thought about populating the partition columns to the RowMap, but it didn't feel right for the RowMap to be aware partitioning details.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/RowMap.java
@@ -157,7 +157,7 @@ public class RowMap implements Serializable {
 		return keys;
 	}
 
-	public String getPartitionKey(List<String> partitionColumns) {
+	public String buildPartitionKey(List<String> partitionColumns) {
 		if (partitionColumns.isEmpty()) {
 			return table;
 		}
--- a/src/main/java/com/zendesk/maxwell/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/RowMap.java
@@ -157,11 +157,7 @@ public class RowMap implements Serializable {
 		return keys;
 	}
 
-	public String buildPartitionKey(List<String> partitionColumns) {
-		if (partitionColumns.isEmpty()) {
-			return table;
-		}
-
+	public String buildPartitionKey(List<String> partitionColumns, String partitionKeyFallback) {
 		String partitionKey="";
 		for (String pc : partitionColumns) {
 			Object pcValue = null;
@@ -171,8 +167,20 @@ public class RowMap implements Serializable {
 				partitionKey += pcValue.toString();
 		}
 		if (partitionKey.isEmpty())
-			return table;
+			return getPartitionKeyFallback(partitionKeyFallback);
 		return partitionKey;
+	}
+
+	private String getPartitionKeyFallback(String partitionKeyFallback) {
+		switch (partitionKeyFallback) {
+			case "table":
+				return this.table;
+			case "primary_key":
+				return pkAsConcatString();
+			case "database":
+			default:
+				return this.database;
+		}
 	}
 
 	private void writeMapToJSON(String jsonMapName, LinkedHashMap<String, Object> data, boolean includeNullField) throws IOException {
--- a/src/main/java/com/zendesk/maxwell/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/RowMap.java
@@ -161,7 +161,7 @@ public class RowMap implements Serializable {
 		if (partitionColumns.isEmpty()) {
 			return table;
 		}
-		// rename keys
+
 		String partitionKey="";
 		for (String pc : partitionColumns) {
 			Object pcValue = null;
--- a/src/main/java/com/zendesk/maxwell/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/RowMap.java
@@ -157,6 +157,24 @@ public class RowMap implements Serializable {
 		return keys;
 	}
 
+	public String getPartitionKey(List<String> partitionColumns) {
+		if (partitionColumns.isEmpty()) {
+			return table;
+		}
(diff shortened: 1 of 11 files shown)


src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java
```diff
@@ -45,7 +52,14 @@ static protected String getPrimaryKey(RowMap r) {
...
+	public void setPartitionColumns(String csvColumns) {
+		this.partitionColumns = Arrays.asList(csvColumns.split(","));
+	}
+
 	public int kafkaPartition(RowMap r, int numPartitions) {
-		return Math.abs(hashFunc.hashCode(provider.getHashString(r)) % numPartitions);
+		if ( this.provider instanceof HashStringColumn )
+			return Math.abs(hashFunc.hashCode(((HashStringColumn) provider).getHashString(r, this.partitionColumns)) % numPartitions);
+		else
+			return Math.abs(hashFunc.hashCode(provider.getHashString(r)) % numPartitions);
```
osheroff: this code could be cleaner; I think it would be preferable to put `partitionColumns` in the initializers for both `MaxwellKafkaPartitioner` and `HashStringColumn`

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -157,6 +157,24 @@ public String pkAsConcatString() {
...
+			Object pcValue = null;
+			if (data.containsKey(pc))
+				pcValue = data.get(pc);
+			if (pcValue != null)
+				partitionKey += pcValue.toString();
+		}
+		if (partitionKey.isEmpty())
+			return table;
+		return partitionKey;
+	}
```
osheroff: should add some documentation somewhere around what the key will be (how various columns end up formatted, etc) -- I've found it's often necessary to re-code partition-key methods in other languages.  I was debating whether or not to separate the values with `:` -- I don't think there's much value there.

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java
smferguson: yeah. i wasn't a fan of this either, but was hesitant to push it into the initializers. thanks. will take care of it.

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java
```diff
@@ -4,14 +4,21 @@
...
  */
 public class MaxwellKafkaPartitioner {
 	HashStringProvider provider;
 	HashFunction hashFunc;
-	public MaxwellKafkaPartitioner(String hashFunction, String partitionKey) {
+	List<String> partitionColumns = new ArrayList<String>();
+
+	private String partitionKeyFallback;
+
+	public MaxwellKafkaPartitioner(String hashFunction, String partitionKey, String csvColumns, String partitionKeyFallback) {
```
osheroff: nitpick: `csvColumns -> partitionKeyColumns`

osheroff: ah, you named it `partitionColumns` in the interface.  that's fine too.

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java
```diff
@@ -26,11 +33,17 @@ public MaxwellKafkaPartitioner(String hashFunction, String partitionKey) {
...
 				break;
+			case "column": this.provider = new HashStringColumn();
+				break;
 			case "database":
 			default:
 				this.provider = new HashStringDatabase();
 				break;
 		}
+		if ( partitionColumns != null )
+			this.partitionColumns = Arrays.asList(csvColumns.split(","));
```
osheroff: `partitionColumns` will never be null -- you're doing the null check on the wrong variable

osheroff: could also account for whitespace, I think; either use `\\s*,\\s*` as the split regex or call `.trim` on each

src/main/java/com/zendesk/maxwell/RowMap.java
```diff
@@ -157,6 +157,32 @@ public String pkAsConcatString() {
...
+				pcValue = data.get(pc);
+			if (pcValue != null)
+				partitionKey += pcValue.toString();
+		}
+		if (partitionKey.isEmpty())
+			return getPartitionKeyFallback(partitionKeyFallback);
+		return partitionKey;
+	}
+
+	private String getPartitionKeyFallback(String partitionKeyFallback) {
```
smferguson: @osheroff given this new config option should the pk fallback be configurable as well?

osheroff: mmmmaybe?  It looks like we're doing wildly different things for two PK fallbacks -- in two cases we slam database/table in there, but in the most common case we put a GUID in.  I'm not sure enough how people use partition-by-PK to really know.

smferguson: thanks. I'll leave it as-is. the current code does seem a bit odd.

osheroff:
tested locally, looks good.  👍

PR #453, opened 2016-10-10
refactor: create replication & row package.  move around others.

osheroff:
My apologies to any open PRs that will conflict with this one.  The root package was growing way too big. 

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/BufferedMaxwell.java
+++ b/src/main/java/com/zendesk/maxwell/BufferedMaxwell.java
@@ -1,6 +1,7 @@
 package com.zendesk.maxwell;
 
 import com.zendesk.maxwell.producer.BufferedProducer;
+import com.zendesk.maxwell.row.RowMap;
 
 import java.io.IOException;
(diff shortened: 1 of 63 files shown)


src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefTest.java
```diff
@@ -0,0 +1,358 @@
+package com.zendesk.maxwell.schema.columndef;
```
dadah89: could've git mv the file

dadah89:
👍  after fixing conflicts and green build

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefTest.java
osheroff: `git mv` is just a delete and an add -- other than that there's just an algorithm to detect renames.  If you change too much of the file, you confuse the rename algorithm.  A pity, really.

Issue #441, opened 2016-09-22
Rework shutdown process in case of exception


bmichaud:
Installed 1.2.1 version yesterday, re-created the Maxwell schema and started up Maxwell. It failed with the below error about 9:30 hours later. 

There is another client for the same source database with a different client ID that did not fail. That node is a fail-over, so does not receive as much data in the binlogs.

Here is the error:
===========
22:29:24,265 WARN  MaxwellReplicator - no heartbeat heard from server in 2002ms.  restarting replication.
java.net.ConnectException: Connection timed out
        at java.net.PlainSocketImpl.socketConnect(Native Method)
        at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
        at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
        at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
        at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
        at java.net.Socket.connect(Socket.java:589)
        at java.net.Socket.connect(Socket.java:538)
        at java.net.Socket.<init>(Socket.java:434)
        at java.net.Socket.<init>(Socket.java:211)
        at com.google.code.or.io.impl.SocketFactoryImpl.create(SocketFactoryImpl.java:37)
        at com.google.code.or.net.impl.TransportImpl.connect(TransportImpl.java:71)
        at com.google.code.or.OpenReplicator.start(OpenReplicator.java:106)
        at com.zendesk.maxwell.MaxwellReplicator.ensureReplicatorThread(MaxwellReplicator.java:127)
        at com.zendesk.maxwell.MaxwellReplicator.getRow(MaxwellReplicator.java:369)
        at com.zendesk.maxwell.MaxwellReplicator.work(MaxwellReplicator.java:153)
        at com.zendesk.maxwell.RunLoopProcess.runLoop(RunLoopProcess.java:35)
        at com.zendesk.maxwell.Maxwell.start(Maxwell.java:163)
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:180)
12:56:00,757 INFO  PositionStoreThread - Storing final position: BinlogPosition[binarylog.051165:86837]
12:56:00,768 ERROR PositionStoreThread - Hit com.zendesk.maxwell.errors.DuplicateProcessException exception in MysqlPositionStore thread.
[...]

PR #454, opened 2016-10-11
Rework shutdown

osheroff:
1 - make sure we shutdown maxwell before we shutdown log4j.  I think this was the source of some deadlocks.
2 - add a connectTimeout on the jdbc connection so we don't spin endlessly.

@zendesk/rules 

https://github.com/zendesk/maxwell/issues/441

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
@@ -16,6 +16,7 @@ public class MaxwellMysqlConfig {
 	public String user;
 	public String password;
 	public ArrayList<String> jdbcOptions = new ArrayList<String>();
+	public Integer connectTimeoutMS = 5000;
 
 	public MaxwellMysqlConfig() {
 		this.host = null;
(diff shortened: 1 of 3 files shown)


dadah89:
b/c the error came from socketConnect I thought it'd need to socket timeout, but afaict if that was the case then the exception would be `SocketTimeoutException`, so I guess it's fine

src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
```diff
@@ -16,13 +16,18 @@
 	public String user;
 	public String password;
 	public ArrayList<String> jdbcOptions = new ArrayList<String>();
+	public Integer connectTimeoutMS = 5000;
```
dadah89: `final`, and all caps if you care about java const name conventions

dadah89:
just a comment about `final`, besides looks good to me 👍

src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
osheroff: nah, figured it'd be loosely configurable, at least in code

PR #457, opened 2016-10-19
Add kafka topic per partition. Based largely on zendesk/maxwell#407. …

smferguson:
…Credit goes to @sschatts.


src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
```diff
@@ -283,18 +283,18 @@ else if ( tokenText.equals(")") )
...
+		if ( ctx.MAXWELL_ELIDED_PARSE_ISSUE() == null )
+			throw new ReparseSQLException(spliceParens(ctx.getStart().getTokenIndex(), 0));
+
+	}
+
+	@Override
+	public void enterSkip_parens_inside_partition_definitions(Skip_parens_inside_partition_definitionsContext ctx) {
+		if ( ctx.MAXWELL_ELIDED_PARSE_ISSUE() == null )
+			throw new ReparseSQLException(spliceParens(ctx.getStart().getTokenIndex(), 1));
+	}
```
smferguson: @osheroff something I keep forgetting to ask about... this code blows up on me with: 
> Error:(287, 38) java: cannot find symbol
>   symbol:   class Skip_parensContext
>   location: class com.zendesk.maxwell.schema.ddl.MysqlParserListener
I'm betting it's something I have configured wrong, I can (should) dig into it a bit, but any chance you know off the top of your head?

osheroff: @smferguson sounds like oddness with your antlr config.  I think maven knows how to do this with the plugin, the Makefile definitely has code to rebuild this... how are you building code?

smferguson: @osheroff make via intellij. I found that it doesn't actually pickup changes in the filters unless I use make on the cmdline so it's gotta be an IDE config.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -129,9 +142,11 @@ public void push(RowMap r) throws Exception {
 
 		ProducerRecord<String, String> record;
 		if (r instanceof DDLMap) {
-			record = new ProducerRecord<>(ddlTopic, this.ddlPartitioner.kafkaPartition(r, this.numPartitions), key, value);
+			String topic = generateTopic(this.ddlTopic, r);
+			record = new ProducerRecord<>(topic, this.ddlPartitioner.kafkaPartition(r, getNumPartitions(topic)), key, value);
```
osheroff: if we're gonna support this feature for the DDL topic (and I'm honestly not sure we should), we need to rework `DDLMap#getDatabase()` and `DDLMap#getTable()` -- at the moment it looks as though those are currently just hard-coded to some defaults, rather than fished out of the `ResolvedSchemaChange` object

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -110,9 +111,21 @@ public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties,
 		this.inflightMessages = new InflightMessageList();
 	}
 
+	private Integer getNumPartitions(String topic) {
+		try {
+			return this.kafka.partitionsFor(topic).size(); //returns 1 for new topics
+		} catch(TimeoutException e) {
+			LOGGER.error("Topic '" + topic + "' name does not exist. Exception: " + e.getLocalizedMessage());
+			throw e;
```
osheroff: wondering how thoroughly we've tested this; does the 0.9 producer do anything smarter when given a non-existent topic?  Or are we really stuck guessing at timeouts?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
smferguson: I'll back this out, just saw the update and thought it would be more complete, but yeah. the hardcoding doesn't play well.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
smferguson: 0.8 doesn't specify its behavior, https://kafka.apache.org/082/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#partitionsFor(java.lang.String)
0.9 and 0.10 have the same behavior so we have a few more exceptions we should catch here,  https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#partitionsFor(java.lang.String)
> Throws:
> WakeupException - if wakeup() is called before or while this function is called
> AuthorizationException - if not authorized to the specified topic
> TimeoutException - if the topic metadata could not be fetched before expiration of the configured request timeout
> KafkaException - for any other unrecoverable errors

osheroff: looked into this; it's basically sucky.  for both 0.8 and 0.9 producers, with auto-creation disabled, an attempt to call `partitionsFor()` on a non-existent topic just results in a `TimeoutException` the same way that Sarah coded for.  In 0.9 you get 

```
13:16:41,371 WARN  NetworkClient - Error while fetching metadata with correlation id 563 : {foijlkj=UNKNOWN_TOPIC_OR_PARTITION}
```

logged, but nothing programattic. 

There's also the `listTopics()` method on the consumer; maybe that would be a more elegant way to do this.

Really what I'd like is for maxwell to crash cleanly if you attempt to produce to a non-existent topic.  The `TimeoutException` method is... sorta viable... but I'll ask on the kafka mailing list.

smferguson: yeah. it's not ideal. I'd love a cached 'check for existence' call, but those seem to use zookeeper, and are unrelated to the producer/consumer side of things.
see https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/admin/AdminUtils.scala#L386
we could go with the consumer.listTopics() call and cache the result (the call makes a roundtrip to the server), but it seems odd having a consumer within the producer.
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1326
I hope the list has a good idea.

osheroff: blows the shit out of the scope, but maybe there's an opportunity here; what about a smallish `kafka-metadata` jar?  Something where you could at first answer queries like: 

- what version of kafka am I on? 
- what topics exist?  

and then, maybe eventually trying to turn into the missing kafka-admin library (create/modify/delete topics programmatically, etc).

Lots of duplication/overlap with what the admin utils already do, but it might be worth it to see how easy they'd be to extract out into a JAR....

or am I a DIY nut?  unclear.

smferguson: hmm... i'm more curious why KafkaProducer.java doesn't have a listTopics() function, but KafkaConsumer.java does. it seems like a better solution to add the missing functionality there. KafkaProducer has a cluster object which it uses for the partitionsFor() call. this also has a topics() call. it seems like the producer could use that call to return a list of topics.
the big downside is that it would be proposing a change to kafka which would be a long poll, but, i think, much cleaner in the long run.
code refs:
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1326
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L629
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L755
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/Cluster.java#L238

src/main/java/com/zendesk/maxwell/schema/ddl/MysqlParserListener.java
smferguson: @osheroff fyi, after running make at the commandline intellj is playing nice. odd. pretty certain i'd done that before.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
osheroff: yeah, saw that.  I also noted in a separate thread that even a `listTopics` doesn't really properly deal with the subject of auto-creation.  Regardless, it doesn't look like there's an easy win so maybe we hack around it?  I'd say we should lower the default metadata timeout to like 3 seconds so that it doesn't take us 60 seconds to crash, and then just keep the crash-on-TimeoutException lameness for now.

config.properties.example
```diff
@@ -76,9 +76,13 @@ password=maxwell
 #kafka.bootstrap.servers=hosta:9092,hostb:9092
 
 # kafka topic to write to
+# this can be static, e.g. 'maxwell', or dynamic, e.g. namespace_%{database}_%{table}
+# in the latter case 'database' and 'table' will be replaced with the values for the row being processed
 #kafka_topic=maxwell
 
 # kafka topic to write DDL to
+# this can be static, e.g. 'maxwell_ddl', or dynamic, e.g. namespace_ddl_%{database}_%{table}
+# in the latter case 'database' and 'table' will be replaced with the values for the row being processed
```
osheroff: this is not true, should actually include a note here about how ddl-topics are *not* dynamic.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -129,9 +142,10 @@ public void push(RowMap r) throws Exception {
 
 		ProducerRecord<String, String> record;
 		if (r instanceof DDLMap) {
-			record = new ProducerRecord<>(ddlTopic, this.ddlPartitioner.kafkaPartition(r, this.numPartitions), key, value);
+			record = new ProducerRecord<>(this.ddlTopic, this.ddlPartitioner.kafkaPartition(r, getNumPartitions(this.ddlTopic)), key, value);
```
osheroff: hey, you even fixed a bug!  nice.

osheroff:
good to go after documentation fixes

osheroff:
oh and maybe lowering the metadata fetch

smferguson:
i left the metadata timeout as-is. when i can connect to kafka i see an immediate response. when i can't it takes a while to timeout, but so do other kafka calls.

osheroff:
tested locally, looks good.

Issue #458, opened 2016-10-26
alter table statements blow up if 'cascade' keyword is specified


smferguson:
15:42:02,852 ERROR MysqlParserListener - (alter_specifications (alter_specification (drop_column DROP COLUMN (old_col_name (name (id `metadata`))))) CASCADE)
15:42:02,852 ERROR SchemaChange - Error parsing SQL: 'ALTER TABLE `journal` DROP COLUMN `metadata` CASCADE'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: CASCADE
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:84)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parseSQL(SchemaChange.java:79)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:91)
	at com.zendesk.maxwell.schema.AbstractSchemaStore.resolveSQL(AbstractSchemaStore.java:47)
	at com.zendesk.maxwell.schema.MysqlSchemaStore.processSQL(MysqlSchemaStore.java:82)
	at com.zendesk.maxwell.replication.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:462)
	at com.zendesk.maxwell.replication.MaxwellReplicator.getRow(MaxwellReplicator.java:432)
	at com.zendesk.maxwell.replication.MaxwellReplicator.work(MaxwellReplicator.java:159)
	at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:35)
	at com.zendesk.maxwell.Maxwell.start(Maxwell.java:160)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:181)
15:42:02,855 INFO  Maxwell - starting shutdown

osheroff:
[...]

PR #460, opened 2016-10-27
Alter table drop column with cascade

smferguson:
fix for issue #458, add support for 'CASCADE' keyword in drop column statements


osheroff:
👍

Issue #451, opened 2016-10-07
Question: metadata.fetch.timeout.ms vs. max.block.ms


bmichaud:
I am getting this message in my error.log on start-up, but I don't see the max.block.ms string anywhere in the project. Is this message out-of-date?

WARN  KafkaProducer - metadata.fetch.timeout.ms config is deprecated and will be removed soon. Please use max.block.ms

osheroff:
no, the message is actually right.  Maxwell can use either the 0.8.2.1 kafka client libs (which use metadata.fetch.timeout.ms), or the 0.9.0.1 version, which doesn't.  I guess I might go ahead and remove the default configuration values for kafka and just drop them into config.properties.example instead.

smferguson:
@osheroff seems like we could just remove the values from the producer and put them back in the .properties file, right? too simple?

osheroff:
we moved the defaults out of maxwell into config.properties in v1.5.0

Issue #363, opened 2016-06-14
add event timestamp in kafka messages. 


xmlking:
Apache Kafka 0.10.0.0 now supports event time,  received time and processing time concept. 
It would not nice if Maxwell set timestamp into messages so that we can do some interesting windowing operations in `kafka streams`

https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message
http://docs.confluent.io/3.0.0/release-notes.html 
`Timestamps in Mesages. Messages are now tagged with timestamps at the time they are produced, allowing a number of future features including looking up message by time and measuring timing. KAFKA-2511`

osheroff:
yup, quite reasonable request.  I think this would mostly just involve packaging maxwell with a 0.10 producer, which means we'd have to release different versions of maxwell depending on your desired kafka version.

smferguson:
unclear how best to move forward, but i'd love to see maxwell keeping up with kafka.
the upgrade does have some notables (http://kafka.apache.org/documentation.html#upgrade_10):
> Notes to clients with version 0.9.0.0: Due to a bug introduced in 0.9.0.0, clients that depend on ZooKeeper (old Scala high-level Consumer and MirrorMaker if used with the old consumer) will not work with 0.10.0.x brokers. Therefore, 0.9.0.0 clients should be upgraded to 0.9.0.1 before brokers are upgraded to 0.10.0.x. This step is not necessary for 0.8.X or 0.9.0.1 clients.

http://kafka.apache.org/documentation.html#upgrade_10_breaking

http://kafka.apache.org/documentation.html#upgrade_10_performance_impact
[...]

PR #461, opened 2016-10-27
Kafka 0.10 support

smferguson:
expecting (hoping) to start a conversation with this. my company's going to want (need?) 0.10 support in the very near future so I'd like to hear thoughts/musings around how it might be accomplished.


osheroff:
I think this is fine.  I can merge as is after some minimal testing (provided that the thing compiles with 0.10.0).  I think @dadah89 put some code in .travis.yml to at least ensure the code compiles against each version of the kafka-clients library, you may want to make sure that that is still working and is compiling maxwell against 0.10.0.

Maybe in a world with more time Maxwell could probe the kafka cluster, discover the version and then automatically choose the right version of the producer library.  But a read-through of the protocol doesn't really seem to indicate a reliable way of determining the cluster version (at least, not straightforward), so this way will do just fine for now.

osheroff:
ok, this *looks* good, but falls on its face pretty hard:

```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/kafka/common/serialization/Serializer
```

meaning the 0.9 and 0.10 libraries aren't interface compatible.  So we'll either have to come up with a solution that works with both interfaces, or we'll have to compile a little Maxwell->Kafka adapter class for each version of kafka.  

the adapter may be the right solution long term anyway, not sure.

osheroff:
never mind that last comment.  I just didn't compile the damn thing correctly.

osheroff:
I'm back to "this looks fine", although:

- the options need documenting (and should probably just remove the --kafka0.8 option
- it should support both `--kafka_version 0.9` and `--kafka_version=0.9`

smferguson:
related to issue #363

docs/docs/config.md
```diff
@@ -32,7 +32,7 @@ kafka_partition_columns        | STRING                              | if partit
 kafka_partition_by_fallback    | [database┃table┃primary_key]        | required when kafka_partition_by=column.  Used when the column is missing |
 kafka_partition_hash           | [default┃murmur3]                   | hash function to use when hoosing kafka partition   | default
 ddl_kafka_topic                | STRING                              | if output_ddl is true, kafka topic to write DDL changes to | *kafka_topic*
-kafka0.8                       | BOOLEAN                             | run maxwell with kafka producer 0.8 (instead of the 0.9) |
+kafka_version                  | [0.8|0.9|0.10]                      | run maxwell with kafka producer 0.8.2, 0.9.0 or 0.10.0 | 0.9.0
```
osheroff: careful, the pipes here will break the table formatting (sadly).  try copy/pasting the funky looking pipe character from above.  Or if you have a better solution I'm all ears.

osheroff: also, check your doc changes with `mkdocs serve`, after installing the python 'mkdocs' package

smferguson: 👍
internets say use "&#124;". i'm messing with that at the moment.

smferguson:
lmk what you think of the alternative ```&#124;```

smferguson:
@osheroff hang on for a bit. think i found a bug.

smferguson:
@osheroff ok. feeling better about this now.

osheroff:
ok.  There's bonus points here for exposing the kafka version in `MaxwellConfig` or something so we can make https://github.com/zendesk/maxwell/issues/451 go away.  But this looks totally fine to me.  👍 and will merge after test

smferguson:
feature creeeeeep! seriously tho. I'll try to take a look at it this week, but can't promise anything. adding 0.10 (along with the other pieces) gets us to the point we can start profiling maxwell/kafka to understand its performance/monitoring needs before we push to prod.
you guys should write a blog about the scale you run maxwell at/what it's like living with over time ;)

PR #464, opened 2016-11-01
push defaults into the properties file

smferguson:
felt best to push them all into properties (as noted in the related issue). i think it future-proofs maxwell a little better.
related to #451

### DIFF
--- a/docs/docs/kafka.md
+++ b/docs/docs/kafka.md
@@ -4,11 +4,10 @@ Any options given to Maxwell that are prefixed with `kafka.` will be passed dire
 (with `kafka.` stripped off).  We use the "new producer" configuration, as described here:
 [http://kafka.apache.org/documentation.html#newproducerconfigs](http://kafka.apache.org/documentation.html#newproducerconfigs)
 
-Maxwell sets the following Kafka options by default via the properties file. You can remove or update them in `config.properties`.
-
+These are recommended properties. You can remove or update them in `config.properties`.
 - kafka.acks = 1
-- kafka.compression.type = gzip
-- kafka.metadata.fetch.timeout.ms=50000
+- kafka.compression.type = snappy
+- kafka.metadata.fetch.timeout.ms=5000
 - kafka.retries=0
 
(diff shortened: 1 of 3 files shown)


docs/docs/kafka.md
```diff
@@ -4,10 +4,12 @@ Any options given to Maxwell that are prefixed with `kafka.` will be passed dire
 (with `kafka.` stripped off).  We use the "new producer" configuration, as described here:
 [http://kafka.apache.org/documentation.html#newproducerconfigs](http://kafka.apache.org/documentation.html#newproducerconfigs)
 
-Maxwell sets the following Kafka options by default, but you can override them in `config.properties`.
+Maxwell sets the following Kafka options by default via the properties file. You can remove or update them in `config.properties`.
 
 - kafka.acks = 1
 - kafka.compression.type = gzip
+- kafka.metadata.fetch.timeout.ms=50000
+- kafka.retries=0
```
osheroff: should change the verbage a bit; you can just say "these are recommended properties" or something.  I'll also put some examples of more strongly consistent setups and such after I merge.

smferguson:
looking forward to the additional examples. I'd love to hear other recommendations for running maxwell at scale. it's going to be a critical piece of our infrastructure before the end of this year.
if you guys weren't so far from boston I'd offer beer in return.

osheroff:
mmm, beer.  here's the settings we ended up at:

```
      'kafka.max.request.size' => 10_000_000,
      'kafka.retries' => 5,
      'kafka.acks' => "all",
      'kafka.compression.type' => "snappy",
      'kafka.batch.size' => 100_000,
```

the retries/acks settings are because we couldn't ever seem to keep our kafka cluster from occasionally going into a leader-election, which would get maxwell dropping messages.  

Snappy > gzip for performance, I think we talked about that.  

The buffer size shouldn't be cranked too hard, but it could still probably go higher for performance reasons.

Running multiple maxwells, each going to dedicated topics is *probably* better than going with one giant firehose and re-filtering it.  

the binlog (and Maxwell, by extension) tends to be fairly serial in nature.  The quicker you get your data into many partitions for parallel processing the better.

Latency and lag tends to be the enemy.

That's all I can think of for now.

smferguson:
👍 thanks!

PR #472, opened 2016-11-09
add backquote to db name in sql

mylesjao:
**maxwell-bootstrap** will generate sql with syntax error if database name contains "-". i.e. "test-db".
I added backquote to database part in sql template to resolve this issue.


osheroff:
cool, thanks for the patch 👍

PR #473, opened 2016-11-09
add confluent schema reg support

smferguson:
step 1 of adding support for avro + confluent schema reg. step 2 would be dynamic schemas for the actual data, but i'm not sure i'll have time to tackle that one soon.
this isn't complete (it is fully functional), but i'd like a sanity check. there are a few things i'm not in love with. the biggest offender being the avro + confluent deps that add a lot of bloat (new jars listed below). other things like the static nature of the fields in avro schema which rm's some of the output flexibilty, bug me, but could be justifiable.
new jars:
 - avro-1.7.6.jar
 - common-config-3.0.1.jar
 - common-utils-3.0.1.jar
 - commons-compress-1.4.1.jar
 - jackson-core-asl-1.9.13.jar
 - jackson-mapper-asl-1.9.13.jar
 - jline-0.9.94.jar
 - kafka-avro-serializer-3.0.1.jar
 - kafka-schema-registry-client-3.0.1.jar
 - netty-3.2.2.Final.jar
 - paranamer-2.3.jar
 - xz-1.0.jar
 - zkclient-0.5.jar
 - zookeeper-3.4.3.jar

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -125,6 +125,12 @@ kafka.metadata.fetch.timeout.ms=5000
 kafka.retries=0
 kafka.acks=1
 
+# confluent schema registry specifics
+#kafka.schema.registry.url=http://localhost:8081
+#kafka.key.serializer=io.confluent.kafka.serializers.KafkaAvroSerializer
+#kafka.value.serializer=io.confluent.kafka.serializers.KafkaAvroSerializer
+
+
 ######### filter stuff ###############
 
 # filter rows out of Maxwell's output.
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,24 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
+  <repositories>
+    <repository>
+      <id>confluent</id>
+      <url>http://packages.confluent.io/maven/</url>
+    </repository>
+  </repositories>
+
   <dependencies>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
+      <version>1.7.6</version>
+    </dependency>
+    <dependency>
+      <groupId>io.confluent</groupId>
+      <artifactId>kafka-avro-serializer</artifactId>
+      <version>3.0.1</version>
+    </dependency>
     <dependency>
       <groupId>mysql</groupId>
       <artifactId>mysql-connector-java</artifactId>
--- a/src/main/java/com/zendesk/maxwell/MaxwellContext.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellContext.java
@@ -220,6 +220,9 @@ public class MaxwellContext {
 		case "kafka":
 			this.producer = new MaxwellKafkaProducer(this, this.config.getKafkaProperties(), this.config.kafkaTopic);
 			break;
+		case "kafka-avro":
+			this.producer = new ConfluentAvroProducer(this, this.config.getKafkaProperties(), this.config.kafkaTopic);
+			break;
 		case "profiler":
 			this.producer = new ProfilerProducer(this);
 			break;
--- a/src/main/java/com/zendesk/maxwell/producer/ConfluentAvroProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/ConfluentAvroProducer.java
@@ -1,0 +1,128 @@
+package com.zendesk.maxwell.producer;
+
+import com.zendesk.maxwell.MaxwellContext;
+import com.zendesk.maxwell.producer.partitioners.MaxwellKafkaPartitioner;
+import com.zendesk.maxwell.replication.BinlogPosition;
+import com.zendesk.maxwell.row.RowMap;
+import com.zendesk.maxwell.row.RowMap.KeyFormat;
+import com.zendesk.maxwell.schema.ddl.DDLMap;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.KafkaException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.sql.SQLException;
+import java.util.Properties;
+
+public class ConfluentAvroProducer extends AbstractProducer {
+    static final Logger LOGGER = LoggerFactory.getLogger(ConfluentAvroProducer.class);
+
+    private final InflightMessageList inflightMessages;
+    private final KafkaProducer<String, Object> kafka;
+    private String topic;
+    private final String ddlTopic;
+    private final MaxwellKafkaPartitioner partitioner;
+    private final MaxwellKafkaPartitioner ddlPartitioner;
+    private final KeyFormat keyFormat;
+
+    public ConfluentAvroProducer(MaxwellContext context, Properties kafkaProperties, String kafkaTopic) {
+        super(context);
+
+        this.topic = kafkaTopic;
+        if ( this.topic == null ) {
+            this.topic = "maxwell";
+        }
+
+        this.kafka = new KafkaProducer<>(kafkaProperties);
+
+        String hash = context.getConfig().kafkaPartitionHash;
+        String partitionKey = context.getConfig().kafkaPartitionKey;
+        String partitionColumns = context.getConfig().kafkaPartitionColumns;
+        String partitionFallback = context.getConfig().kafkaPartitionFallback;
+        this.partitioner = new MaxwellKafkaPartitioner(hash, partitionKey, partitionColumns, partitionFallback);
+        this.ddlPartitioner = new MaxwellKafkaPartitioner(hash, "database", null,"database");
+        this.ddlTopic =  context.getConfig().ddlKafkaTopic;
+
+        if ( context.getConfig().kafkaKeyFormat.equals("hash") )
+            keyFormat = KeyFormat.HASH;
+        else
+            keyFormat = KeyFormat.ARRAY;
+
+        this.inflightMessages = new InflightMessageList();
+    }
+
+    private Integer getNumPartitions(String topic) {
+        try {
+            return this.kafka.partitionsFor(topic).size(); //returns 1 for new topics
+        } catch (KafkaException e) {
+            LOGGER.error("Topic '" + topic + "' name does not exist. Exception: " + e.getLocalizedMessage());
+            throw e;
+        }
+    }
+
+    private String generateTopic(String topic, RowMap r){
+        return topic.replaceAll("%\\{database\\}", r.getDatabase()).replaceAll("%\\{table\\}", r.getTable());
+    }
+
+    @Override
+    public void push(RowMap r) throws Exception {
+        String key = r.pkToJson(keyFormat);
+        String value = r.toJSON(outputConfig);
+
+        if ( value == null ) { // heartbeat row or other row with suppressed output
+            inflightMessages.addMessage(r.getPosition());
+            BinlogPosition newPosition = inflightMessages.completeMessage(r.getPosition());
+
+            if ( newPosition != null ) {
+                context.setPosition(newPosition);
+            }
+
+            return;
+        }
+
+        KafkaCallback callback;
+        ProducerRecord<String, Object> record;
+        if (r instanceof DDLMap) {
+
+            record = new ProducerRecord<>(this.ddlTopic, this.ddlPartitioner.kafkaPartition(r, getNumPartitions(this.ddlTopic)), key, (Object) value);
+
+            callback = new KafkaCallback(inflightMessages, r.getPosition(), r.isTXCommit(), this.context, key, value);
+        } else {
+            MaxwellEnvelope maxwellEnvelope = new MaxwellEnvelope(
+                    r.getDatabase(),
+                    r.getTable(),
+                    r.getRowType(),
+                    r.getTimestamp(),
+                    r.getXid(),
+                    r.isTXCommit(),
+                    r.getPosition().toString(),
+                    r.getServerId(),
+                    r.getThreadId(),
+                    r.getData(outputConfig.includesNulls),
+                    r.getOldData(outputConfig.includesNulls));
+
+            String topic = generateTopic(this.topic, r);
+            record = new ProducerRecord<>(topic, this.partitioner.kafkaPartition(r, getNumPartitions(topic)), key, (Object) maxwellEnvelope);
+
+            callback = new KafkaCallback(inflightMessages, r.getPosition(), r.isTXCommit(), this.context, key, maxwellEnvelope);
+        }
+
+        if ( r.isTXCommit() )
+            inflightMessages.addMessage(r.getPosition());
+
+
+		/* if debug logging isn't enabled, release the reference to `value`, which can ease memory pressure somewhat */
+        if ( !KafkaCallback.LOGGER.isDebugEnabled() )
+            value = null;
+
+        kafka.send(record, callback);
+    }
+
+    @Override
+    public void writePosition(BinlogPosition p) throws SQLException {
+        // ensure that we don't prematurely advance the binlog pointer.
+        inflightMessages.addMessage(p);
+        inflightMessages.completeMessage(p);
+    }
+}
--- a/src/main/java/com/zendesk/maxwell/producer/KafkaCallback.java
+++ b/src/main/java/com/zendesk/maxwell/producer/KafkaCallback.java
@@ -1,0 +1,58 @@
+package com.zendesk.maxwell.producer;
+
+import com.zendesk.maxwell.MaxwellContext;
+import com.zendesk.maxwell.replication.BinlogPosition;
+import org.apache.kafka.clients.producer.Callback;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.errors.RecordTooLargeException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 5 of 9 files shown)


osheroff:
can't say that I'm a big fan of the extra deps either, but the apache-avro ones are unavoidable.

Looks like maybe the confluent ones could be cut down?

```
[INFO] +- org.apache.avro:avro:jar:1.7.6:compile
[INFO] |  +- org.codehaus.jackson:jackson-core-asl:jar:1.9.13:compile
[INFO] |  +- org.codehaus.jackson:jackson-mapper-asl:jar:1.9.13:compile
[INFO] |  +- com.thoughtworks.paranamer:paranamer:jar:2.3:compile
[INFO] |  +- org.xerial.snappy:snappy-java:jar:1.0.5:compile
[INFO] |  +- org.apache.commons:commons-compress:jar:1.4.1:compile
[INFO] |  |  \- org.tukaani:xz:jar:1.0:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.6.4:compile
[INFO] +- io.confluent:kafka-avro-serializer:jar:3.0.1:compile
[INFO] |  +- io.confluent:kafka-schema-registry-client:jar:3.0.1:compile
[INFO] |  |  \- org.slf4j:slf4j-log4j12:jar:1.7.6:compile
[INFO] |  |     \- log4j:log4j:jar:1.2.17:compile
[INFO] |  \- io.confluent:common-config:jar:3.0.1:compile
[INFO] |     \- io.confluent:common-utils:jar:3.0.1:compile
[INFO] |        \- com.101tec:zkclient:jar:0.5:compile
[INFO] |           \- org.apache.zookeeper:zookeeper:jar:3.4.3:compile
[INFO] |              +- jline:jline:jar:0.9.94:compile
[INFO] |              \- org.jboss.netty:netty:jar:3.2.2.Final:compile
```

ie if we picked up `kafka-schema-registry-client` instead of `kafka-avro-serializer`, we'd reduce the deps a lot

osheroff:
I don't mind the static nature of the fields, really, but I'm a little squinty at the shove-JSON-inside-avro techique; it doesn't seem like the feature is really something that should be merged until you're pushing schemas into the registry.  Is the thought that you'd maintain some list of the schemas you've already generated/pushed and push new ones in as you see table updates?

Another general question; do the confluent tools tend to go fetch schemas on demand?  If I have to fetch and update my avro schemas in lock-step with my mysql schema changes, that seems like a schlep. 

A lot of these questions had me skirting around doing avro in the first place.  If you/we can find the right answers to them, there's definitely value in avro; it's far more compressed and has real timestamps/blobs!  It's a pity they didn't include a bucket "map-of-junk-i-don't-want-to-have-a-schema-for" type.

smferguson:
re: actually supporting schemas - yup. I agree about the merge. it's 1/2 a feature right now. ideally our avro producer would dynamically generate the schema based on the rowmap's data - that's the fun part. I'm looking into it now. for maintaining table updates see below (schema evolutions are supported).

re: confluent tools. yes, their avro producer does a side-lookup in the schema reg whenever it gets a schema it doesn't know. it caches that result to keep the roundtrips to a min. on the schema reg side, if it gets a new version of an existing schema it updates that to the latest (it keeps all versions). it stores a schema per topic - there is no way to have multiple schemas for one topic. you can have multiple schema versions (evolutions of the same table) per topic. this also means that maxwell has to send each table to its own topic (obviously). regarding the schema evolutions, the reg has the following compatibility levels (http://docs.confluent.io/3.0.1/schema-registry/docs/api.html):
> Backward compatibility (default): A new schema is backward compatible if it can be used to read the data written in all previous schemas. Backward compatibility is useful for loading data into systems like Hadoop since one can always query data of all versions using the latest schema.
> Forward compatibility: A new schema is forward compatible if all previous schemas can read data written in this schema. Forward compatibility is useful for consumer applications that can only deal with data in a particular version that may not always be the latest version.
> Full compatibility: A new schema is fully compatible if it’s both backward and forward compatible.
No compatibility: A new schema can be any schema as long as it’s a valid Avro.

re: "map-of-junk-i-don't-want-to-have-a-schema-for" type - I thought that's what strings were for ;)

I'll try to get the deps under control and dig into dynamic schemas. thanks for the input.

smferguson:
@osheroff i don't consider this complete yet. still a few outliers and a bit of testing that needs to be done.

smferguson:
@osheroff looking at the build error... looks like it's blowing up b/c it's trying to use the standard maven repo instead of the confluent one. any chance that's a config tweak not a code change?
>$ make test
> java -cp target/dependency-antlr/antlr4-runtime-4.5.jar:target/dependency-antlr/antlr-runtime-3.5.2.jar:target/dependency-antlr/ST4-4.0.8.jar:target/dependency-antlr/org.abego.treelayout.core-1.0.1.jar:target/dependency-antlr/antlr4-4.5.jar org.antlr.v4.Tool -package com.zendesk.maxwell.schema.ddl -lib src/main/antlr4/imports -o target/generated-sources src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
build/maven_fetcher:177:in `get': not found: /maven2/io/confluent/kafka-avro-serializer/3.0.1/kafka-avro-serializer-3.0.1.pom (RuntimeError)

osheroff:
no, of course, that's my craziness for never wanting to deal with maven itself.  I'll take on adding repository support when I get a sec.

smferguson:
thanks. i'd like opinions on the rest of the code when you're in the mood. at the moment i'm inclined to convert bigdecimal to an avro float (vs string or bytes, which is what the complex type compiles down to). i'm also inclined to leave dates and timestamps as strings b/c they're complex avro types too. i could be swayed on it though. and, of course, this needs a bit of doc added.

osheroff:
what do you mean by "compiles down to"?  Are you referring to what the avro looks like in json form?

smferguson:
first, i used the wrong word. i should've said 'logical' types not 'complex' types. then, yes. using decimal as an example the avro type is defined as
> {
>   "type": "bytes",
>   "logicalType": "decimal",
>   "precision": 4,
>   "scale": 2
> }

to get the data to serialize i have to convert bigdecimal to a bytebuffer. on the consumer side the data then looks like bytes:
> "foo": {"bytes": "!"}

instead of
> "foo": 13.13

we're using kafka for inter-service-communication and that syntax change makes life a lot harder on the backend guys. it's arguable that they could deal with it, but they're very pressed for time and, tbh, i'm not a fan of certain values being dictionaries and others being just values
fyi, https://avro.apache.org/docs/current/spec.html#schema_primitive, https://avro.apache.org/docs/current/spec.html#Logical+Types

osheroff:
oh, so I'd presumed that if you want, avro, you want the binary form of it (http://avro.apache.org/docs/1.8.1/spec.html#binary_encoding).  Word I've gotten when I poked my nose into the JSON-encoding is that it's just not quite there; mostly it's for debugging and doesn't appear to be fully-defined -- see for example what various libs do for binary data (do you output invalid JSON?  Do you double-escape?) -- and that you really shouldn't use it in production.

smferguson:
interesting. the big wins for us are being able to use the schema registry for validation and having the schema (with docs) as part of the message. tbh i hadn't heard that json encoding wasn't prod-ready. the posts that are kind of about that in google are fairly old (2012). maybe i'm missing something?

osheroff:
it may very well be prod-ready for some libraries.  We found that the ruby avro library wasn't compatible (at least with blobs) with the java library as far as JSON support went, and then some dev said "yeah, avro-json is really just for debugging.  you should use the binary version."

It may be cargo-cultish, so take it with a grain of salt, but IMO it's probably more "right" to design around the binary encoding and just deal with whatever the JSON encoding offers, since the binary encoding is likely more prevalent.

smferguson:
good points. i saw some interesting commentary around the ruby libraries as well. we're a java/python shop so i think we're in good shape.
other thoughts on the pr? (dare i ask)

smferguson:
rethinking this a bit... i think it  makes more sense to have the user specify a uri that contains schemas for their tables (lookups likely keyed by table name) instead of the dynamic bits.

osheroff:
but what would maxwell do on an incompatible schema change?  I kinda like the idea that maxwell is a producer into a well-known funnel, but it seems oddly restrictive.

smferguson:
the schema reg's permissiveness can be set by the end-user so, in theory, breaking changes could be allowed, but I haven't proven that out.
in my specific case, we want to prevent breaking changes from entering the system (fail fast) so maxwell blowing up immediately is the desired behavior. it's definitely a bit of a mind-shift. being a data guy i love it :)

osheroff:
pinged the avro-issue for more use cases.  If we're gonna insist that users supply schemas up-front, I'd say we go one step further and decouple maxwell from the details of the schema-registry -- just give maxwell a directory where it can find its avro schemas (named by database/table.avro, maybe?), and have it pick up schemas from that directory on demand.

osheroff:
In thinking about it more, I'd *really* like to target binary-encoded avro, meaning utilize complex types.  If you're dead set on using avro/json, let's build in support for field coercion, meaning you can declare a date or timestamp or blob field as a string, and maxwell will understand that and coerce the data into the type you'd like.

osheroff:
did I ever tell you my slightly hare-brained idea, which was that maxwell would just implement the same URI-scheme as the confluent schema registry, dynamically generating avro schemas and serving them itself?  anyway......

smferguson:
re: maxwell picking up the schemas on-the-fly. i'm actually working on something along those lines. it's a little deeper than that, but yeah. def agree.
re: binary/real types. i agree. i'm going to give it another look when i have a spare cycle.
re: hair brained (bugs bunny on the brain now). ha!

osheroff:
the pre-baked approach definitely seems to solve some of the weird edges here, and makes maxwell+avro usuable without external dependencies (love that), so sure, give it a shot as a v1 and go from there.  people can always bolt on schema generation later.

smferguson:
not worth a look yet. just catching it up with a child branch that has a bunch of specific code around our use-case (embedded json in a text column that we want to apply schemas against) in addition to generic code (yay!).
one notable - bumping the avro version from 1.7.7 to 1.8.1 breaks make. hoping you have a little magic here:
>build/maven_fetcher:95:in `get_var': undefined method `get_var' for nil:NilClass (NoMethodError)
	from build/maven_fetcher:95:in `get_var'
	from build/maven_fetcher:95:in `get_var'
	from build/maven_fetcher:101:in `block in subst'
	from build/maven_fetcher:100:in `gsub'
	from build/maven_fetcher:100:in `subst'
	from build/maven_fetcher:111:in `first_text'
	from build/maven_fetcher:115:in `artifact_from_element'
	from build/maven_fetcher:156:in `block in dependencies'
	from build/maven_fetcher:154:in `map'
	from build/maven_fetcher:154:in `dependencies'
	from build/maven_fetcher:226:in `block in resolve'
	from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/set.rb:232:in `each_key'
	from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/set.rb:232:in `each'
	from build/maven_fetcher:223:in `map'
	from build/maven_fetcher:223:in `resolve'
	from build/maven_fetcher:241:in `resolve'
	from build/maven_fetcher:269:in `fetch_pom_deps'
	from build/maven_fetcher:358:in `<main>'
> # Fetch jar so we can run it locally (bin/maxwell)
> build/maven_fetcher -f org.apache.kafka/kafka-clients/0.8.2.2 --skip-dependencies -o target/dependency >/dev/null
> build/maven_fetcher -f org.apache.kafka/kafka-clients/0.10.0.1 --skip-dependencies -o target/dependency >/dev/null
> javac -classpath  -d target/classes -sourcepath src/main/java:src/test/java:target/generated-sources/src/main/antlr4 -g -target 1.7 -source 1.7 -encoding UTF-8 -Xlint:-options -Xlint:unchecked src/main/java/com/zendesk/maxwell/MaxwellConfig.java src/main/java/com/zendesk/maxwell/MaxwellContext.java src/main/java/com/zendesk/maxwell/producer/AbstractKafkaProducer.java src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java src/main/java/com/zendesk/maxwell/producer/SchemaRegistryProducer.java
> javac: invalid flag: target/classes
> Usage: javac <options> <source files>
> use -help for a list of possible options
> make: *** [target/.java] Error 2

osheroff:
```
diff --git a/build/maven_fetcher b/build/maven_fetcher
index d027568..678e33d 100755
--- a/build/maven_fetcher
+++ b/build/maven_fetcher
@@ -89,8 +89,8 @@ module MavenFetcher
     def get_var(k)
       @vars    ||= {}
       @vars[k] ||= case k
-        when "project.version"
-          first_text(@doc.root, "version") || first_text(@doc.root, "parent/version")
+        when /project.(\w+)/
+          first_text(@doc.root, $1) || first_text(@doc.root, "parent/#{$1}")
         else
           first_text(@doc.root, "properties/#{k}") || parent.get_var(k)
         end
```

fixes the pom issue I think

smferguson:
@osheroff i feel like maxwell is trying to force me to learn ruby

> 7.57s$ sudo apt-get -qq update
before_install.2
3.77s$ sudo apt-get install pastebinit
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following NEW packages will be installed:
  pastebinit
0 upgraded, 1 newly installed, 0 to remove and 118 not upgraded.
Need to get 16.1 kB of archives.
After this operation, 176 kB of additional disk space will be used.
Get:1 http://us-central1.gce.archive.ubuntu.com/ubuntu/ precise-updates/universe pastebinit all 1.3-2ubuntu2.1 [16.1 kB]
Fetched 16.1 kB in 0s (0 B/s)
Selecting previously unselected package pastebinit.
(Reading database ... 81948 files and directories currently installed.)
Unpacking pastebinit (from .../pastebinit_1.3-2ubuntu2.1_all.deb) ...
Processing triggers for man-db ...
Setting up pastebinit (1.3-2ubuntu2.1) ...
install
0.00s$ true
7.13s$ make test
java -cp target/dependency-antlr/antlr4-runtime-4.5.jar:target/dependency-antlr/antlr-runtime-3.5.2.jar:target/dependency-antlr/ST4-4.0.8.jar:target/dependency-antlr/org.abego.treelayout.core-1.0.1.jar:target/dependency-antlr/antlr4-4.5.jar org.antlr.v4.Tool -package com.zendesk.maxwell.schema.ddl -lib src/main/antlr4/imports -o target/generated-sources src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
build/maven_fetcher:59:in `join': no implicit conversion of nil into String (TypeError)
	from build/maven_fetcher:59:in `cache_path'
	from build/maven_fetcher:51:in `cachefile'
	from build/maven_fetcher:311:in `get_file_from_cache'
	from build/maven_fetcher:295:in `get_pom'
	from build/maven_fetcher:224:in `block in resolve'
	from /home/travis/.rvm/rubies/ruby-2.2.5/lib/ruby/2.2.0/set.rb:283:in `each_key'
	from /home/travis/.rvm/rubies/ruby-2.2.5/lib/ruby/2.2.0/set.rb:283:in `each'
	from build/maven_fetcher:223:in `map'
	from build/maven_fetcher:223:in `resolve'
	from build/maven_fetcher:241:in `resolve'
	from build/maven_fetcher:269:in `fetch_pom_deps'
	from build/maven_fetcher:358:in `<main>'

osheroff:
I wouldn't get involved if I were you :) this maven/pom.xml business is pretty terrible.  That being said, anytime I've peeked inside a version management system it's been pretty terrible.

osheroff:
it doesn't like 

```
    <dependency>
      <groupId>org.apache.directory.studio</groupId>
      <artifactId>org.apache.commons.io</artifactId>
      <version>2.4</version>
    </dependency>
```

will this work instead?

```
<!-- https://mvnrepository.com/artifact/commons-io/commons-io -->
<dependency>
    <groupId>commons-io</groupId>
    <artifactId>commons-io</artifactId>
    <version>2.5</version>
</dependency>
```

smferguson:
just a heads-up. the pace here is going to slow... i've got some new priorities. i'll try to finish this up during off hours. unclear how that'll go.

osheroff:
hmmmmph.  would you just be finishing the code w/out shipping it anywhere?  For a big change like this I'd really want someone running and living with it in prod, rather than just sorta throwing the code out the door and seeing what's up.

smferguson:
we still need it (which wasn't at all clear from my message). the timeline was pushed back at least a few weeks - possibly until early jan. so it's not my day-job at the moment.

osheroff:
okey doke, no worries then.

smferguson:
slowly getting back to this. opinions on peeling off the refactoring (maven_fetcher, AbstractKafkaProducer.java, KafkaCallback.java, MaxwellKafkaProducer.java, RowMap.java) into a separate PR so it doesn't decay?

osheroff:
PR the maven_fetcher stuff for sure.  the refactors I'd leave, though; don't want to add more abstraction unless necessary.  Other than that, keep this PR merged and I'll try to remember to ping you if anyone goes mucking around in the kafka code

smferguson:
fyi, this officially got the axe from "on high" so even if I was to finish it we won't be running it in prod(!). feel free to close or leave for some other brave soul to finish later. dunno what's best here.

osheroff:
pity!  can you merge one last time to remove conflicts and then leave open?  Oftentimes someone will come along with the same needs....

smferguson:
👍 will do. and yeah :(

Issue #471, opened 2016-11-09
Unable to bootstrap when using a non-default schema_database


lucianjon:
Hey, we've been trying to get bootstrapping working when running multiple instances of maxwell against a binlog. We've had an instance running against the `maxwell` schema for a few months now. It's running v1.1.1 and bootstrapping works without issue.

Now we've set up another instance running v1.4.2/v1.5.0 against the same binlog. They've both been running in parallel for a couple of weeks without issue. The new instance sets the `schema_database` and `client_id` to something else to avoid conflicts. However rows inserted into the `bootstrap` table never get picked up. There's nothing in the logs (with debug) that indicates anything is happening. 

We've tried to keep the config between the two instances as similar as possible, the relevant differences being:

```
client_id=maxwell_two
schema_database=maxwell_two
```

We've tried a couple things to rule out some problems:
* Downgrading the new instance to v1.1.1 and running both in parralel (both crashed due to a binlog parse error in this case)
* Killing the old instance and running the new instance with v1.1.1 
* Killing the old instance and running the new instance with v1.5.0

Nothing has had any effect so far so it seems either we've got horribly wrong somewhere with config or possibly something isn't working with the non-default `schema_database`

osheroff:
Which bootstrap table did the `maxwell-bootstrap` tool end up inserting into, `maxwell.bootstrap` or `maxwell_two.bootstrap`?  I took a look at the code, it definitely pays attention to an alternate `schema_database` name:

```
	public boolean isBootstrapRow(RowMap row) {
		return row.getDatabase().equals(this.context.getConfig().databaseName) &&
			row.getTable().equals("bootstrap");
	}
```

having two maxwell processes pointing at separate schema_databases is a little goofy (in particular, they don't realize that `maxwell.positions` is a maintenance table, and they shouldn't output it), but I can see why you're doing that and there shouldn't, theoretically, be anything that prevents it from working.

lucianjon:
I had a look through the bootstrap utility code and saw it was just inserting the row and waiting on completion. So I've just been inserting the rows manually into `maxwell_two`. 

I also couldn't see a reason why it wouldn't be started given maxwell sees the binlog event. Anyway I may just recompile with some extra print statements to see where it's getting to.

lucianjon:
[...]

PR #479, opened 2016-11-16
fix bootstrapping with custom schema_database

lucianjon:
Fixes bootstrapping binlog events not being picked up when using a custom `schema_database`.
Addresses #471

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -384,6 +384,7 @@ public class MaxwellConfig extends AbstractConfig {
 					excludeTables,
 					blacklistDatabases,
 					blacklistTables,
+					this.databaseName,
 					excludeColumns
 			);
 		} catch (MaxwellInvalidFilterException e) {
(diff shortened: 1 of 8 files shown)


osheroff:
oh, is the missing piece of the puzzle here that you've also got an `include_database=` (or other `include`) filter configured?  that.... would make sense.

osheroff:
thanks for the patch!  tell ya what though, how about we move the hard-coded whitelist outside of the filter?  IMO it was in the wrong place anyway; IMO the place for the hard-coded check should be around line 280 in MaxwellReplicator.java, where we check the filter.

lucianjon:
Yep, we're using an `include_database` as well as `include_tables`. 

Yeah that makes sense to me, passing the `schema_database` into the filter class didn't feel too great. So you're thinking something like this?

```
if ( event.matchesFilter() || bootstrapRow ) {
 addToBuffer
}
```

osheroff:
yup, like that.  If you wanted to get fancy you could name it `isSystemWhitelisted` or something.

lucianjon:
Cool, that looks to be working. Just figuring out a way to add a test for this.

osheroff:
this looks good to me, were you planning on trying a test here?  you could maybe try a a bootstrap-integration test with an include-dbs filter or something.  i dunno.

lucianjon:
Yep, I'm working on a test in BootstrapIntegrationTest. I just have to add the ability to pass a filter down through the JSON helper functions. I also wanted to run it against a non-default maxwell schema but that looks a little tricky as it's hard coded inside `getRowsWithReplicator`

lucianjon:
@osheroff Got the test, does the PR look good to you?

osheroff:
looks good.  thanks for the work! 👍

PR #491, opened 2016-12-06
remove PK column before we remove from the column-list.

osheroff:
Otherwise we could crash, depending on whether we had memoized the column list or not.

Note that I started down a road of refactoring the primary-key list into `TableColumnList`, but I found it to be really, really nasty due to oddities over deserialization (and especially the order in which fields deserialize).


osheroff:
@zendesk/rules

PR #494, opened 2016-12-07
add support for 0.10.1

smferguson:
0.10.0 isn't fully compatible with 0.10.1 so...
I wonder if we want separate builds for the diff versions at some point.


osheroff:
those kafka folks are not crazy about backwards compat, eh?  I'd love to keep a single tarball for as long as possible (basically as long as the producer interface stays the same); the real trick is to somehow probe the cluster and figure out what version you're talking to, and then load the right libraries in runtime.

but that'd be quite some trick.  👍 here, will get this in with the next release.

smferguson:
thanks. and yeah, they made wire protocol changes between 0.9 and 0.10 and added new protocols in 0.10.1. they explicitly state that 0.10.1 producers and consumers require brokers of the same version :/

osheroff:
released in 1.5.2

PR #499, opened 2016-12-12
a few logging tweaks

osheroff:
- more accurate heartbeat logging
- better did-not-find messages in tests
- silence makefile

exceedingly minor.

@zendesk/rules


lileeyao:
👍 on green

PR #500, opened 2016-12-14
let HeartbeatRowMap advance the binlog position

osheroff:
this will cause an extra update to `positions` -- the flow now will go 

1.  Reading data triggers a position update
2.  This causes a heartbeat to send
3.  This advances the binlog position, and we update again.  Note that we do *not* send a heartbeat here.

But this should be more correct.  In particular, this should stabilize `RecoveryTest` immensely.

@zendesk/rules


osheroff:
closed in favor of #506

Issue #502, opened 2016-12-15
[Feature Request] Producer for AWS Kinesis.


thomasdziedzic:
I am investigating using maxwell for a project and would like to send events to kinesis if possible.
The benefits boil down to the benefits of using a managed service over a self-managed service.
Thoughts?

osheroff:
there's https://github.com/zendesk/maxwell/pull/280/files, which never quite made it across the finish line.  Would happily merge it if someone dusted it off and put it through some productionization paces.

osheroff:
wheee, in 1.7.0

PR #280, opened 2016-02-17
Adding kinesis as producer output sink support

jitendra-1217:
**What?**

Currently maxwell supports `kafka|stdout|file` as output sink. Adding [`kinesis`](https://aws.amazon.com/kinesis/) to that.
## 

**Why?**

Many are too much dependent on aws services, they find it difficult to get started with something not on aws. **Kinesis** would be an alternative to Kafka for such cases.
## 

**Implementation notes:**

_Needs to be fixed, will work on them when get time_
1. It puts results _synchronously_ (blocking call) and doesn't usage callback mechanism (as in [**kpl**](http://docs.aws.amazon.com/kinesis/latest/dev/developing-producers-with-kpl.html)).
2. It will stop/break execution in cases of exceptions of kind (aws kinesis stream resource not found, throughput exceeded exception) and won't attempt to re push the row data.
3. Need to ensure stream is already created in kinesis.
4. Using _database name_ as _partition key_.

PR #505, opened 2016-12-16
implement kinesis support

thomasdziedzic:
see https://github.com/zendesk/maxwell/issues/502 for ticket
see https://github.com/zendesk/maxwell/pull/280 for a previous attempt that was never completed

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -125,6 +125,9 @@ kafka.metadata.fetch.timeout.ms=5000
 kafka.retries=0
 kafka.acks=1
 
+######### kinesis ####################
+kinesis_stream=maxwell
+
 ######### filter stuff ###############
 
 # filter rows out of Maxwell's output.
--- a/config.properties.example
+++ b/config.properties.example
@@ -87,22 +87,6 @@ password=maxwell
 # function.
 #kafka_partition_hash=default # [default, murmur3]
 
-# controls the input input into the hash function.  Note that this defines
-# which transactions keep ordering with respect to each other.  Generally
-# here you're making a trade-off between inter-row consistency and balanced
-# partitions.
-#kafka_partition_by=database # [database, table, primary_key, column]
-
-# required when using kafka_partition_by=column
-# otherwise the partitioner will revert to table
-# can be a single or multiple columns, e.g. aggregate_id or aggregate_id,event_type
-#kafka_partition_columns=
-
-# required when using kafka_partition_by=column
-# the fallback partitioning behavior when the specified column(s) do not exist
-# can be any one of [database, table, primary_key]
-#kafka_partition_by_fallback=database
-
 # how maxwell writes its kafka key.
 #
 # 'hash' looks like:
@@ -124,6 +108,25 @@ kafka.compression.type=snappy
 kafka.metadata.fetch.timeout.ms=5000
 kafka.retries=0
 kafka.acks=1
+
+####### producer partitioning #######
+# used by kafka and kinesis
+
+# controls the input input into the hash function.  Note that this defines
+# which transactions keep ordering with respect to each other.  Generally
+# here you're making a trade-off between inter-row consistency and balanced
+# partitions.
+#producer_partition_by=database # [database, table, primary_key, column]
+
+# required when using producer_partition_by=column
+# otherwise the partitioner will revert to table
+# can be a single or multiple columns, e.g. aggregate_id or aggregate_id,event_type
+#producer_partition_columns=
+
+# required when using producer_partition_by=column
+# the fallback partitioning behavior when the specified column(s) do not exist
+# can be any one of [database, table, primary_key]
+#producer_partition_by_fallback=database
 
 ######### kinesis ####################
 kinesis_stream=maxwell
--- a/config.properties.example
+++ b/config.properties.example
@@ -138,7 +138,7 @@ kinesis_stream=maxwell
 # before we send it to kinesis so all the keys work within the key size limit.
 # Values: true, false
 # Default: false
-#kinesis_hash_key=true
+#kinesis_md5_keys=true
 
 ######### filter stuff ###############
 
--- a/config.properties.example
+++ b/config.properties.example
@@ -131,6 +131,15 @@ kafka.acks=1
 ######### kinesis ####################
 kinesis_stream=maxwell
 
+# AWS places a 256 unicode character limit on the max key length of a record
+# http://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html
+#
+# Setting this option to true enables hashing the key with the md5 algorithm
+# before we send it to kinesis so all the keys work within the key size limit.
+# Values: true, false
+# Default: false
+#kinesis_hash_key=true
+
 ######### filter stuff ###############
 
 # filter rows out of Maxwell's output.
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,28 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.amazonaws</groupId>
+        <artifactId>aws-java-sdk-bom</artifactId>
+        <version>1.11.67</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>amazon-kinesis-producer</artifactId>
+      <version>0.12.3</version>
+    </dependency>
     <dependency>
       <groupId>mysql</groupId>
       <artifactId>mysql-connector-java</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -14,22 +14,11 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>com.amazonaws</groupId>
-        <artifactId>aws-java-sdk-bom</artifactId>
-        <version>1.11.67</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
   <dependencies>
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-core</artifactId>
+      <version>1.11.68</version>
     </dependency>
     <dependency>
       <groupId>com.amazonaws</groupId>
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -98,11 +98,15 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts( "producer", "producer type: stdout|file|kafka|kinesis" ).withRequiredArg();
 		parser.accepts( "output_file", "output file for 'file' producer" ).withRequiredArg();
-		parser.accepts( "kafka.bootstrap.servers", "at least one kafka server, formatted as HOST:PORT[,HOST:PORT]" ).withRequiredArg();
-		parser.accepts( "kafka_partition_by", "database|table|primary_key|column, kafka producer assigns partition by hashing the specified parameter").withRequiredArg();
-		parser.accepts( "kafka_partition_columns", "comma separated list of columns, the columns that should be used for partitioning when kafka_partition_by=column").withRequiredArg();
-		parser.accepts( "kafka_partition_by_fallback", "database|table|primary_key, kafka_partition_by fallback when the using 'column' partitioning and the columsn are not present in the row").withRequiredArg();
 
+		parser.accepts( "producer_partition_by", "database|table|primary_key|column, kafka/kinesis producers will partition by this value").withRequiredArg();
+		parser.accepts( "producer_partition_columns", "with producer_partition_by=column, partition by the value of these columns.  comma separated.");
+		parser.accepts( "producer_partition_by_fallback", "database|table|primary_key, fallback to this value when when sing 'column' partitioning and the columns are not present in the row").withRequiredArg();
+
+		parser.accepts( "kafka_partition_by", "[deprecated]").withRequiredArg();
+		parser.accepts( "kafka_partition_columns", "[deprecated]").withRequiredArg();
+		parser.accepts( "kafka_partition_by_fallback", "[deprecated]").withRequiredArg();
+		parser.accepts( "kafka.bootstrap.servers", "at least one kafka server, formatted as HOST:PORT[,HOST:PORT]" ).withRequiredArg();
 		parser.accepts( "kafka_partition_hash", "default|murmur3, hash function for partitioning").withRequiredArg();
 		parser.accepts( "kafka_topic", "optionally provide a topic name to push to. default: maxwell").withOptionalArg();
 		parser.accepts( "kafka_key_format", "how to format the kafka key; array|hash").withOptionalArg();
@@ -153,7 +157,10 @@ public class MaxwellConfig extends AbstractConfig {
 			public String format(Map<String, ? extends OptionDescriptor> options) {
 				this.addRows(options.values());
 				String output = this.formattedHelpOutput();
-				return output.replaceAll("--__separator_.*", "");
+				output = output.replaceAll("--__separator_.*", "");
+
+				Pattern deprecated = Pattern.compile("^.*\\[deprecated\\].*\\n", Pattern.MULTILINE);
+				return deprecated.matcher(output).replaceAll("");
 			}
 		};
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -277,7 +277,7 @@ public class MaxwellConfig extends AbstractConfig {
 		this.producerPartitionColumns = fetchOption("producer_partition_columns", options, properties, null);
 		this.producerPartitionFallback = fetchOption("producer_partition_by_fallback", options, properties, null);
 
-		if(!this.kafkaPartitionKey.equals("database")) {
+		if(this.kafkaPartitionKey != null && !this.kafkaPartitionKey.equals("database")) {
 			LOGGER.warn("kafka_partition_by is deprecated, please use producer_partition_by");
 			this.producerPartitionKey = this.kafkaPartitionKey;
 		}
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -272,24 +272,26 @@ public class MaxwellConfig extends AbstractConfig {
 			}
 		}
 
-		if(this.kafkaPartitionKey != null) {
+		if(this.kafkaPartitionKey != "database") {
 			LOGGER.warn("kafka_partition_by is deprecated, please use producer_partition_by");
 			this.producerPartitionKey = this.kafkaPartitionKey;
+		} else {
+			this.producerPartitionKey = fetchOption("producer_partition_by", options, properties, "database");
 		}
 
 		if(this.kafkaPartitionColumns != null) {
 			LOGGER.warn("kafka_partition_columns is deprecated, please use producer_partition_columns");
 			this.producerPartitionColumns = this.kafkaPartitionColumns;
+		} else {
+			this.producerPartitionColumns = fetchOption("producer_partition_columns", options, properties, null);
 		}
 
 		if(this.kafkaPartitionFallback != null) {
 			LOGGER.warn("kafka_partition_by_fallback is deprecated, please use producer_partition_by_fallback");
 			this.producerPartitionFallback = this.kafkaPartitionFallback;
+		} else {
+			this.producerPartitionFallback = fetchOption("producer_partition_by_fallback", options, properties, null);
 		}
-
-		this.producerPartitionKey		= fetchOption("producer_partition_by", options, properties, "database");
-		this.producerPartitionColumns	= fetchOption("producer_partition_columns", options, properties, null);
-		this.producerPartitionFallback	= fetchOption("producer_partition_by_fallback", options, properties, null);
 
 		this.kinesisStream   = fetchOption("kinesis_stream", options, properties, null);
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -273,25 +273,23 @@ public class MaxwellConfig extends AbstractConfig {
 			}
 		}
 
+		this.producerPartitionKey = fetchOption("producer_partition_by", options, properties, "database");
+		this.producerPartitionColumns = fetchOption("producer_partition_columns", options, properties, null);
+		this.producerPartitionFallback = fetchOption("producer_partition_by_fallback", options, properties, null);
+
 		if(!this.kafkaPartitionKey.equals("database")) {
 			LOGGER.warn("kafka_partition_by is deprecated, please use producer_partition_by");
 			this.producerPartitionKey = this.kafkaPartitionKey;
-		} else {
-			this.producerPartitionKey = fetchOption("producer_partition_by", options, properties, "database");
 		}
 
 		if(this.kafkaPartitionColumns != null) {
 			LOGGER.warn("kafka_partition_columns is deprecated, please use producer_partition_columns");
 			this.producerPartitionColumns = this.kafkaPartitionColumns;
-		} else {
-			this.producerPartitionColumns = fetchOption("producer_partition_columns", options, properties, null);
 		}
 
 		if(this.kafkaPartitionFallback != null) {
 			LOGGER.warn("kafka_partition_by_fallback is deprecated, please use producer_partition_by_fallback");
 			this.producerPartitionFallback = this.kafkaPartitionFallback;
-		} else {
-			this.producerPartitionFallback = fetchOption("producer_partition_by_fallback", options, properties, null);
 		}
 
 		this.kinesisStream  = fetchOption("kinesis_stream", options, properties, null);
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -43,6 +43,7 @@ public class MaxwellConfig extends AbstractConfig {
 	public String producerPartitionFallback;
 
 	public String kinesisStream;
+	public boolean kinesisHashKey;
 
 	public String outputFile;
 	public MaxwellOutputConfig outputConfig;
@@ -293,7 +294,8 @@ public class MaxwellConfig extends AbstractConfig {
 			this.producerPartitionFallback = fetchOption("producer_partition_by_fallback", options, properties, null);
 		}
 
-		this.kinesisStream   = fetchOption("kinesis_stream", options, properties, null);
+		this.kinesisStream  = fetchOption("kinesis_stream", options, properties, null);
+		this.kinesisHashKey = fetchBooleanOption("kinesis_hash_key", options, properties, false);
 
 		this.outputFile         = fetchOption("output_file", options, properties, null);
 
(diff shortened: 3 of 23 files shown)


osheroff:
code looks good, thanks for the PR.  I need to see what's wrong with my dumb-ass maven-fetcher script (why the builds are failing) and then I can merge this.  Do you have a desire to run this in an integration env before I merge?  I'm happy to merge this with a note that it's alpha-quality code.

pom.xml
```diff
@@ -14,8 +14,29 @@
...
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.amazonaws</groupId>
+        <artifactId>aws-java-sdk-bom</artifactId>
+        <version>1.11.67</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
```
osheroff: let's not do this.  it breaks my `make` style of building, and I don't ultimately think it wins anything over just doing straight maven deps.  remove this "import" dep and specficy the `1.11.67` version in sdk-core.   

Errrr.  Are you actually using `sdk-core`?  It doesn't seem to be actually listed as a dep for the kinesis jar.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		this.kinesisStream = kinesisStream;
+
+		this.kinesisProducer = new KinesisProducer();
+	}
+
+	@Override
+	public void push(RowMap r) throws Exception {
+		// why do we need different key formats
```
thomasdziedzic: Thanks for clarifying, switched to using array based only.

thomasdziedzic:
> Do you have a desire to run this in an integration env before I merge?

Please don't merge yet, I want to do some testing before I'm comfortable getting this merged!

thomasdziedzic:
Looks like some of the tests were flaky. Removing an unused import caused the tests to go from green to red.

osheroff:
yeah, the recovery tests are flaky.  have a branch that starts to stabilize them, a bit.

thomasdziedzic:
Added some more configuration options for kinesis.
I managed to do some testing against kinesis today and was happy with the results.

One question I have for you that I still don't fully understand; Why do you need to override the writePosition function in the KafkaProducer?
Link: https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java#L158-L164
Snippet:
```
	@Override
	public void writePosition(BinlogPosition p) throws SQLException {
		// ensure that we don't prematurely advance the binlog pointer.
		inflightMessages.addMessage(p);
		inflightMessages.completeMessage(p);
	}
```

And should this be applied for kinesis also?

thomasdziedzic:
Another question I have, did you want some documentation for this code added somewhere?

osheroff:
just merged a PR that should stabilize the RecoveryTest

osheroff:
> One question I have for you that I still don't fully understand; Why do you need to override the writePosition function in the KafkaProducer?

ah, this is actually a bit of cruft that can be cleaned up.  Before we started pushing DDL statements into the producer, they would need to update the binlog position via a means other than `.push`.  For an async producer, that meant we'd need to move the position forward as soon as all the updates behind it were complete.

Now that we push DDL statements through the producer, though, the manual `.writePosition` codepath can actually go away entirely -- its only call is in `AbstractReplicator#processQueryEvent`, and that call can go away.

Could you do that code change as part of this perhaps?  It helps with my code-review process (sigh) if I can sign off on your changes, otherwise I can to get someone around here to sign-off.

osheroff:
> Another question I have, did you want some documentation for this code added somewhere?

Sure, what-say you change the "kafka" section in the docs to a "producers" section containing kafka & kinesis sub-sections? I can then flesh out a `stdout` and `file` section

thomasdziedzic:
> Now that we push DDL statements through the producer, though, the manual .writePosition codepath can actually go away entirely -- its only call is in AbstractReplicator#processQueryEvent, and that call can go away.

Removed the calls to writePosition.

> Sure, what-say you change the "kafka" section in the docs to a "producers" section containing kafka & kinesis sub-sections?

Added, please let me know what you think. I didn't know how to generate the output website html though so I didn't do a test that I got this correct.

osheroff:
will look.  I'm using the 'mkdocs' python package.  I think I'm going to put together a developer's guide over the winter break.

osheroff:
docs are close enough, I will reformat a bit after merge.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -0,0 +1,131 @@
...
+	private final String kinesisStream;
+
+	public MaxwellKinesisProducer(MaxwellContext context, String kinesisStream) {
+		super(context);
+
+		this.inflightMessages = new InflightMessageList();
+
+		this.kinesisStream = kinesisStream;
+
+		KinesisProducerConfiguration config = KinesisProducerConfiguration.fromPropertiesFile("kinesis-producer-library.properties");
```
osheroff: will this crash if the file doesn't exist?  Is it valid to simply run kinesis without extra configuration?

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -0,0 +1,131 @@
...
+			if(newPosition != null) {
+				context.setPosition(newPosition);
+			}
+
+			return;
+		}
+
+		if(r.isTXCommit()) {
+			inflightMessages.addMessage(r.getPosition());
+		}
```
osheroff: in theory this is all shared logic from the kafka producer, could I guess abstract it into an `AsyncProducer` class  if we wanted to.  but I'm also fine with doing that once the 3rd async-producer comes along.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
thomasdziedzic: You're right, this will crash with:

```15:52:26,534 INFO  KinesisProducerConfiguration - Attempting to load config from file kinesis-producer-library.properties
java.lang.RuntimeException: Error loading config from properties file
	at com.amazonaws.services.kinesis.producer.KinesisProducerConfiguration.fromPropertiesFile(KinesisProducerConfiguration.java:172)
	at com.zendesk.maxwell.producer.MaxwellKinesisProducer.<init>(MaxwellKinesisProducer.java:95)
	at com.zendesk.maxwell.MaxwellContext.getProducer(MaxwellContext.java:224)
	at com.zendesk.maxwell.Maxwell.start(Maxwell.java:151)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:189)
Caused by: java.io.FileNotFoundException: kinesis-producer-library.properties (No such file or directory)
	at java.io.FileInputStream.open(Native Method)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileInputStream.<init>(FileInputStream.java:93)
	at com.amazonaws.services.kinesis.producer.KinesisProducerConfiguration.fromPropertiesFile(KinesisProducerConfiguration.java:169)
	... 4 more```

I will add some checks.

Comment (thomasdziedzic) [inline src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java]: Just pushed a check which should fix this.

**Comment by osheroff** [comment]
took the code out for a spin.  

1.   Wasn't clear from the AWS error messages that I hadn't configured my credentials correct.   Got 
```
	at java.lang.Thread.run(Thread.java:745) [?:1.7.0_121]
Caused by: com.amazonaws.SdkClientException: Unable to load AWS credentials from any provider in the chain
	at com.amazonaws.auth.AWSCredentialsProviderChain.getCredentials(AWSCredentialsProviderChain.java:131) ~[aws-java-sdk-core-1.11.68.jar:?]
	at com.amazonaws.services.kinesis.producer.Daemon.makeSetCredentialsMessage(Daemon.java:548) ~[amazon-kinesis-producer-0.12.3.jar:?]
	at com.amazonaws.services.kinesis.producer.Daemon.startChildProcess(Daemon.java:423) ~[amazon-kinesis-producer-0.12.3.jar:?]
	at com.amazonaws.services.kinesis.producer.Daemon.access$100(Daemon.java:62) ~[amazon-kinesis-producer-0.12.3.jar:?]
	at com.amazonaws.services.kinesis.producer.Daemon$1.run(Daemon.java:132) ~[amazon-kinesis-producer-0.12.3.jar:?]
	... 5 more
```

Which I traced down through some investigation, would be nice to trap this error and point people to a webpage (if possible).

2.  Went through some fuckery with my system clock being way drifted.
3.  In the midst of this, wasn't clear to me whether I should specify `maxwell` or the full `arn:maxwell/horseshit:SHA\whatever` url.  We should be clear in the docs that we mean the short-name.
3.  Got maxwell to write data, though it still spits:

```
06:16:55,136 WARN  LogInputStreamReader - [2016-12-20 06:16:55.136360] [0x00002e57][0x00007f3c82ffd700] [warning] [AWS Log: WARN](AWSErrorMarshaller)Encountered AWSError
AccessDenied
User: arn:aws:iam::261350968068:user/kinesis is not authorized to perform: cloudwatch:PutMetricData:
```

not sure what I need to do there.  I just sort of muddled through a configuration and gave "kinesisFullAccess" or something. 

4.  oh, something is rotten in the state of denmark.  A stream sample:

image

something going wrong with the bytebuffer code perhaps?

**Comment by osheroff** [comment]
no, now it looks fine.  huh.

**Comment by osheroff** [comment]
that SQL analytics thing is hard to get started with.  looks insanely powerful, but lord-a-mercy is it weird to start using.

**Comment by osheroff** [comment]
I think the refactor is... ok.  I can't say I'm super totally thrilled with the code shape, but I don't actually have a better take on it, and it definitely reduces duplication.  One thing to note is that `.toJSON()` is called twice, and that's actually a reasonably expensive bitsy, so we gotta stop doing that; presumably we can have a `RowMap#shouldOutput` or something that say whether it's a normal row or a position-only type row.

**Comment by thomasdziedzic** [comment]
> presumably we can have a RowMap#shouldOutput or something that say whether it's a normal row or a position-only type row.

Yaks all the way down :)

How do I go about implementing this check? Looking at the code, it's actually not obvious to me when `String value = r.toJSON(outputConfig);` would ever equal `null`. I see that `toJSON` creates a `JsonGenerator` instance, populates some fields, and then returns `jsonFromStream`. Then `jsonFromStream` uses the `ByteArrayOutputStream` that was populated by the `JsonGenerator` to output a `String`.

**Comment by osheroff** [comment]
check the subclasses; it's always null for `HeartbeatRowMap`, and sometimes true sometimes false for `DDLRowMap`.  Always true for `RowMap` I think.

**Comment by thomasdziedzic** [comment]
I think I understand, thanks! Will work on this some more tomorrow.

**Comment by alejobs** [comment]
Hi @thomasdziedzic, @osheroff,

I am currently implementing a CDC component and I found the maxwell project. I also want to produce the events to a kinesis stream. I have managed to run Maxwell successfully with the contributions of @thomasdziedzic!

I would like to suggest an improvement: as AWS suggest, records should be inserted into the kinesis stream in batch so that you get a much better performance: http://docs.aws.amazon.com/streams/latest/dev/developing-producers-with-sdk.html#kinesis-using-sdk-java-add-data-to-stream

We could add a worker thread where we send all the events via a shared queue. When the queue reaches a limit or we reach a timeout, we send those events to kinesis. What do you think about it?

**Comment by alejobs** [comment]
Also I have found a bug on the code: the partition key used to send the kinesis events is extracted from the pk json, but in some cases the resulting string is not compatible with kinesis partition key format.

**Comment by thomasdziedzic** [comment]
@alejobs 

> I would like to suggest an improvement: as AWS suggest, records should be inserted into the kinesis stream in batch so that you get a much better performance

You don't have to wait! My code already makes this possible! :)
This is one of the reasons why I decided to use the KPL instead of doing raw api calls to aws.
http://docs.aws.amazon.com/streams/latest/dev/developing-producers-with-kpl.html

> Also I have found a bug on the code: the partition key used to send the kinesis events is extracted from the pk json, but in some cases the resulting string is not compatible with kinesis partition key format.

A reproducible example with a stacktrace would help me here.

**Comment by thomasdziedzic** [comment]
@osheroff Added a `RowMap#shouldOutput` function.

**Comment by osheroff** [comment]
ok, @thomasdziedzic, the changes are mergable as near as I can tell.  Being super clear about what the user needs to do in order to setup an IAM user, give proper permissions, etc (with links to the aws docs in question) would be great, as would figuring out if we could easily trap the "you fucked up the AWS credentials error".  But I think this is good to go for some alpha-quality stuff, so you lemme know when you want it merged.

**Comment by thomasdziedzic** [comment]
@osheroff great I'll create a checklist on what I plan on doing:

- [ ] write clear documentation about what the user needs to do in order to setup an IAM user, give proper permissions, etc (with links to the aws docs in question)
- [ ] trap the "you fucked up the AWS credentials error"

If @alejobs provides details of his issue, I will also look into addressing that.

**Comment by alejobs** [comment]
Hi @thomasdziedzic ,

The problem is when partition key is longer than maximum allowed: http://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html

Thank you for the clarification about the way KPL sends records to kinesis!

**Comment by thomasdziedzic** [comment]
@alejobs according to those docs, the max partitionkey size is 256 chars. 

I use the Array Key method which generates an array with the database, table, and primary keys. To me this seems like it should easily fit under 256 chars.

I'm not sure how I would go about working around this. I guess one could hash the topic so that it fits under 256 chars.

@osheroff what is your opinion on this?
I'm not sure what kafka does in this case.
Does the kafka client prehash the topic before sending it to the server? Is there a max partition key size in kafka?
Should this be a generic feature that gets added as a separate PR if kafka is affected too? If not, this might make sense to make it a kinesis only feature.

**Comment by osheroff** [comment]
@thomasdziedzic what does kinesis use the key for?  If it's like kafka and the key identifies a "unique" message (ie all old messages with the same key as a new message will get compacted), then yeah, a SHA of the key should be fine.  We lose some information but it's better than nothing.

> I'm not sure what kafka does in this case.  Does the kafka client prehash the partition key before sending it to the server? Is there a max partition key size in kafka?

I looked, looks like kafka doesn't really have a max message-size.

**Comment by thomasdziedzic** [comment]
@osheroff re: what does kinesis use the key for?

from: http://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html

> Determines which shard in the stream the data record is assigned to. Partition keys are Unicode strings with a maximum length limit of 256 characters for each key. Amazon Kinesis uses the partition key as input to a hash function that maps the partition key and associated data to a specific shard. Specifically, an MD5 hash function is used to map partition keys to 128-bit integer values and to map associated data records to shards. As a result of this hashing mechanism, all data records with the same partition key map to the same shard within the stream.

**Comment by osheroff** [comment]
oh, i see.  the plot thickens.  So the key that you're sending is not what you want to send -- sending the full kafka-key will basically mean only getting random partitioning.

What needs doing is do extract a partition-value out of the record, as specified by the user, hash that, and send it to kinesis.  In other words, we need to kinesis equivalent of the kafka options `kafka_partition_by`, `kafka_partition_by_columns`, and `kafka_partition_by_fallback`; except that the output of this code would be a value sent along with the record instead of an explicit partition choice.

see https://github.com/zendesk/maxwell/blob/a29e62a804704fd9cfee5c78ddc3fc8d7e270d73/src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java as a starting point.  I think that the code there should be reasonably generic enough.

**Comment by thomasdziedzic** [comment]
I did a quick test and I found the following.

I created a "test" database, with a "pet" table with a "name" column.
I inserted a row with the value "ala".

Here is what I got from my debug:

```
08:42:11,996 DEBUG KinesisCallback - ->  key:["test","pet",[]], shard id:shardId-000000000000, sequence number:49569109671770186845875607311422589622092798012969451522
```

The key does not seem random to me and it looks like this is what I would expect in this case.
So I don't see the reason why we need to add support for a custom partitioner in this initial version.

**Comment by thomasdziedzic** [comment]
re: trapping `com.amazonaws.SdkClientException: Unable to load AWS credentials from any provider in the chain`

I don't know how to trap this error since it looks like it's being thrown by a separate thread.
The error message is good enough™ that I'm not too worried about confusing the user.

I'm tempted to not worry about it for now with this PR unless the answer is obvious.

**Comment by osheroff** [comment]
> I'm tempted to not worry about it for now with this PR unless the answer is obvious.

ok

**Comment by osheroff** [comment]
> So looks like it does make sense to add support for a custom partitioner after all.

Yeah, what we want to send to kinesis is something like `MD5(TABLE_NAME | DATABASE_NAME | PK_VALUES)`

**Comment by thomasdziedzic** [comment]
Do you think I should provide separate config parameter names specific for kinesis partitioning? Or since they would be the same, I should look into making a unified partitioner not specific to kafka and use it for both?

**Comment by thomasdziedzic** [comment]
In addition to my question above, I was reading some of the kafka producer code on how it does the partitioning. I found it interesting that rather than relying on the key to route the record, the code is specifying the partition number manually. It's my understanding that this is already the default behavior [0] when specifying just a key.

Why does the code manually specify the partition to route the message to?

[0] - https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html

**Comment by osheroff** [comment]
1.  I think you could unify the partitioning options if it makes sense.
2.  The key in kafka serves a purpose other than partitioning: it functions as a unique identifier for the row, and kafka will garbage collect all but one copy of rows with the same key.  This means in maxwell's case that the key should always be a unique identifier for the row (we use a tuple of table/database/primary_key), so that you can instruct kafka to only hold on to the last modified copy of the row (more info: https://kafka.apache.org/documentation/#compaction).  So we always want that as a key, but people want to partition their data more flexibly; at zendesk our shards are laid out as identical copies of the schema, so partitioning by database makes sense.  Others want to partition by a primary key, which gets better data distribution but has weaker serialization properties.  In any partitioning scenario, it still makes good sense to write the key as a unique identifier for the row.

**Comment by thomasdziedzic** [comment]
I tried to pull out the common code for the partitioner so that it can be reused for both the kafka and kinesis producers. Please let me know what you think of the approach.

**Comment by osheroff** [comment]
partitioner approach looks good.  I forget, the kinesis lib will take the MD5 of the key before sending it to the server?  Or do we still need to do that to protect against super-long keys?

**Comment by thomasdziedzic** [comment]
I'm thinking that the key hashing should be optional. I still haven't come close to the limit with my testing so I don't have a need for this feature yet. But obviously this issue does exist since it's been encountered.

The reasoning behind wanting to make this optional is that kinesis api exposes the key and the value when you process the record.

When you write a consumer downstream, if you print out the key, it will just be some kind of hash that is hard to extract info from. On the other hand if it's in plain text, you can very clearly see what type of data is stored in the shard.

**Comment by thomasdziedzic** [comment]
Added an option to md5 hash the kinesis key in case you run into any limits with AWS.

I think this completes the PR. Please review and test!

File: config.properties.example [hunk 5]
```diff
@@ -125,6 +109,37 @@ kafka.metadata.fetch.timeout.ms=5000
...
+kinesis_stream=maxwell
+
+# AWS places a 256 unicode character limit on the max key length of a record
+# http://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html
+#
+# Setting this option to true enables hashing the key with the md5 algorithm
+# before we send it to kinesis so all the keys work within the key size limit.
+# Values: true, false
+# Default: false
+#kinesis_hash_key=true
```
osheroff: let's call this `kinesis_md5_keys`.  `kinesis_hash_key` made me think there as another something to configure before I read the text.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -263,6 +273,30 @@ private void setup(OptionSet options, Properties properties) {
 			}
 		}
 
+		if(this.kafkaPartitionKey != "database") {
```
osheroff: use `.equals` here please

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -263,6 +273,30 @@ private void setup(OptionSet options, Properties properties) {
...
+			LOGGER.warn("kafka_partition_by is deprecated, please use producer_partition_by");
+			this.producerPartitionKey = this.kafkaPartitionKey;
+		} else {
+			this.producerPartitionKey = fetchOption("producer_partition_by", options, properties, "database");
```
osheroff: I don't particularly care if the `kafka_` set of options take precedence or the `producer_` ones do, but this might read cleaner if one just clobbered the other, ie:

```
  this.producerPartitionKey = fetchOption("producer_partition_by", options, properties, "database");
  this.producerPartitionColumns = fetchOption("producer_partition_columns", options, properties, null);
  ...
  if(!this.kafkaPartitionKey.equals("database")) {
    LOGGER()
  }
  if(this.kafka...)
```

src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private InflightMessageList inflightMessages;
+
+	public AbstractAsyncProducer(MaxwellContext context) {
+		super(context);
+
+		this.inflightMessages = new InflightMessageList();
+	}
+
+	public abstract void push(RowMap r, CallbackCompleter cc) throws Exception;
```
osheroff: while we're naming stuff, how about this being called `sendAsync()` rather that it being polymorphic?

src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
```diff
@@ -0,0 +1,63 @@
...
+
+	@Override
+	public final void push(RowMap r) throws Exception {
+		if(!r.shouldOutput(outputConfig)) {
+			inflightMessages.addMessage(r.getPosition());
```
osheroff: bonus points for adding a comment explaining what this block is doing

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -21,18 +21,14 @@
...
+	private final AbstractAsyncProducer.CallbackCompleter cc;
 	private final BinlogPosition position;
-	private final boolean isTXCommit;
 	private final String json;
 	private final String key;
 
-	public KafkaCallback(InflightMessageList inflightMessages, BinlogPosition position, boolean isTXCommit, MaxwellContext c, String key, String json) {
-		this.inflightMessages = inflightMessages;
-		this.context = c;
+	public KafkaCallback(AbstractAsyncProducer.CallbackCompleter cc, BinlogPosition position, String key, String json) {
```
osheroff: if `callbackCompleter` became `AsyncProduceRequest` and exposed `key`, `json`, and `position`, that'd reduce some duplication here.   But that can be for another day.

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellPartitioner.java
```diff
@@ -0,0 +1,49 @@
+package com.zendesk.maxwell.producer.partitioners;
+
+import com.zendesk.maxwell.row.RowMap;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class MaxwellPartitioner {
```
osheroff: I guess this should now be `AbstractPartitioner` eh?  sigh..

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -111,7 +111,7 @@ private String pkToJsonHash() throws IOException {
 		return jsonFromStream();
 	}
 
-	private String pkToJsonArray() throws IOException {
+	public String pkToJsonArray() throws IOException {
```
osheroff: this change is no longer needed.

osheroff:
goal posts are in sight.  I can taste them.  They taste like goal posts.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
thomasdziedzic: Ok, I'll leave that job up to another PR.

thomasdziedzic:
I made all the changes you requested. Thanks for the review!

On a side note, I introduced a null pointer exception and the test messages were not helpful at all:

```
testMultipleRowBootstrap(com.zendesk.maxwell.BootstrapIntegrationTest)  Time elapsed: 0.09 sec  <<< ERROR!
java.lang.NullPointerException
	at com.zendesk.maxwell.BootstrapIntegrationTest.testMultipleRowBootstrap(BootstrapIntegrationTest.java:20)
```

The error was in the MaxwellConfig class but the stack trace was not helpful at all. If the error wasn't in such a bad choice and travis didn't run after every commit I pushed, this would have been much harder to debug.

wushujames:
Hi Ben, Thomas,

I thought I'd chime in, since I'm pretty familiar with Kinesis and Kafka and Maxwell.

As you know, when producing the Kinesis, you can specify a partitionKey. The way partitioning in Kinesis is:
* If a partitionKey is specified, then on the AWS backend (not in the Java SDK layer, but within AWS itself) an MD5 is applied to it, which generates a 128-bit number. A Kinesis stream can have multiple shards, which divvy up the entire 128-bit space (with no gaps). The 128-bit number generated by the MD5 is used to choose which shard the message is sent to.
* If you want to, you can specify an explicit 128-bit number yourself. (See http://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html) This is what you would have to do if you want your partitioning to NOT be based on the partitionKey. In the Kinesis Producer Library that this pull request is using, you would do this by using the other version of addUserRecord() (See https://github.com/awslabs/amazon-kinesis-producer/blob/master/java/amazon-kinesis-producer/src/main/java/com/amazonaws/services/kinesis/producer/KinesisProducer.java#L411)

So, with regards to this pull request, if you want to control the partitioning of records so that all records of a certain type go to the same place, you can do any of the following:

* Use whatever you want as the Kinesis partitionKey, and make sure to set the Kinesis ExplicitHashKey. The ExplicitHashKey will be done by doing an MD5 of whatever Maxwell partitioning key you’ve decided to use (database/table/specific-columns). The partitionKey is effectively not used, but it is still accepted and accompanies the Kinesis message as it goes to any consumers.
* Use your Maxwell partitioning key as the Kinesis partitionKey. This will only work if the Maxwell partitioning key is <= 256 characters. AWS will then do its own MD5 and assign it to a shard.
* Do your own MD5 of the Maxwell partitioning key, and use that as the Kinesis partitionKey. AWS will do its own MD5 of your partitionKey (which is itself an MD5), and assign it to a shard.

It looks like this PR uses a combination of the 2nd and 3rd bullets above, but I wanted to point out the 1st bullet (ExplicitHashKey) in case you weren't aware of it. Using bullet 1 would allow you to include the MySQL primary key as the Kinesis primaryKey, just like how it works with Kafka, but still let you do partitioning by database/table/etc. But again, you'd be subject to the 256-character limit that Kinesis has on primaryKey's.

-James

thomasdziedzic:
@wushujames very insightful comment!

This was definitely one of the harder decisions I had to make in this PR but ultimately the reasoning behind my decisions was because I wanted to avoid key truncation. It introduces a couple of issues I can think of e.g. you wouldn't be able to assume the key was valid json and it could introduce the potential of having key conflicts for keys that share a common prefix up to the first 256 characters.

I did find ExplicitHashKey during my investigation on how key hashing worked, but I ultimately decided against it because I would have to use the hash as the key as to avoid the issues mentioned with key truncation.

I don't know how well this method will work in practice since the 256 char limit doesn't affect my work personally.

alejobs:
Hi @thomasdziedzic,

I have reviewed your code again and, after looking at this blog post: https://aws.amazon.com/es/blogs/big-data/implementing-efficient-and-reliable-producers-with-the-amazon-kinesis-producer-library/ (see from "using the kpl onwards)

I would suggest 3 changes:
- Add a while loop for avoiding backpressure: while (kinesis.getOutstandingRecordsCount() > 1e4) {
            Thread.sleep(1);
        }
- Improve onFailure logs with all attempts messages
- Add a mechanism to call some functionalities when stopping replicator: I suggest modifying MaxwellContext "terminate" method and adding a call to a new method on AbstractProducer called "release".

osheroff:
ok @thomasdziedzic I'm gonna go ahead and merge, yeah?

thomasdziedzic:
@osheroff ok

osheroff:
68 comments later the PR is merged.  Thanks @thomasdziedzic for your dedicated work.  This should make a great addition to maxwell.  Look for a 1.7.0 release of maxwell with this (probably still marked alpha?) in the next few days.

thomasdziedzic:
Yes I would definitely label as alpha especially because @alejobs made some good suggestions

PR #506, opened 2016-12-17
Separate heartbeats into their own table

osheroff:
This is something I should have done from the beginning; representing heartbeats as an update to the `positions` table meant we couldn't start heartbeating until we had a binlog position.  This was a real hassle for tests, and ultimately made the code much much less clean.

@zendesk/rules

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/AbstractReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/AbstractReplicator.java
@@ -41,9 +41,10 @@ public abstract class AbstractReplicator extends RunLoopProcess implements Repli
 	 * Possibly convert a RowMap object into a HeartbeatRowMap
 	 *
 	 * Process a rowmap that represents a write to `maxwell`.`heartbeats`.
-	 * If it's a write for a different client_id, we return just the RowMap,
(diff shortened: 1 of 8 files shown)


src/main/java/com/zendesk/maxwell/replication/AbstractReplicator.java
```diff
@@ -40,33 +40,33 @@ public AbstractReplicator(String clientID, AbstractBootstrapper bootstrapper, Po
...
-			Long thisHeartbeat = (Long) heartbeat_at;
-			if ( !thisHeartbeat.equals(lastHeartbeatRead) ) {
-				LOGGER.debug("prcessing heartbeat: " + thisHeartbeat + " @" + row.getPosition());
-				this.lastHeartbeatRead = thisHeartbeat;
-
-				return HeartbeatRowMap.valueOf(row.getDatabase(), row.getPosition(), thisHeartbeat);
-			}
-		}
-		return row;
+		this.lastHeartbeatRead = (Long) row.getData("heartbeat");
```
dadah89: no need for the `!thisHeartbeat.equals(lastHeartbeatRead)` anymore?

PR #509, opened 2016-12-21
Add HTTP POST Producer

namabile:
This PR adds the ability to send Maxwell binlog events via HTTP POST requests to an arbitrary endpoint. Users can optionally sign the request with an HMAC secret for authentication and validation.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -86,7 +86,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts( "__separator_3" );
 
-		parser.accepts( "producer", "producer type: stdout|file|kafka" ).withRequiredArg();
+		parser.accepts( "producer", "producer type: stdout|file|kafka|httpPost" ).withRequiredArg();
 		parser.accepts( "output_file", "output file for 'file' producer" ).withRequiredArg();
 		parser.accepts( "kafka.bootstrap.servers", "at least one kafka server, formatted as HOST:PORT[,HOST:PORT]" ).withRequiredArg();
 		parser.accepts( "kafka_partition_by", "database|table|primary_key|column, kafka producer assigns partition by hashing the specified parameter").withRequiredArg();
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -41,6 +41,8 @@ public class MaxwellConfig extends AbstractConfig {
 	public MaxwellOutputConfig outputConfig;
 	public String log_level;
 
+    public String httpPostEndPoint;
+
 	public String clientID;
 	public Long replicaServerID;
 
@@ -86,7 +88,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts( "__separator_3" );
 
-		parser.accepts( "producer", "producer type: stdout|file|kafka|httpPost" ).withRequiredArg();
+		parser.accepts( "producer", "producer type: stdout|file|kafka|httppost" ).withRequiredArg();
 		parser.accepts( "output_file", "output file for 'file' producer" ).withRequiredArg();
 		parser.accepts( "kafka.bootstrap.servers", "at least one kafka server, formatted as HOST:PORT[,HOST:PORT]" ).withRequiredArg();
 		parser.accepts( "kafka_partition_by", "database|table|primary_key|column, kafka producer assigns partition by hashing the specified parameter").withRequiredArg();
@@ -100,6 +102,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts( "__separator_4" );
 
+		parser.accepts( "httppost_endpoint", "provide an endpoint to send data to using the httppost producer" ).withOptionalArg();
 		parser.accepts( "output_binlog_position", "produced records include binlog position; [true|false]. default: false" ).withOptionalArg();
 		parser.accepts( "output_commit_info", "produced records include commit and xid; [true|false]. default: true" ).withOptionalArg();
 		parser.accepts( "output_nulls", "produced records include fields with NULL values [true|false]. default: true" ).withOptionalArg();
@@ -261,6 +264,8 @@ public class MaxwellConfig extends AbstractConfig {
 
 		this.outputFile         = fetchOption("output_file", options, properties, null);
 
+        this.httpPostEndPoint   = fetchOption("httppost_endpoint", options, properties, null);
+
 		this.includeDatabases   = fetchOption("include_dbs", options, properties, null);
 		this.excludeDatabases   = fetchOption("exclude_dbs", options, properties, null);
 		this.includeTables      = fetchOption("include_tables", options, properties, null);
@@ -349,7 +354,10 @@ public class MaxwellConfig extends AbstractConfig {
 		} else if ( this.producerType.equals("file")
 				&& this.outputFile == null) {
 			usageForOptions("please specify --output_file=FILE to use the file producer", "--producer", "--output_file");
-		}
+		} else if ( this.producerType.equals("httppost")
+                && this.httpPostEndPoint == null)
+			usageForOptions("please specify --httppost_endpoint=URI to use the httppost producer", "--producer", "--httppost_endpoint");
+
 
 		if ( !this.bootstrapperType.equals("async")
 				&& !this.bootstrapperType.equals("sync")
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -42,6 +42,7 @@ public class MaxwellConfig extends AbstractConfig {
 	public String log_level;
 
     public String httpPostEndPoint;
+	public String httpPostHmacSecret;
 
 	public String clientID;
 	public Long replicaServerID;
@@ -102,7 +103,8 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts( "__separator_4" );
 
-		parser.accepts( "httppost_endpoint", "provide an endpoint to send data to using the httppost producer" ).withOptionalArg();
+		parser.accepts( "httppost_endpoint", "provide an endpoint to send data to using the httppost producer" ).withRequiredArg();
+		parser.accepts( "httppost_hmacsecret", "secret key for calculating HMAC digest for httppost producer" ).withRequiredArg();
 		parser.accepts( "output_binlog_position", "produced records include binlog position; [true|false]. default: false" ).withOptionalArg();
 		parser.accepts( "output_commit_info", "produced records include commit and xid; [true|false]. default: true" ).withOptionalArg();
 		parser.accepts( "output_nulls", "produced records include fields with NULL values [true|false]. default: true" ).withOptionalArg();
@@ -265,6 +267,7 @@ public class MaxwellConfig extends AbstractConfig {
 		this.outputFile         = fetchOption("output_file", options, properties, null);
 
         this.httpPostEndPoint   = fetchOption("httppost_endpoint", options, properties, null);
+		this.httpPostHmacSecret = fetchOption("httppost_hmacsecret", options, properties, null);
 
 		this.includeDatabases   = fetchOption("include_dbs", options, properties, null);
 		this.excludeDatabases   = fetchOption("exclude_dbs", options, properties, null);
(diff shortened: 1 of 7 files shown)


osheroff:
cool, thanks for the PR!  The whitespace needs cleaning up, that's a first step.  Maxwell uses exclusively tabs for indentation.

src/main/java/com/zendesk/maxwell/producer/HttpPostProducer.java
```diff
@@ -0,0 +1,115 @@
...
+        this.httpPostHmacSecret = httpPostHmacSecret;
+	}
+
+	@Override
+	public void push(RowMap r) throws Exception {
+
+        Set<String> eventTypeWhiteList = new HashSet<String>(Arrays.asList("insert", "update", "delete"));
+
+        String rowType = r.getRowType();
+        if ( eventTypeWhiteList.contains(rowType) ) {
```
osheroff: this seems odd; what event types are you avoiding here?  #505 will add `RowMap#shouldOutput(outputConfig)`, which might get you what you want.

src/main/java/com/zendesk/maxwell/producer/HttpPostProducer.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import java.io.IOException;
+
+public class HttpPostProducer extends AbstractProducer {
+    static final String HTTP_HMAC_HEADER = "x-signature";
+    static final String HTTP_CONENT_TYPE_HEADER = "application/json";
+    static final String HMAC_ALG = "HmacSHA1";
```
osheroff: the hmac code seems awfully specific to your use case.  Is there any way to make it somehow generic?  i poked briefly around the web to see if anyone had done a signed-request standard, but not how I could see.

schmohlio: how about following spec [here](https://tools.ietf.org/html/draft-cavage-http-signatures-05) and providing HMAC via authorization, e.g., `Authorization: Signature keyId="hmac-key-1",algorithm="hmac-sha1",signature="Base64(HMAC-SHA1(signing string))"`. note optional headers are excluded for body signature verification mode. Also, what if we had `HmacAuthHttpPostProducer` as an example of an `AbstractHttpPostProducer` where users can implement their own headers, etc.

src/main/java/com/zendesk/maxwell/producer/HttpPostProducer.java
schmohlio: we can remove this.

src/main/java/com/zendesk/maxwell/producer/HttpPostProducer.java
osheroff: sure, at least there's a reference implementation (https://github.com/joyent/node-http-signature/blob/master/http_signing.md).  Would you mind contributing whatever signature verification code you end up writing too?  Since all of this is a bit non-standard it's best to have some example code. 

If you want to do it as an abstract class that's fine.  It's on my list somewhere to release maxwell as a .jar so that this kind of functionality can be added as user-supplied producers.

osheroff:
i think favor composition over inheritance here.  Ie there should one producer, ie an `http` producer that has configuration options that enable/disable HMAC, enable/disable exponential backoff, etc.  Otherwise you get to a place where you can't use both hmac+exponential backoff -- and I assume you want both?

schmohlio:
thanks for feedback @osheroff. Hmac inherits the backoff producer, and the thinking here was maybe that inheritance would be easier in the long run with jar refactoring. Nevertheless, I can refactor to a single composition per your suggestion.

Also,

osheroff:
thanks; to my mind inheritance makes a little less sense just due to the fact that people may want a mix and match of various different features.  a simple pure-http class with a couple of supporting classes (backoff, signing, etc) should be cleaner as far as enabling/disabling features.

schmohlio:
@osheroff Updated the PR with composition based on best practices of the Google HTTP client. Includes support for exponential backoff, basic auth, and hmac auth. 

It seems the travis build is broken with the following:
```
javac: invalid flag: target/classes
Usage: javac <options> <source files>
use -help for a list of possible options
make: *** [target/.java] Error 2
```

src/main/java/com/zendesk/maxwell/producer/StdoutProducer.java
```diff
@@ -12,8 +12,11 @@ public StdoutProducer(MaxwellContext context) {
 	public void push(RowMap r) throws Exception {
 		String output = r.toJSON(outputConfig);
 
-		if ( output != null )
-			System.out.println(output);
+		if ( output != null ) {
+            String result = String.format("Table %s, database %s, event type is %s", r.getTable(), r.getDatabase(), r.getRowType());
+			System.out.println(result);
```
osheroff: cruft, plz remove.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -116,6 +126,10 @@ protected OptionParser buildOptionParser() {
 
 		parser.accepts( "__separator_4" );
 
+		parser.accepts( "httppost_endpoint", "provide an endpoint to send data to using the httppost producer" ).withRequiredArg();
+		parser.accepts( "httppost_interceptor", "optional values include 'hmac', 'basicauth'" ).withRequiredArg();
+		parser.accepts( "httppost_authkey", "secret alias (keyId) for calculating HMAC digest or Username for Basic Auth for httppost producer" ).withRequiredArg();
+		parser.accepts( "httppost_authsecret", "secret key for calculating HMAC digest or password for Basic Auth for httppost producer" ).withRequiredArg();
```
osheroff: rename these from `httppost_...` -> `http_...`

osheroff: A little odd that basic auth and HMAC signing share these options.  I think maybe just have 3 options:

- http_hmac_key
- http_hmac_secret
- http_basic_auth

where http_basic_auth is just "username:password"

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -392,9 +411,18 @@ public void validate() {
 		} else if ( this.producerType.equals("file")
 				&& this.outputFile == null) {
 			usageForOptions("please specify --output_file=FILE to use the file producer", "--producer", "--output_file");
-		} else if ( this.producerType.equals("kinesis") && this.kinesisStream == null) {
+
+  	} else if ( this.producerType.equals("kinesis") && this.kinesisStream == null) {
```
osheroff: whitespace issues

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -116,6 +126,10 @@ protected OptionParser buildOptionParser() {
 
 		parser.accepts( "__separator_4" );
 
+		parser.accepts( "httppost_endpoint", "provide an endpoint to send data to using the httppost producer" ).withRequiredArg();
+		parser.accepts( "httppost_interceptor", "optional values include 'hmac', 'basicauth'" ).withRequiredArg();
```
osheroff: just call this `http_authtype`, "interceptor" is weird and doesn't mean much

osheroff:
I'll have to look into the build failures.  my refusal to use maven directly is a constant thorn.  

We'll have to document well the hmac-stuff since it's kinda non-standard, thanks for writing up the example server.

I'll checkout the branch and put the code through some paces later on.

osheroff:
more deeply, I read up on https://en.wikipedia.org/wiki/Digest_access_authentication, and it's... well, it's different, seemingly, than what's implemented here.  as near as I can tell, the ideal behavior would be to simply supply maxwell with a username/password, and have it support the "Digest" authentication in the standard way laid out in the RFC:

1. Make a request, get a 401 back with the server-nonce.
2. sign the POST request with the user/password/nonce/etc, send it back
3. subsequent requests can re-use the server-nonce but issue a new client-nonce.

I guess the bigger question is: what's your authentication use-case like?

schmohlio:
Thanks @osheroff . Basically HMAC was the first method that came to mind for something that provides auth without transport level encryption (although not recommended), and also verifies the _integrity_ of the message (that it has not been altered). 

The `keyId` allows us to potentially keep track of multiple passwords for different applications and to make sure that a receiver of the "webhooks" doesn't accept an altered message, or an unintended message from an authenticated sender that accidentally deployed with the incorrect configuration. 

This article : https://www.wolfe.id.au/2012/10/20/what-is-hmac-authentication-and-why-is-it-useful/ provides some useful context.

It seems that a few other projects that provide "webhook" integrations for change data capture use a variant of HMAC: Shopify and segment.io "webhook" integrations, along with some AWS services.

schmohlio:
@osheroff Let me know what works best. I can switch to the digest access auth if you'd prefer. Or we can just use Basic auth for now.

osheroff:
I definitely can see the logic behind hmac-based stuff; I've implemented stuff that way too, just sort of loosely following what AWS does.

What's a real pity here is that Amazon is not much of a give-back type company and never submitted a spec for this kinda stuff, so we're all left in this "how do I do this in a standard way?" type thing. 

I guess I'd say that doing HMAC based signing is good, but it should be basically "do whatever amazon does," maybe with the header being configurable or something. 

I think it'd also be great to support rfc-standard DIGEST authentication; it doesn't provide any message integrity guarantees but it's sure as hell more secure than basic-auth.

osheroff:
this will fix your build failures, plz cherry-pick https://github.com/zendesk/maxwell/commit/52a1da4d48407a915dc10bdb399e138256c0a2b2

schmohlio:
Here's my suggestion to keep it simple:

We tell users to use HTTPS in production, since TLS guarantees _security_ and _integrity_. Even with DIGEST auth or HMAC over HTTP, the message payload is still exposed and subject to Man-in-middle attacks.

I propose we keep the `HttpPostProducerConfiguration` (which i will rename from `HttpPostProducerInitializer` which allows us to compose backoff strategies, header initialization, and message "interceptors". I will leave the Basic Auth as a an example of using this composition class. Basic Auth will be useful over HTTPS, especially for guarding against accidental deployments into the incorrect environment. 

This composition class will be useful if users want the jar and to use their library to compose their own configurations.

I will also incorporate your changes and use the cherry-pick. Thanks again for overseeing this, I think it will get a lot of usage!

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -96,7 +100,8 @@ protected OptionParser buildOptionParser() {
 
 		parser.accepts( "__separator_3" );
 
-		parser.accepts( "producer", "producer type: stdout|file|kafka|kinesis" ).withRequiredArg();
+		parser.accepts( "producer", "producer type: stdout|file|kafka|kinesis|httpp" ).withRequiredArg();
```
osheroff: typo in 'htpp'

src/main/java/com/zendesk/maxwell/producer/HttpProducer.java
```diff
@@ -0,0 +1,80 @@
...
+    protected HttpRequest lastRequest() { return lastRequest; }
+
+    public String getDateString() {
+        return new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz", Locale.US).format(new Date());
+    }
+
+    public static String digestHeader(String algo, Charset charset, String payload) throws NoSuchAlgorithmException {
+        byte[] digest = MessageDigest.getInstance(algo).digest(payload.getBytes(charset));
+        return String.format("%s=%s", algo, new String(Base64.encodeBase64(digest), charset));
+    }
```
osheroff: cruft at this point, no?

src/main/java/com/zendesk/maxwell/producer/HttpProducer.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public void push(RowMap r) throws Exception {
+        String payload = r.toJSON(outputConfig);
+        ByteArrayContent content = ByteArrayContent.fromString(Json.MEDIA_TYPE, payload);
+        HttpRequest request = requestFactory.buildPostRequest(requestUrl, content);
+
+        HttpHeaders headers = request.getHeaders();
+        headers.setContentLength(content.getLength());
+        headers.setDate(getDateString());
+        headers.set("digest", digestHeader(DEFAULT_DIGEST_ALGO, DEFAULT_CHARSET, payload));
```
osheroff: left over I think

src/main/java/com/zendesk/maxwell/producer/StdoutProducer.java
```diff
@@ -12,8 +12,7 @@ public StdoutProducer(MaxwellContext context) {
 	public void push(RowMap r) throws Exception {
 		String output = r.toJSON(outputConfig);
 
-		if ( output != null )
-			System.out.println(output);
+		System.out.println(output);
```
osheroff: the previous code was correct, I think

osheroff:
prepping for a release, I think this is pretty much ready after merge conflicts?

schmohlio:
yes, merge conflicts are new. will fix by tomorrow.

PR #511, opened 2016-12-21
Minor improvements to Maxwell#start() method

dmytrokoval:
By design, `Maxwell#start()` method is meant to throw any sort of exceptions that may occur up to the caller. Indeed, both `Maxwell#run()` and `Maxwell#main()` surround calls to the `start()` method with try-catch block where the actual logging takes place. With this in mind, I took the liberty of removing `catch (SQLException) {...}` clause (lines 145-148).

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -130,6 +130,7 @@ public class Maxwell implements Runnable {
 	}
 
 	protected void onReplicatorStart() {}
+
 	private void start() throws Exception {
 		try ( Connection connection = this.context.getReplicationConnection();
 			  Connection rawConnection = this.context.getRawMaxwellConnection() ) {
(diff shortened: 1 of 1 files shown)


osheroff:
hi @dmytrokoval, thanks for the PR, but the catch block was there to suppress a strack trace when you get a SQL error, so that the simplest of errors (you misconfigured the permissions) could be easily read without scrolling the terminal.  Maybe move the catch block into `main`?

dmytrokoval:
I don't think that moving the `catch(SQLException e)` block into `main()` will solve the problem completely, as there are yet a few places within the `Maxwell#start()` method that throw `SQLException`. For instance, constructors of `MysqlSchemaStore` and `MaxwellReplicator` classes may throw `SQLException`. To me, it makes no difference what kind of Exception got propagated up to the `main()` method because, in this case, we have no other option but to log the exception and shutdown Maxwell. In addition to my initial commit, may I suggest something like:
```
    public static void main(String[] args) {
		try {
			MaxwellConfig config = new MaxwellConfig(args);
                        ...
			maxwell.start();
		} catch ( Exception e ) {
			LOGGER.error("maxwell encountered unexpected exception. Shutting down gracefully.", e);
			System.exit(1);
		}
    }
```
?

I replaced `e.printStackTrace()` with `LOGGER.error()` to make sure all LogEvents events get recorded by the configured appender.

What do you think @osheroff ?

timbertson:
This diff no longer applies, please rebase + reopen if you're still keen to merge this. FWIW the code has been moved to `main()`, so it won't suppress exceptions for anyone calling `start()` explicitly.

Issue #513, opened 2016-12-23
'Couldn't find table ha_health_check' when used with aliyun RDS


hanguanghong:
We are using maxwell 1.5.2 with aliyun RDS as mysql server. We got the error 'Couldn't find table ha_health_check'. I found a similar issue [https://github.com/zendesk/maxwell/issues/469](url) was fixed, but the fix did not cover my case.
`13:32:39,384 INFO  OpenReplicator - starting replication at mysql-bin.000007:251630
java.lang.RuntimeException: Couldn't find table ha_health_check
        at com.zendesk.maxwell.replication.TableCache.processEvent(TableCache.java:30)
        at com.zendesk.maxwell.replication.MaxwellReplicator.getTransactionRows(MaxwellReplicator.java:298)
        at com.zendesk.maxwell.replication.MaxwellReplicator.getRow(MaxwellReplicator.java:439)
        at com.zendesk.maxwell.replication.MaxwellReplicator.work(MaxwellReplicator.java:159)
        at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:35)
        at com.zendesk.maxwell.Maxwell.start(Maxwell.java:167)
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:188)`
Thanks!

osheroff:
gotcha.  will look.

hanguanghong:
Thanks a lot for the fix!

osheroff:
fixed in 1.6.0

PR #516, opened 2016-12-24
move system-level table blacklist to MaxwellFilter, re-use

osheroff:
This lets us treat mysql.ha_health_check as blacklisted somewhere else.

Addresses #513 

@zendesk/rules


lileeyao:
lgtm 👍

Thanks for saving china e-commerce :)

PR #517, opened 2016-12-27
Binlog connector 3

osheroff:
I think this is the one.  chained pr.

@zendesk/rules


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -0,0 +1,313 @@
...
+
+			switch(event.getEvent().getHeader().getEventType()) {
+				case WRITE_ROWS:
+				case UPDATE_ROWS:
+				case DELETE_ROWS:
+				case EXT_WRITE_ROWS:
+				case EXT_UPDATE_ROWS:
+				case EXT_DELETE_ROWS:
+					Table table = tableCache.getTable(event.getTableID());
+					event.setLastHeartbeat(lastHeartbeatRead);
```
dadah89: I think I get this now, the old replicator would instantiate subclasses of `AbstractRowsEvent` with the `lastHeartbeatRead`, so when building RowMap's it was already present; but now the equivalent `BinlogConnectorEvent` deals with types (e.g. update) itself, so it needs lastHeartbeatRead set, so it can instantiate RowMap's.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -0,0 +1,313 @@
...
+
+		this.binlogEventListener = new BinlogConnectorEventListener(client, queue);
+		this.client.setBlocking(!stopOnEOF);
+		this.client.registerEventListener(binlogEventListener);
+		this.client.setServerId(replicaServerID.intValue());
+		this.client.setBinlogFilename(start.getFile());
+		this.client.setBinlogPosition(start.getOffset());
+
+		/*
+		this.shouldHeartbeat = shouldHeartbeat;
```
osheroff: yeah.  they have the client ping the server instead of having the server ping (heartbeat) the client.  It's ok.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
osheroff: you got it

PR #523, opened 2016-12-29
update binlog position on rds heartbeat inserts

smferguson:
it looks to me like maxwell doesn't update its position on rds heartbeats - on very low-throughput services i never see binlog updates corresponding with anything other than actual data changes, and the heartbeat events that come through aren't processed.
hopefully this corrects that, though i'm not overly fond of what appears to be a delay in the flushing of the position to the db.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
@@ -383,8 +383,8 @@ public class MaxwellReplicator extends AbstractReplicator implements Replicator 
 
 	private void processRDSHeartbeatInsertEvent(QueryEvent event) throws Exception {
 		processRDSHeartbeatInsertEvent(
-				event.getDatabaseName().toString(),
-				eventBinlogPosition(event)
+			event.getDatabaseName().toString(),
+			eventBinlogPosition(event)
 		);
 	}
 
--- a/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
@@ -1,6 +1,8 @@
 package com.zendesk.maxwell.replication;
 
 import java.sql.SQLException;
+import java.util.Objects;
+import java.util.List
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


osheroff:
well, maxwell runs on a 1-second position sync loop.  more is just too much traffic.

src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
```diff
@@ -268,6 +268,12 @@ private RowMapBuffer getTransactionRows() throws Exception {
 						// to us starting on a WRITE_ROWS event -- we sync the schema position somewhere
 						// kinda unsafe.
 						processQueryEvent(qe);
+					} else if (sql.toUpperCase().startsWith("INSERT INTO MYSQL.RDS_HEARTBEAT")) {
+						// RDS heartbeat events take the following form:
+						// INSERT INTO mysql.rds_heartbeat2(id, value) values (1,1483041015005) ON DUPLICATE KEY UPDATE value = 1483041015005
+						// As a result they are processed as query events.
+						// When these occur we need to update to update our position.
+						processRDSHeartbeatEvent(qe);
```
osheroff: I'd just remove all the indirection here and process this inside `processQueryEvent`, unless there's a reason not to.

smferguson: I was a little hesitant to do that b/c processQueryEvent() expects only schema changes so i felt like there wasn't a clear way to rewire it to deal with this

smferguson:
@osheroff re "well, maxwell runs on a 1-second position sync loop. more is just too much traffic."
I see that, and it works when running the stdout producer which calls the abstractproducer's writePosition(). that call saves the position on the next sync loop. 
maxwellkafkaproducer overrides the default writePosition() (https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java#L159). when its writePosition() call is run there is no corresponding database position update until either, 1) a "real" (non-heartbeat) event flows through or 2) the process terminates normally.
killing the process skips the onStop() so it fails to write its last know position. in this case when maxwell restarts the binlog it knew about may no longer exist.

osheroff:
@smferguson yeah, the whole `writePosition` codepath is actually going to go away (see #505); it was too cumbersome to get right in all cases.  The thing to do instead is to generate a `RowMap` event and send it through the producer.  It can either have output or not. 

@thomasdziedzic  any chance you could peel off the refactoring bits from #505 for merging?  Looks like you and scott are crashing a bit here.

smferguson:
@osheroff 👍 good to know about the refactor. hopefully this code looks better... nothing like year-end time crunches :/

src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
```diff
@@ -375,6 +381,13 @@ private void processQueryEvent(QueryEvent event) throws Exception {
 		);
 	}
 
+	private void processRDSHeartbeatInsertEvent(QueryEvent event) throws Exception {
+		processRDSHeartbeatInsertEvent(
+				event.getDatabaseName().toString(),
+				eventBinlogPosition(event)
```
osheroff: just one tab here.  After that 👍

osheroff:
merging.  The next release will be a big one including 5.7 support, and a port to shyko's mysql-binlog-connector.  You fine with waiting for that release?  Might be awhile.

smferguson:
@osheroff cool stuff (5.7 & the connector). i've got a nasty deadline (tuesday) so waiting won't work. i can roll my own for our internal use until you issue the next real one.

PR #524, opened 2016-12-30
build under 5.7

osheroff:

PR #527, opened 2016-12-31
JSON data type support

osheroff:
This is chained atop #524.  Provides JSON column support under 5.7.  Only with the binlog-connector library. 

@zendesk/rules


dadah89:
Logic seems fine to me. Can you add some tests? (ColumnDef and RowMap toJson maybe?)

src/main/java/com/zendesk/maxwell/schema/columndef/JsonColumnDef.java
```diff
@@ -0,0 +1,34 @@
...
+		try {
+			String jsonString = JsonBinary.parseAsString((byte[]) value);
+			return new RawJSONString(jsonString);
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	@Override
+	public String toSQL(Object value) {
```
dadah89: Is this not used?

osheroff: it's not, yeah.  It's old, dangling code that I'm somewhat loathe to get rid of.  But maybe i should just get rid of it.

osheroff:
re: tests.  yes.  I swear I had already written them.

osheroff:
ok tests added

Issue #532, opened 2017-01-05
bootstrapping from a specific time/date (event replay)


smferguson:
in cases where maxwell has missed data(!) or where it has been down long enough that the binlog it knew about is gone we would need to replay table(s) from a known point-in-time so that there is zero data loss to the outside world (kafka, in our case). i wonder if we could leverage the bootstrap code to accept a date/time when the replay "starts". i see that it's using a simple select in the sync case (https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java#L181) so it _feels_ like a straightforward addition of a where clause. thoughts?

osheroff:
a where clause option into the bootstrapper sounds like a fine idea.  go for it.  be careful of SQL injection.

smferguson:
I don't see upgrade scripts in maxwell, what's the approved technique for altering a table? (adding a column?)

osheroff:
https://github.com/zendesk/maxwell/blob/c414ad45c94aa1a1c7f384bf7f2ee4781b9fe564/src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java#L154

smferguson:
👍 thanks. for some reason i never look in java for sql.

osheroff:
released in 1.7.1

PR #535, opened 2017-01-10
Issue #532 - add event replay (bootstrapping from a point-in-time)

smferguson:


osheroff:
hmmm, pretty specific to one use case; I could see a use for doing a query on an integer or string or any number of things.  probably best to just allow the user a full `WHERE` clause

smferguson:
👍  will do. i started down that road and talked myself out of it.

smferguson:
musings: do we want this in the bootstrap cli tool?
for some reason i feel lot less easy about putting it in there. there are arguments on both sides, but adding full `where` clause access from the cli makes me feel dirty.

osheroff:
well, mysqldump allows it, and that's essentially what bootstrap is, so... I'd say go for it.  Making people resort to raw SQL to get the feature feels wrong.

README.md
```diff
@@ -10,7 +10,8 @@ include ETL, cache building/expiring, metrics collection, and search indexing.
 
 advanced features:
 
-- Can do `SELECT * from table` (bootstrapping) initial loads of a table
+- Can do `SELECT * from table` (bootstrapping) initial loads of a table.
+  - Optionally can use a `WHERE` clause to load (replay) part of the table, `SELECT * FROM table WHERE _field_name_ >= _field_value_`.
```
osheroff: just put this in the bootstrapping section, not the front page

docs/docs/bootstrapping.md
```diff
@@ -19,6 +20,16 @@ Alternatively you can insert a row in the `maxwell.bootstrap` table to trigger a
...
 mysql> insert into maxwell.bootstrap (database_name, table_name) values ('fooDB', 'barTable');
 ```
+Optionally, you can include a where clause to replay part of the data.
+```
+mysql> insert into maxwell.bootstrap (database_name, table_name, where_clause) values ('fooDB', 'barTable', 'my_date >= \'2017-01-01 11:07:13\'');
+
+```
+or
+```
+mysql> insert into maxwell.bootstrap (database_name, table_name, where_clause) values ('fooDB', 'barTable', 'my_id = 13');
```
osheroff: fix examples to show using the tool instead methinks

src/main/java/com/zendesk/maxwell/bootstrap/MaxwellBootstrapUtility.java
```diff
@@ -151,22 +151,37 @@ private Long getTotalRowCount(Connection connection, Long bootstrapRowID) throws
...
 		resultSet.next();
 		return resultSet.getLong(1);
 	}
 
-	private long insertBootstrapStartRow(Connection connection, String db, String table, Long totalRows) throws SQLException {
+	private long insertBootstrapStartRow(Connection connection, String db, String table, String whereClause, Long totalRows) throws SQLException {
 		LOGGER.info("inserting bootstrap start row");
-		String sql = "insert into `bootstrap` (database_name, table_name, total_rows) values(?, ?, ?)";
+		String sql = null;
+		if ( whereClause != null ) {
```
osheroff: I think you can make the SQL unconditionally include the column, just go ahead and pass `null` to the `where` column on line 179 no?

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -178,21 +184,29 @@ private void ensureTable(String tableName, Database database) {
...
+        }
+
+        return statement.executeQuery(sql);
+    }
+
+    private Statement createBatchStatement(Connection connection) throws SQLException, InterruptedException {
+        Statement statement = connection.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
+        statement.setFetchSize(Integer.MIN_VALUE);
+        return statement;
+    }
```
osheroff: introduced a ton of whitespace changes, plz fix.  tabs only.

osheroff:
needs a test (just a bootstrap integration test will do) that exercises the code paths.

smferguson:
think this is gtg. did i miss anything?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -110,6 +110,9 @@ public static void upgradeSchemaStoreSchema(Connection c) throws SQLException, I
 			performAlter(c, "alter table `bootstrap` modify column inserted_rows bigint unsigned not null default 0");
 		}
 
+		if ( !getTableColumns("bootstrap", c).containsKey("where_clause") ) {
+			performAlter(c, "alter table `bootstrap` add column where_clause varchar(255)");
```
osheroff: maybe expand this to 1024 or so?  255 seems like it could limit some stuff.

src/test/resources/sql/json/bootstrap-multiple-row-with-whereclause
```diff
@@ -0,0 +1,15 @@
+truncate table maxwell.bootstrap;
+truncate table shard_1.minimal;
+
+insert into minimal set account_id = 1, text_field='hello';
+insert into minimal set account_id = 2, text_field='bonjour';
+insert into minimal set account_id = 3, text_field='goeiedag';
+insert into maxwell.bootstrap set database_name = 'shard_1', table_name = 'minimal', where_clause = 'id > 1';
+-> { database: "shard_1", table: "minimal", type: "insert",          data: {id: 1, account_id: 1, text_field: "hello"} }
+-> { database: "shard_1", table: "minimal", type: "insert",          data: {id: 2, account_id: 2, text_field: "bonjour"} }
+-> { database: "shard_1", table: "minimal", type: "insert",          data: {id: 3, account_id: 3, text_field: "goeiedag"} }
```
osheroff: no need to assert on the standard maxwell output.  It's ok if you do though.

osheroff:
good to go after column size expansion

osheroff:
cool, tested locally and works.  even passes the sneaky give-it-a-wrong column test since the frontend tries to calculate the row count and crashes.  👍

smferguson:
nice of that rowcount to be first :)

PR #541, opened 2017-01-19
Adding encryption capability

Bradyta:
Optionally encrypts the data field using input parameters.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,6 +122,11 @@
       <artifactId>lz4</artifactId>
       <version>1.3.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20160212</version>
+    </dependency>
   </dependencies>
 
   <build>
@@ -137,7 +142,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-	<version>3.6.0</version>
         <configuration>
           <source>1.7</source>
           <target>1.7</target>
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -126,7 +126,6 @@ public class MaxwellConfig extends AbstractConfig {
 		parser.accepts("encryption_key", "The 128 bit key used for encryption").withOptionalArg();
 		parser.accepts("secret_key", "The secret key for the AES encryption").withOptionalArg();
 		parser.accepts("use_encryption", "boolean flag to enable encryption").withOptionalArg();
-		parser.accepts( "ddl_kafka_topic", "optionally provide a topic name to push DDL records to. default: kafka_topic").withOptionalArg();
 
 		parser.accepts( "__separator_5" );
 
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -46,7 +46,6 @@ public class MaxwellConfig extends AbstractConfig {
 	public boolean kinesisMd5Keys;
 
 	public String outputFile;
-	//changed by Brady
 	public MaxwellOutputConfig outputConfig;
 	public String log_level;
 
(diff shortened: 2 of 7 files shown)


Bradyta:
Not sure why the tests fail in Travis but not locally..

Bradyta:
So for some reason, the encrypted string is changing once it gets onto travis, but not locally. It's going from 

1ub4DX5FGXqLnDWK6nhrZAY6sbwNvItFnTPzpByMNX4BuIRIgNq7rybo+hztXCdD
to
WXGOoMWluiiRiH4+T4ugzWQn1VPE4edWOXBTFDfFK9D32QEqdILuWSib9xVzHyTH  
  
Please advise.

osheroff:
can't tell you for sure, but it's likely that your method of re-parsing the `data` value after it's already generated is triggering untestable behavior -- there's no guarantee that the ordering of data will stay the same. 

Suggest that you move your encryption to be in the neighborhood of https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/row/RowMap.java#L260 -- so instead of generating the whole JSON blob, when you go to write the `data` part of the json you write to a separate byte-array that you then encrypt, inline.  This means you won't have to re-parse the JSON, and will also mean that the key order will stay stable.

Out of curiousity, is your business requirement that the entire payload be encrypted, or is it just some fields?  Would it be just as good to encrypt the entire payload (not just the data section), or to selectively encrypt fields inside the data payload?  Just strikes me as a bit odd that we're tunneling the hash inside the `data` payload.

Bradyta:
I agree that it should be moved inline, I'm working on that now and hopefully that solves the issue, although we've yet to have issues with that ever on our platform.  
  
In terms of necessity we would only only *need* certain fields encrypted, but they're numerous and having the databases and tables unencrypted (we add in the primary key values into a sorted comma delimited string too) allows us to perform some basic ETL processes before getting into the gnitty gritty. Having the fields encrypted individually would cause performance issues because then we're not doing 1 decryption we're determining what needs it and then doing several.

osheroff:
would the unencrypted info in the kafka key be enough to do your basic filtering?   It should have (database,table,primary-key-values). Just seems like that way we could encrypt the whole payload and push the message as binary data rather than having to do the weird encrypt-inside-json thing

Bradyta:
That's an interesting point, we just don't use the kafka component ourselves, we use the stdout producers to write to file. I moved things around but am still working on figuring out why the encrypted string is different on travis than it is locally.

osheroff:
hmmm, interesting.  what if the stdout producer could option prefix the message with what would be the kafka key?  In general I see a lot of format/producer PRs that are specific to single business use case, and while I love features, I'm generally trying to make them somewhat general.

I've also thought about trying to open up formatting a message with a little DSL or something but it seems overblown.

Bradyta:
I'm having a hard time thinking this one through pros-cons  
If you encrypt the whole thing and prefix it with the kafka key, maybe someone wanted the primary key columns encrypted and you're redundant.  
If you encrypt the whole thing and don't prefix it, maybe someone wants to be able to do certain ETL processes without decrypting.  
If you encrypt the whole thing without kafka and then have to prefix you're being redundant.  
If you encrypt the data field without kafka like I have it, then your data decryption step is messy and you fail the travis tests. :)   
  
I guess in regards to this PR what's important is, what's your preferred solution?

osheroff:
Thanks for thinking of my preferences :)  I think for a first pass my preferred solution is this:

1.  Encrypt the entire payload.  Add a flag to toggle base-64 encoding.
2.  Add a flag to the stdout producer `--stdout_key_prefix=SEPARATOR`, which enables prefixing the message with the kafka key (unencrypted).  Should be able to use `\n` and `\t` here, as well as `:` or whatever. 

That should I think satisfy your use case ok, as well as being generalized enough for most use cases.   The only thing this doesn't support is encrypting the kafka keys, but I think that's ok for now, if someone comes along and *must* have that they can add it.

smferguson:
@Bradyta looks like this might be getting stale. is it still on your radar?

BradyPT:
@osheroff @smferguson  
  
Got some time to revisit this today while updating to 1.8.1, I think the best option would be to only encrypt the data attribute right now as a default if use_encryption is set to true, and then add an additional flag to encrypt the other attributes if desired. I think if you were to encrypt the whole payload and have the option to prefix it you're adding too much redundancy and mess unless you then go in and strip out the non-data attributes.

Bradyta:
@osheroff @smferguson

Got some time to revisit this today while updating to 1.8.1, I think the best option would be to only encrypt the data attribute right now as a default if use_encryption is set to true, and then add an additional flag to encrypt the other attributes if desired. I think if you were to encrypt the whole payload and have the option to prefix it you're adding too much redundancy and mess unless you then go in and strip out the non-data attributes.

PR #542, opened 2017-01-20
don't regex the topic unless needed

osheroff:
This is worth about 6% of main-thread performance.  Something found while profiling. 

@gabetax @prudhvi @zendesk/rules


prudhvi:
Good catch 👍  looks safe to me just codewise

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -97,7 +99,10 @@ private Integer getNumPartitions(String topic) {
 	}
 
 	private String generateTopic(String topic, RowMap r){
```
prudhvi: Also does this method need to take `topic` parameter? Can't it rely on `this.topic`?

osheroff: yeah.   but it doesn't matter.  And in many cases anything you do to make something more of a "pure" function is better.

smferguson:
6%?! ugh. thanks for finding/fixing :)

PR #543, opened 2017-01-22
separate KafkaProducer#push from the main thread

osheroff:
This should be worth roughly 25% of cpu-concurrency, by running
binlog-event-to-RowMap code on a different thread than the
json-marshalling and kafka-data-compression/batching.  We could go
deeper than this by also parceling out the json-marshal calls into a
separate thread, but at that point I think I'd adopt rx-java or some
kind of pipeline framework.

Also note that we're doing fairly dumb stuff with any exceptions that the guts of `#push` might throw, but I *think* they're ok as runtime exceptions.  I should probably validate that they still crash the process in the same way.

@zendesk/rules @gabetax @prudhvi

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -18,7 +18,7 @@ import org.slf4j.LoggerFactory;
 
 import java.sql.SQLException;
 import java.util.Properties;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ArrayBlockingQueue;
 
 
 class KafkaCallback implements Callback {
@@ -57,14 +57,14 @@ class KafkaCallback implements Callback {
 
 
 public class MaxwellKafkaProducer extends AbstractProducer {
-	private final LinkedBlockingQueue<RowMap> queue;
+	private final ArrayBlockingQueue<RowMap> queue;
 	private final MaxwellKafkaProducerWorker worker;
 
 	public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties, String kafkaTopic) {
 		super(context);
-		this.queue = new LinkedBlockingQueue<>(100);
+		this.queue = new ArrayBlockingQueue<>(100);
 		this.worker = new MaxwellKafkaProducerWorker(context, kafkaProperties, kafkaTopic, this.queue);
-		new Thread(this.worker, "maxwell-kafka-producer").start();
+		new Thread(this.worker, "maxwell-kafka-worker").start();
 
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -53,7 +55,26 @@ public void onCompletion(RecordMetadata md, Exception e) {
...
+
+public class MaxwellKafkaProducer extends AbstractProducer {
+	private final LinkedBlockingQueue<RowMap> queue;
+	private final MaxwellKafkaProducerWorker worker;
+
+	public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties, String kafkaTopic) {
+		super(context);
+		this.queue = new LinkedBlockingQueue<>(100);
+		this.worker = new MaxwellKafkaProducerWorker(context, kafkaProperties, kafkaTopic, this.queue);
+		new Thread(this.worker, "maxwell-kafka-producer").start();
```
prudhvi: Should the name of the threa be `maxwell-kafka-producer-worker` instead?

osheroff: sure.  any other comments?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -53,7 +55,26 @@ public void onCompletion(RecordMetadata md, Exception e) {
...
 
-public class MaxwellKafkaProducer extends AbstractAsyncProducer {
+
+public class MaxwellKafkaProducer extends AbstractProducer {
+	private final LinkedBlockingQueue<RowMap> queue;
+	private final MaxwellKafkaProducerWorker worker;
+
+	public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties, String kafkaTopic) {
+		super(context);
+		this.queue = new LinkedBlockingQueue<>(100);
```
prudhvi: Just 🤔  if it's of fixed max size of 100 and if we are sure the queue is mostly blocked on the producer thread(most of the times queue being full), may be `ArrayBlockingQueue` has better performance wrt to upfront memory allocation and no node creation(put) and deletion(dequeu) overhead and GC

prudhvi:
LGTM have a question about queue type selection 👍

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
osheroff: sure.  I can buy that.  I'll switch all the queues to that, according to stackoverflow it's a little less locky-and-gcy

PR #548, opened 2017-01-26
Improve kinesis producer

alejobs:
Hi,

As I commented on the original PR when Kinesis producer was created (see: [https://github.com/zendesk/maxwell/pull/505#issuecomment-270169122](url), I had some improvements:

- Add a backpressure control to avoid exhausting system resources
- Add a mechanism to stop kinesis producer
- Add a retry mechanism for kinesis producer
- Improve kinesis producer logging

Many of these ideas have been taken from this AWS blog post: https://aws.amazon.com/es/blogs/big-data/implementing-efficient-and-reliable-producers-with-the-amazon-kinesis-producer-library/

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
@@ -172,11 +172,17 @@ public class MaxwellKinesisProducer extends AbstractAsyncProducer {
 		String key = this.partitioner.getKinesisKey(r);
 		String value = r.toJSON(outputConfig);
 
-		ByteBuffer encodedValue = ByteBuffer.wrap(value.getBytes("UTF-8"));
+        ByteBuffer encodedValue = ByteBuffer.wrap(value.getBytes("UTF-8"));
 		ListenableFuture<UserRecordResult> future = kinesisProducer.addUserRecord(kinesisStream, key, encodedValue);
 
 		FutureCallback<UserRecordResult> callback = new KinesisCallback(cc, r.getPosition(), key, value, successRecords, errorRecords, kinesisProducer, kinesisStream, 0, maxAttempts, lastInfoLog);
 
 		Futures.addCallback(future, callback);
 	}
+
+	@Override
+    public void release() {
+	    kinesisProducer.flushSync();
+	    kinesisProducer.destroy();
+    }
 }
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
@@ -143,6 +143,7 @@ public class MaxwellKinesisProducer extends AbstractAsyncProducer {
 	private final KinesisProducer kinesisProducer;
 	private final String kinesisStream;
 	private final int maxAttempts;
+	private final int maxBufferedRecords;
 	private AtomicInteger successRecords = new AtomicInteger(0);
 	private AtomicInteger errorRecords = new AtomicInteger(0);
 	private AtomicLong lastInfoLog = new AtomicLong(0L);
@@ -157,6 +158,7 @@ public class MaxwellKinesisProducer extends AbstractAsyncProducer {
 		this.kinesisStream = kinesisStream;
 
 		this.maxAttempts = context.getConfig().kinesisMaxAttempts;
+		this.maxBufferedRecords = context.getConfig().kinesisMaxBufferedRecords;
 
 		Path path = Paths.get("kinesis-producer-library.properties");
 		if(Files.exists(path) && Files.isRegularFile(path)) {
@@ -171,6 +173,11 @@ public class MaxwellKinesisProducer extends AbstractAsyncProducer {
 	public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) throws Exception {
 		String key = this.partitioner.getKinesisKey(r);
 		String value = r.toJSON(outputConfig);
+
+        // Wait for KPL to finish sending events
+        while(kinesisProducer.getOutstandingRecordsCount()> this.maxBufferedRecords) {
+            Thread.sleep(1);
+        }
 
         ByteBuffer encodedValue = ByteBuffer.wrap(value.getBytes("UTF-8"));
 		ListenableFuture<UserRecordResult> future = kinesisProducer.addUserRecord(kinesisStream, key, encodedValue);
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
@@ -1,9 +1,16 @@
 package com.zendesk.maxwell.producer;
 
+import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.FutureCallback;
@@ -21,6 +28,7 @@ import com.amazonaws.services.kinesis.producer.KinesisProducerConfiguration;
 import com.amazonaws.services.kinesis.producer.UserRecordFailedException;
 import com.amazonaws.services.kinesis.producer.UserRecordResult;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -32,11 +40,28 @@ class KinesisCallback implements FutureCallback<UserRecordResult> {
 	private final String json;
 	private final String key;
 
-	public KinesisCallback(AbstractAsyncProducer.CallbackCompleter cc, BinlogPosition position, String key, String json) {
+	private AtomicInteger successRecords;
+	private AtomicInteger errorRecords;
+	private final KinesisProducer kinesisProducer;
+	private final String kinesisStream;
+	private int attempts = 0;
+	private final int maxAttempts;
+	private AtomicLong lastInfoLog;
+
+    private static final Random RANDOM = new Random();
+
+	public KinesisCallback(AbstractAsyncProducer.CallbackCompleter cc, BinlogPosition position, String key, String json, AtomicInteger successRecords, AtomicInteger errorRecords, KinesisProducer kinesisProducer, String kinesisStream, int attempts, int maxAttempts, AtomicLong lastInfoLog) {
 		this.cc = cc;
 		this.position = position;
 		this.key = key;
 		this.json = json;
+		this.successRecords = successRecords;
+		this.errorRecords = errorRecords;
+		this.kinesisProducer = kinesisProducer;
+		this.kinesisStream = kinesisStream;
+		this.attempts = attempts;
+        this.maxAttempts = maxAttempts;
+        this.lastInfoLog = lastInfoLog;
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


osheroff:
would love @thomasdziedzic to have a look-see, but some high-level stuff first:

1.  Whitespace issues.  Maxwell is tabs instead of spaces.
2. The backpressure mechanism (`Thread.sleep(1)`) is a bit primitive.  I recently merged https://github.com/zendesk/maxwell/pull/543, which slots a queue between the main maxwell thread and the kafka producer.  Could this be abstracted and re-used to provide back-pressure for kinesis?
3.  I'd probably just ditch the fancy throwing-dice-to-debug-log; the intention of debug logging is exactly that it spews every last bit of into possible.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -32,82 +31,168 @@
...
+			cc.markCompleted();
+		} else {
+			attempts++;
 
-		logger.error("Exception during put", t);
+			logger.warn(String.format("Record failed to put. Retrying. Attempt: %d ", attempts));
 
-		cc.markCompleted();
+			try {
+				ByteBuffer encodedValue = ByteBuffer.wrap(json.getBytes("UTF-8"));
```
thomasdziedzic: Any way to DRY this up? The code here is very similar to the code in the main handler.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -32,82 +31,168 @@
...
+				ListenableFuture<UserRecordResult> future = kinesisProducer.addUserRecord(kinesisStream, key, encodedValue);
+
+				FutureCallback<UserRecordResult> callback = new KinesisCallback(cc, position, key, json, successRecords, errorRecords, kinesisProducer, kinesisStream, attempts, maxAttempts);
+
+				Futures.addCallback(future, callback);
+			} catch (UnsupportedEncodingException e) {
```
thomasdziedzic: I think you should either add this same exception to the main handler or remove it here to stay consistent.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -32,82 +31,168 @@
...
 }
 
-public class MaxwellKinesisProducer extends AbstractAsyncProducer {
+public class MaxwellKinesisProducer extends AbstractProducer {
+	private final ArrayBlockingQueue<RowMap> queue;
+	private final MaxwellKinesisProducerWorker worker;
+
+	public MaxwellKinesisProducer(MaxwellContext context, String kinesisStream) {
+		super(context);
+		this.queue = new ArrayBlockingQueue<>(100);
```
thomasdziedzic: Does it make sense to unhardcode this `100` and expose a config for the user?

alejobs: @osheroff could you review this please? I see this 100 was already on the kafka producer

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -182,12 +173,22 @@ public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) thro
 			Thread.sleep(1);
 		}
 
-		ByteBuffer encodedValue = ByteBuffer.wrap(value.getBytes("UTF-8"));
-		ListenableFuture<UserRecordResult> future = kinesisProducer.addUserRecord(kinesisStream, key, encodedValue);
+		sendRow(cc, r.getPosition(), key, value, successRecords, errorRecords, kinesisProducer, kinesisStream, 0, maxAttempts);
+	}
 
-		FutureCallback<UserRecordResult> callback = new KinesisCallback(cc, r.getPosition(), key, value, successRecords, errorRecords, kinesisProducer, kinesisStream, 0, maxAttempts);
+	public static void sendRow(AbstractAsyncProducer.CallbackCompleter cc, BinlogPosition position, String key, String json, AtomicInteger successRecords, AtomicInteger errorRecords, KinesisProducer kinesisProducer, String kinesisStream, int attempts, int maxAttempts) {
```
alejobs: Please see the new method. I had to make it static in order to be able to call it both from producer worker and producer callback.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
osheroff: I pulled this number out of thin air; the only requirement is that the other threads can get some work done in paralell, and don't get *too* far ahead.  

Do note that this is, more or less, the same as MaxBufferedRecords.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -32,82 +31,169 @@
...
 	@Override
 	public void onFailure(Throwable t) {
 		logger.error(t.getClass().getSimpleName() + " @ " + position + " -- " + key);
 		logger.error(t.getLocalizedMessage());
 
-		if(t instanceof UserRecordFailedException) {
-			Attempt last = Iterables.getLast(((UserRecordFailedException) t).getResult().getAttempts());
-			logger.error(String.format("Record failed to put - %s : %s", last.getErrorCode(), last.getErrorMessage()));
-		}
+		if(attempts>=maxAttempts) {
```
osheroff: just a sideways-glance question -- does the kinesis library have no internal retry mechanisms?

alejobs: Yes, KPL has a retry mechanism (see: http://docs.aws.amazon.com/streams/latest/dev/kinesis-producer-adv-retries-rate-limiting.html) but from my experience, the only way of guaranteeing an at-least-once semantincs is adding a retry mechanism on your side. Please @thomasdziedzic correct me if I am wrong

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
thomasdziedzic: @alejobs this is a good point since this will mean that the number of retries for a record wont be `maxAttempts` but instead `kplRetries * maxAttempts`

I don't have much opinion on this topic, but since the KPL does include retry logic, I personally wouldn't include it unless I encountered some kind issue and figured that the retry knobs that KPL exposes aren't enough to fix it.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
thomasdziedzic: @alejobs this is a good point since this will mean that the number of retries for a record wont be maxAttempts but instead kplRetries * maxAttempts

I don't have much opinion on this topic, but since the KPL does include retry logic, I personally wouldn't include it unless I encountered some kind issue and figured that the retry knobs that KPL exposes aren't enough to fix it.

osheroff: from https://github.com/awslabs/amazon-kinesis-producer/blob/master/java/amazon-kinesis-producer/src/main/java/com/amazonaws/services/kinesis/producer/KinesisProducerConfiguration.java#L1216:

>  If you do not wish to lose records and prefer to retry indefinitely, set record_ttl to a large value like INT_MAX. This has the potential to cause head-of-line blocking if network issues or throttling occur. You can respond to such situations by using the metrics reporting functions of the KPL. You may also set fail_if_throttled to true to prevent automatic retries in case of throttling.

So let's just document that config setting better in mawell and ditch the custom retry logic eh?

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -32,82 +31,169 @@
...
 
-		ByteBuffer encodedValue = ByteBuffer.wrap(value.getBytes("UTF-8"));
-		ListenableFuture<UserRecordResult> future = kinesisProducer.addUserRecord(kinesisStream, key, encodedValue);
-
-		// release the reference to ease memory pressure
-		if(!KinesisCallback.logger.isDebugEnabled()) {
-			value = null;
+		// Wait for KPL to finish sending events
+		while(kinesisProducer.getOutstandingRecordsCount()> this.maxBufferedRecords) {
+			Thread.sleep(1);
```
osheroff: remind me again what we're doing here and why?  Is this also not handled by properties of the kinesis library?

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
alejobs: Hi @osheroff and @thomasdziedzic : I have been testing thoroughly maxwell with Kinesis in an stressed scenario (bootstraping a table with 200M records) and I can assure you that even setting record_ttl to Integer.INT_MAX I get some failures. But I understand this may be my concrete scenario and you don't want to extend It to your library

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
alejobs: This is for avoiding backpressure when Kinesis can not handle the events at the same speed you produce them. KPL has an internal buffer but if It grows infinitely It can consume your memory or events can reach their ttl.

nddeluca:
@alejobs Thought I'd comment here since we are working on solving some of the same issues.  I implemented a back pressure mechanism in the linked PR that avoids resorting to `Thread.sleep(1)` by using lock conditions instead, allowing for more efficient thread management.

Instead of the maxwell thread checking the condition in a loop, it allows the KPL callback to signal maxwell that it finished processing another record and so maxwell can try again.  It's pretty much like the implementations that classes like `ArrayBlockingQueue` use.  I'd be interested in how it works for your use case, and it should be a more performant under heavy loads.

I'm also curious what errors you are running into with KPL retries? I haven't seen any issues with it yet in my bootstrapping tests, and the KPL property tuning is very good, so I'm thinking maxwell shouldn't implement extra logic too.

Increasing the TTL for bootstraping is a must though, since the default is only `30 seconds` and any delay in processing that causes records to sit beyond that is going to return failures to maxwell.  I'm using a value of 12 hours for my TTL.  I also increase the record buffering time to 1 second (from 100ms), so maybe that increases reliability too.

The only real issue I've run into with KPL so far, is it does assume that your network upload speed is able to handle at least `numberOfShards * 1 MB/s` throughput.  So, if the network attached to the machine running maxwell can't handle that then it's going to fail a lot of message sends and POSTS to cloudmetrics.  Which is pretty much only an issue when testing bootstraping on my laptop and work wifi.

alejobs:
Hi @nddeluca,

Thank you very much for your contribution, as I am dealing with the same scenario as you (bootstraping very large tables with AWS Kinesis) and I really appreciate your comments.

I will try your solution and tell you if it performs ok and raises any exception.

Regarding the KPL retries, I was having some issues specially when the destination stream was not able to ingest at the rate maxwell was emitting. The main error I was getting was:

`ProvisionedThroughputExceededException, Message: Rate exceeded for shard shardId-000000000048 in stream ...`

But some other times I was also getting InternalFailureException. Nevertheless, I still have to test in a more thoroughly way.

nddeluca:
Cool, thanks for the info.  I'll keep an eye out for any InternalFailureExceptions in my testing too.  Also, I think the default settings on KPL are designed to hit the rate limit and go over, so we should be seeing `ProvisionedThroughputExceededException` errors.  It's explained a little bit in the Rate Limiting section at http://docs.aws.amazon.com/streams/latest/dev/kinesis-producer-adv-retries-rate-limiting.html.  

Their advice is to keep the KPL limit at 150% and scale up the number of shards if too many throttles occur.  If you haven't found it yet, https://github.com/awslabs/amazon-kinesis-scaling-utils looks like it'd be great to allow kinesis to scale up during bootstraps and then scale back down for normal traffic to prevent excessive throttling.

PR #549, opened 2017-01-26
Started Dockerfile for building maxwell

r4j4h:
Hey there,

We love the work you have done with maxwell but fretted at the Dockerfile downloading a pre-built binary elsewhere and wanted to see the source build.

I wanted to add some documentation as well but don't want to waste time if you are against this sort of approach, as there's room for improvement:

As it is, this does the work of the normal Dockerfile by default. I like that yours avoids making us take the whole JDK when we only need the JRE at that point. It'd be nice if there was a step that extracted the built artifacts, etc.

You obviously already have a build step for releasing that concerns itself with most of that, but it is opaque as far as I could see so I wanted to ease development to encourage more contributors.

I am thinking that potential docs would describe testing changes out before submitting MRs by trying to build the Dockerfile, or at least providing an environment to cleanly build, without requiring dependencies on their host machine.

```
$> docker build -f Dockerfile.build -t maxwell-local .
$> docker run --rm -ti maxwell-local
$> docker run --rm -ti maxwell-local bin/maxwell --user=foo --password=bar...
```

The image uses `/workspace` as working directory for code when building.
After building it defaults to `/workspace/target/maxwell-<version>/` so it functions similar to the normal.

This means though unfortunately for iterative development `-w /workspace` is needed:

```
$> docker run --rm -ti -v /local/cloned/codebase/:/workspace/ -w /workspace maxwell-test make test
```

(They'd also need to link any mysql containers themselves etc but I'll leave that as an exercise for the user)

Let me know what you think, or feel free to close this if you don't feel it fits in. :)


osheroff:
docker is a, at least at the moment, a fairly secondary concern for my workflow, so I really haven't put much time into it, other than releasing the artifacts when I get the time.  Any improvements welcome, at this point I usually make a release and then totally forget to release the "official" image. 

I'm fine with the approach of just requiring openjdk and doing a `make package` as part of the docker build, esp since that fits into more docker-centric workflows.  Go ahead and have this PR modify the main `Dockerfile`, and I'll accept it.  If you want to uninstall the JDK at the end and leave the JRE, that's fine, if not, fine too.  

re: documentation, I've got a branch somewhere that changes README.md to have more developer-centric docs on github, leaving the rest of it for maxwells-daemon.io.  That'd be the time & place to note different docker setups etc.

r4j4h:
I did as you mentioned, except I did not uninstall the JDK at the end. After looking at the differences between the upstream JDK/JRE there were more differences than I expected and they pulled FROM different upstream images, so for fear I would be polluting the image layers with BOTH, I punted on that for a future MR. :)

Thanks for being so receptive and helping get this through!

Sounds great about the docs, and totally agree that'd be the better time to mention that stuff.

r4j4h:
Awww apparently I broke a build!

It may just be a timeout issue though:

https://travis-ci.org/zendesk/maxwell/jobs/195725655#L265

osheroff:
with this change there is no chance you've broken the build :)  tested locally, works.  👍

PR #550, opened 2017-01-26
add a fix for table-id re-use

osheroff:
if maxwell survives a server restart, or the table_id wraps (unlikely), or is itself down at the time of server-restart, we may get duplicate entries for a particular table_id and totally fail hilariously.  See
https://github.com/zendesk/maxwell/issue

@zendesk/rules cc @nmaquet


osheroff:
master recovery is assumed to happen in a fresh process.  yes, binlog files always rotate on server bounce.

PR #551, opened 2017-01-27
MySQL GTID support

HenryCaiHaiying:
Enable the GTID support through shyiko's binlog connector.

* Added a column gtid_set in positions table for gtid persistence
    * gtid_set takes the form of concatenated string,
	e.g. Server1:1-5, Server2:1-3, Server1:8-10
      this represents all the executed gtids on maxwell side.
      On maxwell startup time, this string will be sending over to MySQL master and MySQL master will send back the new binlog stream continued from this gtid set
    * currently the new columns is defined as varchar(255), it's probably going to be fine for a while:
	* on gtid format, the 1st part is server's uuid string which is 36 chars and the 2nd/3rd part if the transaction id which can be dozen chars
	* shyiko's GtidSet class has the ability of merging adjacent gtid intervals so we not creating longer and longer gtid string, but this string might get longer when MySQL side has too many replication failover with different servers or multi-threading replication creating too many gtid interval ranges;

* From MySQL side, we actually can get both gtid and the old binlog pos, so both information are persisted into positions table in case people want to switch between gtid and non-gtid mode;

* I only added the code to deal with the gtid processing during normal execution path, but not the recovery code since we are not using recovery mode as of now.

* gtid string is added into BinLogPosition as an optional attribute

Please take a look at mysql doc on gtid and gtid_set format:

https://dev.mysql.com/doc/refman/5.6/en/replication-gtids-concepts.html


src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
```diff
@@ -1,34 +1,64 @@
...
 		rs.next();
-		return new BinlogPosition(rs.getInt("Position"), rs.getString("File"));
+		long l = rs.getInt(POSITION_COLUMN);
+		String file = rs.getString(FILE_COLUMN);
+		String gtidStr = null;
+		try {
+			gtidStr = rs.getString(GTID_COLUMN);
+		} catch (SQLException ex) {
+			// gtid column might not exist if gtid mode is not turned on
+			LOGGER.info("Unable to retrieve column " + GTID_COLUMN, ex);
```
osheroff: no need to log here, this is a very normal case if the server doesn't have gtid-support.

src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
```diff
@@ -43,15 +73,27 @@ public Long getHeartbeat() {
...
+	}
+
+	public GtidSet getGtidSet() {
+		return new GtidSet(gtidStr);
+	}
+
 	@Override
 	public String toString() {
-		return "BinlogPosition[" + file + ":" + offset + "]";
+		return "BinlogPosition[" + file + ":" + offset + ":" + gtidStr + "]";
```
osheroff: should only append if gtidStr is not null

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -230,8 +230,9 @@ public String toJSON(MaxwellOutputConfig outputConfig) throws IOException {
 		}
 
 		if ( outputConfig.includesBinlogPosition )
-			g.writeStringField("position", this.nextPosition.getFile() + ":" + this.nextPosition.getOffset());
-
+			g.writeStringField("position",
+				(nextPosition.getGtidStr() != null ? nextPosition.getGtidStr() + ":" : "")
+				+ this.nextPosition.getFile() + ":" + this.nextPosition.getOffset());
```
osheroff: probably either the gtid here *or* the binlog position?   Better yet, move gtid into its own field `gtid_position` or something.

osheroff:
Ok, now the bigger issues:

1.  Would definitely like to see a test suite going around gtid support.  Boot up a onetimeserver with gtid enabled, run some queries, verify that the position is stored correctly, stuff like that. 
2.  For the initial release, I'd like to make this functionality opt-in.  Meaning that to turn on gtid, you have to pass the flag `--gtid=true`.  In the future we can make this flag the default, but I'd like for there to be a way for people to try out gtid mode (especially when the code is still alpha-quality) but to be able to rollback to regular positions if there's problems.
3.  Recovery: this is actually easier (and harder, in terms of code changes) with GTID: there's basically no recovery mechanism.  GTIDs themselves *are* the recovery mechanism.  This means a few things: (a) `--gtid=true` and `--master_recovery=true` should be mutually exclusive.  (b) in GTID mode, we should only ever store the GTID set for the *cluster*.  This means that the primary key of the `positions` table, `(server_id, client_id)` should really only be `(client_id)` if we're in GTID mode.  Maybe this means we should just set `server_id` to NULL or `-1` if we're in GTID mode.  

We'll also have to deal with GTIDs + the schema store -- the schema store assumes that it can fetch the most recent schema by picking the last schema that's lesser than the target position.  I don't think the same mechanism will quite apply to GTIDs.

Given that easy master swaps are really the selling point of GTIDs, I'm not sure if I'd want to release GTID support until we have code and tests that say we can survive a master-change.

4.  Upgrading to GTID.  This is a really really tricky one.  For the first release, I'm ok if our only upgrade path is telling people "turn on GTIDs.  drop the maxwell schema.  start it up fresh."  But long term, we'll need some way of enabling a graceful upgrade from regular position replication to GTID; maybe it'd be something like: "capture the current gtid set and binlog position from the master.  run maxwell in position-based mode until it hits that sync-point, and at that point setup the gtid-set with the captured thing."  might work.

osheroff:
> Recovery and schema store integration with gtid mode, this seems a lot of work. Can that part be done later since we are not really using these features;

Unfortunately I don't think it's something we can defer; my concern is that we'd get into a state where the design of gtid-mode is subtly wrong from a recovery/schema-store perspective, and that it'd end up being more work in the long run.  If we can say "here's the long term plan", but leave gaps in the code, ok, but I don't want to get to state where the system design makes doing recovery infeasible in the future.

osheroff:
> In terms of checkin, I would prefer smaller incremental checkins otherwise the PR would become huge and difficult to integrate back to mainline. Can I address your code comments and deep thoughts 1&2 for the initial checkin?

Incremental PRs are fine, and yeah, if there's a solid way forward I can merge the initial GTID pass without it being fully feature complete.  Will you be able to dedicate the time to do the recovery/schema-store stuff?  I just don't want to get into a state where GTID sits half done for months.

HenryCaiHaiying:
Ben, I looked at the code in recovery and schema store.  Probably would just do what you suggested for recovery: make sure attemptMasterRecovery and gtidMode is not turned on at the same time.

For schema store with gtid, gtid_set also has the concept of containedWith.  I think I can add the gtid column in maxwell.schemas table, in MysqlSavedSchema.findSchema method, I will loop through all rows in schemas table (in schema id descending order) to find the first row whose gtid is contained within the input gtid.  I think the method isContainedWithin in GtidSet class can do this job.  So this logic would be very similar to current one in findSchema() when you compare binlog position.

osheroff:
@HenryCaiHaiying sounds like a good plan.  for server_id, if gtid_mode is on you should probably consider *any* server_id to be valid (not just the server_id of the current server).  that would probably be enough to get this merged.

HenryCaiHaiying:
@osheroff I think I addressed most of the review comments in the latest commit:

    * Address the review comments for gtid mode
        - added schema restore support
        - added handling during failover
        - ignore the server id in position store when gtid_mode is on
    
    * Fixed the failing integration tests.
    * Added more integration tests on failover and schema restore for gtid mode
    * added gtid mode test suite into travis CI

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -99,6 +99,10 @@ public static void upgradeSchemaStoreSchema(Connection c) throws SQLException, I
 			performAlter(c, "alter table `schemas` add column deleted tinyint(1) not null default 0");
 		}
 
+		if ( !getTableColumns("schemas", c).containsKey("gtid_set") ) {
+			performAlter(c, "alter table `schemas` add column gtid_set varchar(255)");
```
osheroff: i think we should lengthen these out as much as possible.  Maybe even use a `TEXT` here?  If there's no reason against it, might as well future-proof....

osheroff:
looks like `testFailOver` is a bit flaky?  These master/slave/etc integration tests are incredibly hard to write solidly.

osheroff:
from a reading of the code this looks mostly mergable, but I'm going to have to set aside some time and get a working GTID setup to run it through some local testing.

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
HenryCaiHaiying: Use 'Text' seems a big jump since it's not that big and TEXT is stored out of line of the row.

MySQL can allow big varchar (up to 64K bytes for the whole row size), do you want to use 4K for now?

In the normal situation, for one server, gtid_set probably takes 30-40 chars.

HenryCaiHaiying:
testFailOver a bit flaky, it finally all passing on 7 test modes.  But it's an important test for gtid mode.

osheroff:
I guess 4096 is ok

HenryCaiHaiying:
OK, pushed the PR update to change column length from 255 to 4096

HenryCaiHaiying:
@osheroff Ben, how was your gtid testing go?  Do you want to merge in this PR first and then incrementally make more improvements?

osheroff:
been slammed-busy, got started spinning up a cluster last night but got interrupted by sleep.  hope to run through it in the next few days.  are you running this in an integration env?

HenryCaiHaiying:
I am running it on my laptop test environment.

osheroff:
got some time today!  works very nice!  Some notes.

1 - upgraded an existing install of maxwell to a GTID-enabled setup; maxwell actually performed very well, and re-captured the schema and position, but we should note this in documentation, that the current story in upgrading from non-gtid to gtid based maxwell means that it will *not* recover from the last given position; this will be important for people who are upgrading and want to maintain consistency.  

Bonus points for providing some chunks of SQL that might help them re-position the server.

2.  I'm not a fan of how the `position` key in the output works under GTID-mode.  Currently when maxwell is run with `--output_binlog_position`, the key changes its meaning.  Instead of the binlog filename/position, suddenly each row is output with the full GTID-based position (not just the position of the row, the full GTID set).  Can you look into fixing this?  Would be ok for v1 if it just kept outputting the binlog file / position for now

HenryCaiHaiying:
@osheroff for GTID upgrade, I haven't done any testing since we don't require it.  Can you send me a readme snippet you want me to add since you did the testing?

For binlog position output, I feel it's OK to output gtid_set instead of binlog file/position since with the gtid enablement, people's mindset will change to think about gtid all the time.  If you don't want to use that output_binlog position parameter, do you want me to introduce a new output config param?

osheroff:
@HenryCaiHaiying don't worry about the snippet then, I'll write it post-merge.

re: `output_binlog_position` -- it's ok to re-use the config flag to enable outputting a GTID positon (although I think I'd still prefer a single position, "XXXX:1" rather than "XXXX:1-4,YYYY:2-8"), but what I don't want is to re-use the same json *field*; people who parse the JSON should be able to expect the `position` field to contain a consistent thing, so if GTID mode is enabled we should output `gtid_position` or just `gtid` instead.

HenryCaiHaiying:
@osheroff for the output gtid position.  Since we are using shyko's BinaryLogClient, he only kept gtidSet there.   To extract the gtid from that set is not trivial, we need to figure out the UUID for the current master and even with that, there can be multiple gtid ranges for the same master.

Or we can try to convince him to store both gtid and gtidSet in BinaryLogClient, but not sure whether he will take that suggestion.

The other option is to parse the GtidEvent ourself and store the gtid somewhere, but I need to co-ordinate the gtid with the rest of the events in the same transaction, it's quite a bit of work.

Is this output config mostly for debugging purpose?  If so, output the gtid set instead of gtid number would make much a difference?

osheroff:
At zendesk we use it to sequences of events; we'll capture a master position from the server while doing sql snapshots and then check the maxwell position against the captured position to determine ordering.

I've worked with Shyko before, he's super responsive and excited to have maxwell as a user so I'm sure it wouldn't be too hard.  I'm also ok just keeping "position" as is for now and re-adding the output feature in a later release.

HenryCaiHaiying:
I can work with Shyko on keeping the GTID in BinaryLogClient.

For the current PR, is it good to merge in?  Or do you still want me to do
something with that output position for gtid mode?

osheroff:
oops I had a comment half-written here but forgot to send it.  For merging, let's just revert the `position` field to always just include the filename/position, and then we'll add a `gtid` field in the output once we've got the stuff settled with shyko.  Sound like a plan?

HenryCaiHaiying:
@osheroff I made the change you requested in RowMap.java for output position, take a look to see if you merge this PR.

HenryCaiHaiying:
@osheroff Since shyko doesn't want me to add the Gtid directly into BinaryLogClient.  I captured that information in BinLogEventListener, and updated RowMap's output to include this gtid instead.

Please take a look.

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -231,9 +231,10 @@ public String toJSON(MaxwellOutputConfig outputConfig) throws IOException {
 
 		if ( outputConfig.includesBinlogPosition )
 			g.writeStringField("position",
-				nextPosition.getGtidStr() != null
-					? nextPosition.getGtidStr()
-					: this.nextPosition.getFile() + ":" + this.nextPosition.getOffset());
+				nextPosition.getGtid() != null
+				? nextPosition.getGtid()
+				: this.nextPosition.getFile() + ":" + this.nextPosition.getOffset());
```
osheroff: ok, but let's still not re-use the `position` field.  If gtid mode is enabled we should be writing two keys, `position` and `gtid`, if not we should just write `position.

HenryCaiHaiying:
OK, I introduced a new flag: output_gtid_position (instead of
output_binlog_position).  Please take a look.

osheroff:
nice work henry.  merging.

HenryCaiHaiying:
It was great working with you too, thanks.

PR #559, opened 2017-02-08
Make Maxwell compatible with MariaDB MaxScale

adaszko:
There are 2 logical changes in the PR, each in its own commit:

 * Add separate, dedicated connection for retrieving database schema during Maxwell startup

This is required when `replication_host` option is set to MaxScale.  MaxScale does not keep a copy of database schema.  The commit introduces a new `schema_{host,port,user,password}` options that are to be configured with a location of a MySQL/MariaDB's node holding the schema.

 * Change default heartbeat interval from 0.5s to 1s

MaxScale is not capable of handling sub-second heartbeat intervals and requesting one, results (due to integer truncation) in heartbeat interval set to 0, which disables heartbeats altogether.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -89,6 +89,7 @@ public class Maxwell implements Runnable {
 				MysqlSchemaStore oldServerSchemaStore = new MysqlSchemaStore(
 					context.getMaxwellConnectionPool(),
 					context.getReplicationConnectionPool(),
+					context.getSchemaConnectionPool(),
 					recoveryInfo.serverID,
 					recoveryInfo.position,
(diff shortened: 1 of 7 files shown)


adaszko:
Hi, do these look acceptable for mainline Maxwell?  The relevant changes on the MaxScale end have just been merged it looks like they will land in the next release.

osheroff:
sorry for the lag, I've been travelling.  will pull off some time to review soon.

adaszko:
Cool, thanks 🙂

osheroff:
needs a test or two to exercise the case where the schema-server is different than the replication server.  Otherwise looks good to merge.

adaszko:
Rebased and added a test case.

osheroff:
looks like test failures were unrelated, kicked the build.

osheroff:
(:+1)

PR #560, opened 2017-02-10
add RowMap getters for `data` and `oldData` maps

abacaphiliac:
add getters for `data` and `oldData` maps. this increases usability of the data model in Java consumer apps. returned maps are copies, protecting the source maps from manipulations on the fetched maps.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowMap.java
@@ -377,4 +377,14 @@ public class RowMap implements Serializable {
 	public boolean shouldOutput(MaxwellOutputConfig outputConfig) {
 		return true;
 	}
+
+	public LinkedHashMap<String, Object> getData()
+	{
+		return new LinkedHashMap<>(data);
+	}
+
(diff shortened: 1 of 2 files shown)


osheroff:
cool 👍.  are you using maxwell as a jar?  mind if I ask what for?

abacaphiliac:
@osheroff bingo. i've got a kafka consumer written in java, and i thought i'd use your data model rather than straight-up copy yours or resort to using JSONObject. now that i think of it, i've got a JsonToRowMap factory method too. i'll port it (as an overloaded constructor?) and the tests into this lib and present you with a merge request for that piece too : )

osheroff:
yeah I've definitely rewritten `RowMap` about 3 times now.  `maxwell.jar` will have quite a few extra deps, but hey, if you're ok picking those up, go for it.

abacaphiliac:
hasn't been a problem yet, but yeah a little heavy for just a data model. have you given any thought to splitting the data model (and maybe the transformers) into a separate jar? would be cleaner for consumers.

PR #561, opened 2017-02-10
add RowMap constructor that accepts JSON.

abacaphiliac:
adds a RowMap constructor that accepts JSONObject. constructor params, xid, data, and old data are set into the model.

i couldn't come up with a sane default for BinlogPosition, so i left it null and i feel awful about that. any suggestions?

i'm also not too keen on the added dependency, but JSONObject is what i was working with in my consumer app. i'd be happy to take a crack at using Jackson's object-mapper but i think i need to learn about mix-ins with their creator feature if i want to avoid adding annotations to the model. again, up for suggestions : )

~~i chose to use a version range for the JSONObject dependency in case you were ok with that addition. the src and tests work with the earliest and latest stable versions of the dependency so i thought there was no reason to be forceful.~~

the integration tests read JSON from a file because i ran into a little issue with JSONObject where null was wrapped in an object.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,6 +122,11 @@
       <artifactId>lz4</artifactId>
       <version>1.3.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20160810</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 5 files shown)


abacaphiliac:
i hang my head in shame. had to remove the dependency range because it broke the build. makes me like the added dependency even less. please tell me i need to figure this out with Jackson instead : P

abacaphiliac:
closing in favor of #562

PR #562, opened 2017-02-10
add Jackson deserializer to convert JSON string to RowMap data model.

abacaphiliac:
direct competitor of https://github.com/zendesk/maxwell/pull/561

this solution uses existing dependencies to perform the transformation.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
@@ -1,0 +1,139 @@
+package com.zendesk.maxwell.row;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.ValueNode;
+import com.zendesk.maxwell.errors.ParseException;
+
+import java.io.IOException;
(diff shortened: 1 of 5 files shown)


osheroff:
tabs instead of spaces please.

src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
```diff
@@ -0,0 +1,120 @@
...
+    JsonNode ts = node.get("ts");
+    JsonNode xid = node.get("xid");
+    JsonNode data = node.get("data");
+    JsonNode oldData = node.get("old");
+
+    RowMap rowMap = new RowMap(
+        type == null ? null : type.asText(),
+        database == null ? null : database.asText(),
+        table == null ? null : table.asText(),
+        ts == null ? null : ts.asLong(),
```
osheroff: these 4 fields should always be there in the JSON.  Rather than passing null and having unexpected things happen down the road, better to throw some kind of parse-exception if they don't exist.

src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
```diff
@@ -0,0 +1,120 @@
...
+        new ArrayList<String>(),
+        null
+    );
+
+    if (xid != null) {
+      rowMap.setXid(xid.asLong());
+    }
+
+    if (data instanceof ObjectNode) {
```
osheroff: parse error here too in an else

osheroff:
other than minor comments  looks good

abacaphiliac:
thanks for the feedback, @osheroff . i will correct all of these things.

abacaphiliac:
@osheroff fixed.

osheroff:
cool 👍

abacaphiliac:
thanks for the merge. know when the new tag will show up in maven central?
https://mvnrepository.com/artifact/com.zendesk/maxwell

osheroff:
oh, right, need to make maven deploy part of my deploy process.  will push.

PR #563, opened 2017-02-11
implement backpressure support for bootstraping (all async producers)

nddeluca:
This is an alternative to the backpressure implementation in #548.

I'm using Maxwell with Kinesis and bootstrapping will push records to KPL faster than it can send them to AWS, resulting in memory consumption issues, slowdown of KPL, the the bootstrap utility giving inaccurate estimates since the synchronous bootstraper reports the rows as inserted before the callbacks complete.

Kinesis only allows 1MB/s per shard ingest and the streaming query during bootstrapping can pull data from mysql at a much faster rate which fills KPL's buffer very quickly.  Kafka most likely can match the throughput of the mysql data link, so I'm guessing that's why something like this PR hasn't been needed yet.  I haven't done any testing with Kafka though, so I'm uncertain the affects this PR will have there (if any).

This PR modifies the `InflightMessageList` to keep track of TXCommit and NonTXCommit (the bootstrap records) separately, and enforces a limit on the number of NonTXCommit messages using lock conditions.  ~~Also, RowMap now has a UUID generated I use for hash keys since bootstrap records don't have binlog positions.~~  I don't think changing the hash key for TXCommits affects behavior.

Edit: I now use an atomicinteger value for the rowid to avoid the overhead of uuid generate and size of the string.  The atomicinteger should be fine since we will never have more than `Integer.MAX_VALUE` rows inflight at one time and it will roll over once it reaches the max value.  Could also use AtomicLong here if we want the rowId's to be unique for a longer period of time.

I avoided using the `kinesisProducer.getOutstandingRecordCount()` because it's producer specific and it's getting the size of a concurrentHashMap underneath which feels like something that shouldn't be called every message (see http://stackoverflow.com/questions/10754675/concurrent-hashmap-size-method-complexity).

TXCommit and NonTXCommit messages have their own locks and do not contend with each other, and I decided to not implement back pressure for TXCommit messages, though it's an easy addition from here . Maybe should be added to improve robustness of maxwell generally?  I figured for normal usage when not bootstraping, the number of kinesis shards should be scaled so it's not as much of an issue.  Also, the current setup should give TXCommit some priority over a bootstrap flood since it will still allow TXCommit messages to pass through on the bootstrap replication thread when we block NonTXCommit's on the async bootstrap thread.

I chose a limit of 10000 inflight messages in order to make sure KPL has enough messages to aggregate properly, though the ideal number may be based off the average record size and number of shards.   Maybe `numberOfInflightMessage = (numberOfShards * 1000000 / averageMessageSizeInBytes)` which would allow KPL to have around 1 second worth of messages in memory to figure out its aggregation stuff.  Maybe be worth exposing as a configuration option?

Thanks for this awesome library and wanted to get this here in front of eyes.  I still need to update/add more unit tests for my additions.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowMap.java
@@ -23,6 +23,7 @@ public class RowMap implements Serializable {
 
 	static final Logger LOGGER = LoggerFactory.getLogger(RowMap.class);
 
+	private final String uuid;
 	private final String rowType;
 	private final String database;
 	private final String table;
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -38,36 +39,39 @@
...
-			};
+				g.setRootValueSeparator(null);
+				return g;
+			}
+		};
 
 	public RowMap(String type, String database, String table, Long timestamp, List<String> pkColumns,
 			BinlogPosition nextPosition) {
+		rowIdCounter.compareAndSet(Integer.MAX_VALUE, 1);
+		this.rowId = rowIdCounter.getAndIncrement();
```
nddeluca: Hmm I'm not sure that this method to prevent negative overflow is thread safe.. think I'll make a AtomicRowId class and move the logic for a atomic positive integer there.

osheroff:
cool, backpressure is nice, but I'm wondering if we can make the code-shape a little better.  What about implementing or stealing a `BoundedLinkedHashMap` implementation which is a fixed-size LinkedHashMap where it blocks when you try to fill it up?

nddeluca:
Good suggestion, I'll work on extracting that out and cleaning things up.

alejobs:
Hi @nddeluca,

Could you review this? If not, I may try doing it.

Thanks.

alejobs:
Hi @nddeluca, @osheroff,

It seems this feature was already solved when adding a backpressure mechanism on InflightList: https://github.com/zendesk/maxwell/pull/745

Can we close this?

Thanks.

nddeluca:
@alejobs Thanks for reminding me about this PR.  This PR implements back pressure support for bootstrapping or messages without a binlog position (non TX messages), while #745 implements backpressure support for binlog/TX messages only.

My use case required back pressure support only for boostraping since maxwell could stream records from mysql much faster than kinesis could accept them (and I didn't want to increase the shard count for bootstraping).

It looks like I could still refactor this PR with @osheroff's feedback and put up another PR? It looks like we could also add a configurable capacity, since I found 1000 (the current limit in the other PR) to be too small for the kinesis producer to give the best performance.  Thoughts?

PR #567, opened 2017-02-13
Drastically improved start up speed

JeroenMinnaert:
At my company, we have a pretty unique use case in which we have about 7000 MySQL databases with each the same 100 tables with on average about 20 columns per table.

The current implementation (maxwell:1.7.2) goes of of memory in a number of ways in this scenario:
- When capturing that data, ResultSet objects are kept open while looping over them. I implemented the capture in such a way that the ResultSets are closed as soon as possible. You can very clearly see that in my code, I try to get the data out as soon as possible before doing any other processing or executing any other queries.  
- Not all PreparedStatements are cached on class level. I improved the capture code by caching all the statements on class level, so both MySQL and Maxwell use less memory and processing.

The current implementation (maxwell:1.7.2) also crashes when trying to store the schema takes too long (more than 30 seconds). I implemented a check that validates if the connection is still valid before trying to store the schema to the maxwell database.

I also improved the full schema restore by using one big query instead of a lot of separate ones. The maxwell tables are well indexed so doing a big inner join takes a lot less time than doing a query per column per table per database. 

I also added more logging to make this slow process more visible to the end user.

With this PR, the initial capture went from not completing or going OOM after 12h to complete after less than an hour. This PR also improves schema restore speeds from about an hour to two minutes. Also memory usage went from > 8GB (after which it crashed) to a stable 4GB. 

Let me know what you think! :)


osheroff:
great work, thanks.

first; tabs instead of spaces, please.  I'm not sure why no damn java editor in the world seems to be able to autodetect the formatting of the file, but maxwell uses tabs.

after that, it looks like your PR has failed out the test suite, so that needs to be fixed (or the actual bugs!).  

after tests pass and whitespace is cleaned I can look deeper.

thanks again for the PR!

JeroenMinnaert:
@osheroff I added an [.editorconfig](http://editorconfig.org/) so tabs will be enforced in multiple editors.

I was trying to run your tests locally but I'm unsuccessful at it. I first build the docker image and run the tests from in there. I keep getting `com.mysql.jdbc.exceptions.jdbc4.CommunicationsException: Communications link failure`. Any idea how to solve this? I really wanna fix my tests

osheroff:
hmmm, there may be unresolved dependencies for running the tests from a docker image; the tests use a script (`onetimeserver`) that downloads and bootstraps a mysql DB.  What (native) platform are you on?

osheroff:
what's the failure you get locally?

JeroenMinnaert:
@osheroff I'm running my Docker on an Ubuntu 16 VM on Mac OSX. Can you take

JeroenMinnaert:
To give you more details, runnning:
`src/test/onetimeserver --mysql-version=5.6 --log-slave-updates --log-bin=master --binlog_format=row --innodb_flush_log_at_trx_commit=0 --server_id=4321 --character-set-server=utf8 --sync_binlog=0 --default-time-zone=+00:00 --verbose`

does not start up a new process in Docker...

JeroenMinnaert:
OK, I'm getting somewhere (I added -debug):

src/test/onetimeserver --mysql-version=5.6 --log-slave-updates --log-bin=master --binlog_format=row --innodb_flush_log_at_trx_commit=0 --server_id=4321 --character-set-server=utf8 --sync_binlog=0 --default-time-zone=+00:00 --verbose -debug
2017/02/17 18:48:52 booting mysql server (version 5.6.26)
[mysqld] 2017-02-17 18:48:53 0 [Warning] TIMESTAMP with implicit DEFAULT value is deprecated. Please use --explicit_defaults_for_timestamp server option (see documentation for more details).
[mysqld] 2017-02-17 18:48:53 0 [Note] /root/.onetimeserver/bin/mysql/5.6.26/bin/mysqld (mysqld 5.6.26-log) starting as process 979 ...
[mysqld] 2017-02-17 18:48:53 979 [Warning] Using pre 5.5 semantics to load error messages from /root/.onetimeserver/bin/mysql/5.6.26/share/english/.
[mysqld] 2017-02-17 18:48:53 979 [Warning] If this is not intended, refer to the documentation for valid usage of --lc-messages-dir and --language parameters.
[mysqld] 2017-02-17 18:48:53 979 [ERROR] Fatal error: Please read "Security" section of the manual to find out how to run mysqld as root!
[mysqld]
[mysqld] 2017-02-17 18:48:53 979 [ERROR] Aborting
[mysqld]
[mysqld] 2017-02-17 18:48:53 979 [Note] Binlog end
[mysqld] 2017-02-17 18:48:53 979 [Note] /root/.onetimeserver/bin/mysql/5.6.26/bin/mysqld: Shutdown complete
[mysqld]
{"extra_args":["--log-slave-updates","--log-bin=master","--binlog_format=row","--innodb_flush_log_at_trx_commit=0","--server_id=4321","--character-set-server=utf8","--sync_binlog=0","--default-time-zone=+00:00","--verbose","-debug"],"mysql_path":"/root/.onetimeserver/tmp/mysql214222201","output":"/tmp/onetimeserver.4ZKWMQ","parent_pid":1,"port":51297,"server_pid":979}

JeroenMinnaert:
@osheroff alrighty, all green!

JeroenMinnaert:
In order to solve the problem I had on Docker, I had to add "--user=root" as a parameter to `onetimeserver` in MysqlIsolatedServer. However, I'm assuming this will not work in Travis or in your own local env. Should we tackle this in this PR?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -414,69 +418,111 @@ private void restoreSchemaMetadata(Connection conn, Long schemaID) throws SQLExc
...
+			int columnLengthInt = rs.getInt("columnLength");
+			String columnEnumValues = rs.getString("columnEnumValues");
+			String columnCharset = rs.getString("columnCharset");
+			String columnType = rs.getString("columnColtype");
+			int columnIsSigned = rs.getInt("columnIsSigned");
+
+			if (currentDatabase == null || !currentDatabase.getName().equals(dbName)) {
+				if (currentDatabase != null) {
+					long end = System.currentTimeMillis();
+					LOGGER.debug("Processed database " + dbName + " after " + (end - dbStart) + "ms");
```
osheroff: thinking the timing info is kinda not-important even for debugging purposes?  Maybe kill it?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -414,69 +418,111 @@ private void restoreSchemaMetadata(Connection conn, Long schemaID) throws SQLExc
...
-		}
+					columnLength
+			);
+			currentTable.addColumn(c);
 
-		if ( pks != null ) {
-			List<String> pkList = Arrays.asList(StringUtils.split(pks, ','));
-			t.setPKList(pkList);
 		}
+		rs.close();
```
osheroff: don't we need to process whatever is left in `currentTable` and `currentDatabase` after the loop exits?  Am I reading the code wrong?

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
```diff
@@ -48,111 +78,145 @@ public Schema capture() throws SQLException {
...
+		int counter = 1;
+		for (Database db : databases) {
+			LOGGER.debug(counter + "/" + size + " Augmenting " + db.getName());
+			long start = System.currentTimeMillis();
+			captureDatabase(db);
+			long end = System.currentTimeMillis();
+			LOGGER.debug(counter + "/" + size + " Augmented " + db.getName() + " after " + (end - start) + "ms");
+			counter++;
+		}
+		LOGGER.debug("Databases Augmented!");
```
osheroff: here too, ditch the timing information, no one will use it.  at some point I'll finish metrics and it can go there

osheroff:
@JeroenMinnaert great!  minor feedback needed.

re: docker; let's find an environment variable and conditionally add `--user=root`

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
JeroenMinnaert: No, everything is processed in the while loop.

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
JeroenMinnaert: done! I changed the language a little to reflect better what is happening

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
JeroenMinnaert: Done, also here, I changed the language to reflect better what is happening

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
osheroff: oh, gotcha, the objects are added as they're seen and get filled out as the loop progresses.  ok.

src/test/java/com/zendesk/maxwell/MysqlIsolatedServer.java
```diff
@@ -32,6 +32,10 @@ public void boot(String xtraParams) throws IOException, SQLException, Interrupte
 		if ( xtraParams == null )
 			xtraParams = "";
 
+		// By default, MySQL doesn't run under root. However, in an environment like Docker, this is a reality.
+		// By checking the HOME env variable for "/root", we can identify whether we are the root user
+		boolean isRoot = System.getenv().get("HOME").equals("/root");
```
osheroff: hmmmm that's a bit shaky.  I guess it's just in the test env so ok, but could we not check for a DOCKER_ something env var?

JeroenMinnaert: I agree it's a little shaky. Another option is to do System.getProperty("user.name").equals("root") here. I added it just now.

I wouldn't add an env variable to Docker just for testing. I though about doing it first but then I felt stronger keeping the Docker image clean (as it's being used for production) and making sure the tests just run regardless of it's environment (there is no security issue here I guess). It also keeps the dependency on the root user local and visible to anyone editing that class.

osheroff:
very nice work.  👍

JeroenMinnaert:
Thanks!! :)

PR #573, opened 2017-02-22
improve enum values extraction

sinceat:
Currently enum values containing commas will lead to a StringIndexOutOfBoundsException


osheroff:
can you write a test exposing the issue?

sinceat:
I've added a test, although I'm not sure why the build didn't kick in.

osheroff:
travis and pretty much the rest of the known internet have been having problems bcs of aws issues today

osheroff:
fails with `make test`.  you probably want to change your test to do `CREATE TABLE IF NOT EXISTS`

osheroff:
hmmm, if we're gonna fix this let's fix it for real.  How about simply serializing them into a JSON array in schema version 4?  that should handle problems with pathological columns that include a `'` character too. 

@vanchi-zendesk I owe you a coffee on this one.

sinceat:
json be it then.

PR #580, opened 2017-03-02
basic metrics

smferguson:
throwing this out there. didn't add much, but both slf4j and jmx work. tbh i'm not sure what other metrics to add or things to wire up. the big one for us is lag. the others are very-nice-to-haves.


pom.xml
```diff
@@ -151,6 +151,11 @@
       <artifactId>lz4</artifactId>
       <version>1.3.0</version>
     </dependency>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
+      <version>3.1.0</version>
+    </dependency>
   </dependencies>
```
HenryCaiHaiying: Do we want to add support for http reporting?
http://metrics.dropwizard.io/3.1.0/getting-started/

<dependency>
    <groupId>io.dropwizard.metrics</groupId>
    <artifactId>metrics-servlets</artifactId>
    <version>${metrics.version}</version>
</dependency>

src/main/java/com/zendesk/maxwell/MaxwellMetrics.java
```diff
@@ -0,0 +1,38 @@
...
+		if (metricsReportingType.equalsIgnoreCase("slf4j")) {
+			final Slf4jReporter reporter = Slf4jReporter.forRegistry(registry)
+					.outputTo(LoggerFactory.getLogger("com.zendesk.maxwell.metrics"))
+					.convertRatesTo(TimeUnit.SECONDS)
+					.convertDurationsTo(TimeUnit.MILLISECONDS)
+					.withLoggingLevel(Slf4jReporter.LoggingLevel.DEBUG)
+					.build();
+			reporter.start(metricsReportingInteval, TimeUnit.SECONDS);
+			LOGGER.info("Slf4j metrics reporter enabled");
+		} else if (metricsReportingType.equalsIgnoreCase("jmx")) {
```
HenryCaiHaiying: Is that possible to start two reporters (e.g. jmx and ganglia) at the same time?

HenryCaiHaiying:
@smferguson this work is also very useful for us.  Hope it can be merged soon.

pom.xml
smferguson: definitely. thanks for the pointer.

src/main/java/com/zendesk/maxwell/MaxwellMetrics.java
smferguson: i was just wondering that. going to test it/refactor a little based on the test results.

HenryCaiHaiying:
As long as you can make jmx work, that would be good for us.

Also maybe adding a counter metrics on # of messages published to kafka, (to pair with the one on bin log side).

smferguson:
@HenryCaiHaiying added kafka produced and failed message counts. also updated the config so you can output slf4j and jmx at the same time. they both work. unless you and/or @osheroff feel strongly i think i'd like to take a staged approach here:
1. in this pr: get basic jmx/slf4j going (hopefully done tho i may have a build problem to deal with monday)
2. in a following pr: wire in a jetty server and add http reporting
i can add things like graphite & ganglia integrations and kinesis metrics, but i'm mildly hesitant to do so b/c we don't use them (i can't eat my own dog food on those).

HenryCaiHaiying:
I think the features in this PR is good enough for us, jmx is also all we need for now.

Do you know why the travis-ci is failing?

smferguson:
@HenryCaiHaiying 👍 not sure why travis is failing yet. i probably won't have time to look until monday. the error doesn't seem to help much:
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

osheroff:
code looks good, needs documentation.

I'd love an extra push here to get this to also expose a simplistic healthcheck (http://metrics.dropwizard.io/3.1.0/manual/healthchecks/), which a lot of people have asked for.

src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -180,6 +181,14 @@ private void start() throws Exception {
 		this.context.start();
 		this.onReplicatorStart();
 		replicator.runLoop();
+
+		// Register the replication lag metric here because there are codepaths that allow for creating multiple
+		// replicators (at least in the tests).
+		// Doing so results in attempting to create multiple metrics with the same name which causes an exception.
+		MaxwellMetrics.registry.register(
+			MetricRegistry.name(Maxwell.class, "replication", "lag"),
```
HenryCaiHaiying: I feel all the metrics name should be defined as constant in MaxwellMetrics class.

smferguson:
@osheroff the addition of 2 dropwiz deps blows up the mvn fetch. any chance you can take a look? (https://github.com/zendesk/maxwell/pull/580/commits/40a5affe2b5ed017355d51e5522629694058d673)

> java -cp target/dependency-antlr/antlr4-runtime-4.5.jar:target/dependency-antlr/antlr-runtime-3.5.2.jar:target/dependency-antlr/ST4-4.0.8.jar:target/dependency-antlr/org.abego.treelayout.core-1.0.1.jar:target/dependency-antlr/antlr4-4.5.jar org.antlr.v4.Tool -package com.zendesk.maxwell.schema.ddl -lib src/main/antlr4/imports -o target/generated-sources src/main/antlr4/com/zendesk/maxwell/schema/ddl/mysql.g4
build/maven_fetcher:59:in `join': can't convert nil into String (TypeError)
	from build/maven_fetcher:59:in `cache_path'
	from build/maven_fetcher:51:in `cachefile'
	from build/maven_fetcher:311:in `get_file_from_cache'
	from build/maven_fetcher:295:in `get_pom'
	from build/maven_fetcher:224:in `block in resolve'
	from /usr/lib/ruby/1.9.1/set.rb:221:in `block in each'
	from /usr/lib/ruby/1.9.1/set.rb:221:in `each_key'
	from /usr/lib/ruby/1.9.1/set.rb:221:in `each'
	from build/maven_fetcher:223:in `map'
	from build/maven_fetcher:223:in `resolve'
	from build/maven_fetcher:241:in `resolve'
	from build/maven_fetcher:269:in `fetch_pom_deps'
	from build/maven_fetcher:358:in `<main>'

osheroff:
shouldn't the groupId be `io.dropwizard.metrics`?

smferguson:
not according to mvn central, https://mvnrepository.com/artifact/io.dropwizard/dropwizard-core/1.0.6
```
<!-- https://mvnrepository.com/artifact/io.dropwizard/dropwizard-core -->
<dependency>
    <groupId>io.dropwizard</groupId>
    <artifactId>dropwizard-core</artifactId>
    <version>1.0.6</version>
</dependency>
```

osheroff:
https://mvnrepository.com/artifact/io.dropwizard.metrics/metrics-core/3.2.0

osheroff:
(way way too old of a version)

smferguson:
👍 thanks. that sums up my week nicely.

smferguson:
I feel like there are unanswered questions around healthchecks
* how do we want to expose them? afaik they can't be reported via the metricregistry (there's a distinct healthcheckregistry) or jmx (https://github.com/dropwizard/metrics/issues/832).
i think the natural answer is http. the admin servlet answers this, but it means rolling jetty into maxwell (or something like it), http://metrics.dropwizard.io/3.2.0/getting-started.html#reporting-via-http, http://metrics.dropwizard.io/3.2.0/getting-started.html#health-checks
* if we go that route, do we want to offer security around the admin pages? (username/password, oauth, ?)

ultimately i'm looking for a sanity check here

osheroff:
http is definitely the way to go for the health checks/metrics, how much bloat does jetty add? 

I think for a v1 we can just let the user choose a port, if they want to do user/password they can slap nginx in front of it.

smferguson:
i don't think jetty will add much, but it's my first foray into wiring it up so we'll see ;)
👍 on keeping it lightweight. 
with a little luck i'll have this buttoned up this weekend.

osheroff:
cool, looking forward to it

smferguson:
this is wired up. http endpoints are /metrics, /healthcheck & /ping. i'm still not thrilled with skipping metrics when using bufferedmaxwell, but i'm not sure how else to deal with it without digging deeper into the tests/rewriting them.
i definitely cheated on the healthcheck. it seems subjective given that maxwell already terminates if it can't connect to kafka/mysql. watching the metrics feels like the right thing to do. i could see allowing configurable thresholds and tying that to the healthcheck in the future, but then it starts to feel like datadog (maybe not a bad thing).

there are intermittent test failures that seem to be timing-related. ugh.

src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -179,6 +182,29 @@ private void start() throws Exception {
...
+
+		// Background: Dropwizard throws an exception if you try to register multiple metrics with the same name.
+		// We register everything here for simplicity. In reality we only need to register the replication.lag metric
+		// here.
+		//
+		// More details:
+		// Register the replication lag metric here because there are codepaths that create multiple replicators
+		// (at least in the tests).
+		// In addition, the tests create multiple BufferedMaxwells so we need to avoid those cases too...
+		if ( !(this instanceof BufferedMaxwell) ) {
```
osheroff: hmmm, what if instead of doing this we instead just prevented ourselves from double-registering metrics, something like:

```
  if ( !MaxwellMetrics.isRegistered(MetricRegistry.name(...)) ) { 

  }
```

smferguson: 👍  yeah. that's a much better idea.

smferguson:
@HenryCaiHaiying when you have a min can you take another look?

src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
```diff
@@ -0,0 +1,19 @@
...
+public class MaxwellHealthCheck extends HealthCheck {
+
+	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellHealthCheck.class);
+
+	public MaxwellHealthCheck() { }
+
+	@Override
+	protected Result check() throws Exception {
+		// TODO: actually return something useful
+		return Result.healthy();
```
HenryCaiHaiying: Maybe when the message consumption rate or production rate dropped significantly, return unhealthy.

src/main/java/com/zendesk/maxwell/metrics/MaxwellServer.java
```diff
@@ -0,0 +1,55 @@
...
+		this.server.start();
+		this.server.join();
+	}
+
+	@Override
+	public void run() {
+		try {
+			startServer();
+		} catch (Exception e) {
+			throw new RuntimeException(e);
```
HenryCaiHaiying: We need an uncaughtExceptionhandler for this kind of exceptions from Runnable.  Otherwise when this thread dies, no one will notice it.  The uncaughtExceptionHandler can just call System.exit

src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
smferguson: it assumes something about the system though.
in our case, we're about to have new alpha-version services running production. these will get very sporadic use for the first few months as we introduce select customers to them and get feedback. i feel like that scenario would trip most (all?) non-configurable thresholds.

smferguson:
@HenryCaiHaiying 1 more look?

HenryCaiHaiying:
Looks fine to me, but Ben has the approval rights.

smferguson:
👍 thanks. just hoping to make it easier on ben.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -166,6 +169,12 @@ protected OptionParser buildOptionParser() {
 
 		parser.accepts( "__separator_8" );
 
+		parser.accepts( "metrics_reporting_type", "how maxwell metrics will be reported, at least one of slf4j|jmx|http" ).withOptionalArg();
+		parser.accepts( "metrics_reporting_interval", "the frequency metrics are emitted to the log, in seconds, when slf4j reporting is configured" ).withOptionalArg();
+		parser.accepts( "metrics_reporting_port", "the port the server will bind to when http reporting is configured" ).withOptionalArg();
```
osheroff: all these args should be required, no?

osheroff: unless I forget the semantics of `withOptionalArg`

src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
```diff
@@ -0,0 +1,29 @@
...
+
+	public MaxwellHealthCheck(MetricRegistry metricRegistry) {
+		this.metricRegistry = metricRegistry;
+	}
+
+	@Override
+	protected Result check() throws Exception {
+		// TODO: this should be configurable.
+		if (this.metricRegistry.getMeters().get(AbstractAsyncProducer.failedMessageMeterName).getFifteenMinuteRate() > 0) {
+			return Result.unhealthy(">1 messages failed to be sent to Kafka in the past 15minutes");
```
osheroff: just say "no events were processed in the last 15 minutes". 

Note that this definitely needs work.

src/main/java/com/zendesk/maxwell/metrics/MaxwellServer.java
```diff
@@ -0,0 +1,64 @@
+package com.zendesk.maxwell.metrics;
```
osheroff: would call this `MaxwellHTTPServer` for clarity

src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
osheroff: oh I see.  you're checking for failures to the producer.  hmmmmmmmmm ok I guess it's a good enough start.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
smferguson: i think it's ok. none are actually required (when nothing's supplied no metrics are output anywhere).

smferguson:
fyi, this still gets occasional failures in the recoverytest. seems timing related. also seems similar to (or exactly like) the failure here: https://travis-ci.org/zendesk/maxwell/jobs/211093207 from pr https://github.com/zendesk/maxwell/pull/597
> Time: 293.663
> There was 1 failure:
> 1) testFailOver(com.zendesk.maxwell.recovery.RecoveryTest)
java.lang.AssertionError: expected:<501> but was:<322>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)

smferguson:
any thoughts about adding datadog as an output type? so [slf4j,jmx,http,datadog]. it would report metrics to datadog on a specified interval.
we'd find it very useful, i just wonder if it's too specific to our use-case.

osheroff:
yes plz on datadog.  we use it too

smferguson:
i know you said you had a torpedo coming your way so i'm definitely not looking to push. just to plan a bit so i can communicate better with stakeholders here. wild guess when you might have cycles to give this another look?

osheroff:
the torpedo hit!  I'll have a bunch of airport time soon and can take a look "real-soon-now".  sorry for all the lag scott.

smferguson:
👍 thanks again. safe travels.

docs/docs/config.md
```diff
@@ -180,6 +180,11 @@ exclude_tables                 | PATTERN                             | ignore up
 blacklist_dbs                  | PATTERN                             | ignore updates AND schema changes from databases (see warnings below) |
 blacklist_tables               | PATTERN                             | ignore updates AND schema changes from tables named like PATTERN (see warnings below) |
 &nbsp;
+**monitoring**
+metrics_reporting_type         | [slf4j &#124; jmx &#124; http]      | how maxwell metrics will be reported, at least one of slf4j &#124; jmx &#124; http |
```
osheroff: missing "datadog" here yeah?

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -486,6 +519,14 @@ public void validate() {
 		} catch (MaxwellInvalidFilterException e) {
 			usage("Invalid filter options: " + e.getLocalizedMessage());
 		}
+
+		if ( this.metricsDatadogType.contains("http") && StringUtils.isEmpty(this.metricsDatadogAPIKey) ) {
+			usageForOptions("please specify metrics_datadog_apikey when metrics_datadog_type = http");
+		}
+
+		if ( this.metricsDatadogType.contains("udp") && StringUtils.isEmpty(this.metricsDatadogHost) ) {
+			usageForOptions("Please specify metrics_datadog_host when metrics_datadog_type = udp" );
```
osheroff: could just default to localhost, pretty common config to run dd-agent on all yr hosts.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -321,7 +344,17 @@ private void setup(OptionSet options, Properties properties) {
 		this.kinesisStream  = fetchOption("kinesis_stream", options, properties, null);
 		this.kinesisMd5Keys = fetchBooleanOption("kinesis_md5_keys", options, properties, false);
 
-		this.outputFile         = fetchOption("output_file", options, properties, null);
+		this.outputFile = fetchOption("output_file", options, properties, null);
+
+		this.metricsReportingType = fetchOption("metrics_type", options, properties, null);
+		this.metricsSlf4jInterval = fetchLongOption("metrics_slf4j_interval", options, properties, 60L);
+		this.metricsHTTPPort = Integer.parseInt(fetchOption("metrics_http_port", options, properties, "8080"));
+		this.metricsDatadogType = fetchOption("metrics_datadog_type", options, properties, "");
```
osheroff: maybe explicitly default to "udp" since that's how the actual code seems to behave.

osheroff:
👍 from me after config nits.  @timbertson @liulikun might want to take a quick look too

smferguson:
fyi, previous test failures where the timing related issue i've seen elsewhere:
>There was 1 failure:
>1) testFailOver(com.zendesk.maxwell.recovery.RecoveryTest)
java.lang.AssertionError: expected:<501> but was:<322>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)

osheroff:
yeah the recovery tests have always been a bit flaky.  very very very hard to get those many-moving-parts integration tests correct.

src/main/java/com/zendesk/maxwell/metrics/MaxwellMetrics.java
```diff
@@ -0,0 +1,111 @@
...
+
+			reporter.start(config.metricsDatadogInterval, TimeUnit.SECONDS);
+			LOGGER.info("Datadog reporting enabled");
+		}
+	}
+
+	private static ArrayList<String> getDatadogTags(String datadogTags) {
+		ArrayList<String> tags = new ArrayList<>();
+		for (String tag : datadogTags.split(",")) {
+			tags.add(tag);
```
timbertson: nit: not sure if datadog cares, but we should skip tag if it's equal to `""`

timbertson:
LGTM (with a tiny nit) 👍

docs/docs/config.md
```diff
@@ -180,6 +180,11 @@ exclude_tables                 | PATTERN                             | ignore up
 blacklist_dbs                  | PATTERN                             | ignore updates AND schema changes from databases (see warnings below) |
 blacklist_tables               | PATTERN                             | ignore updates AND schema changes from tables named like PATTERN (see warnings below) |
 &nbsp;
+**monitoring**
+metrics_reporting_type         | [slf4j &#124; jmx &#124; http &#124; datadog]      | how maxwell metrics will be reported, at least one of slf4j &#124; jmx &#124; http &#124; datadog| 
+metrics_reporting_interval     | INT                                 | the frequency metrics are emitted to the log, in seconds, when slf4j reporting is configured | 
+metrics_reporting_port         | INT                                 | the port the server will bind to when http reporting is configured | 
+&nbsp;
```
liulikun: These config item names are out of date.

docs/docs/monitoring.md
```diff
@@ -0,0 +1,60 @@
+### Monitoring
+***
+Maxwell exposes certain metrics through either its base logging mechanism, JMX, HTTP or by push to Datadog. This is configurable through commandline options
+or the `config.properties` file. These can provide insight into system health.
+At present certain metrics are Kafka-specific - that is, not yet supported other producers.
+
+### Metrics
+All metrics are prepended with `MaxwellMetrics.`
```
liulikun: It'll be great if we can config this prefix, `MaxwellMetrics` can be a default value.

pom.xml
```diff
@@ -151,6 +151,31 @@
...
+    </dependency>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-servlets</artifactId>
+      <version>3.1.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+      <version>9.4.2.v20170220</version>
```
liulikun: We're still running on Java 7, I find this version doesn't compile with Java 7, I have to change it to '9.2.21.v20170120'

src/main/java/com/zendesk/maxwell/metrics/MaxwellHTTPServer.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class MaxwellHTTPServer {
+	public MaxwellHTTPServer(int port, MetricRegistry metricRegistry, HealthCheckRegistry healthCheckRegistry) {
+		MaxwellHTTPServerWorker maxwellHTTPServerWorker = new MaxwellHTTPServerWorker(port, metricRegistry, healthCheckRegistry);
+		Thread thread = new Thread(maxwellHTTPServerWorker);
+
+		thread.setUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
+			public void uncaughtException(Thread t, Throwable e) {
+				e.printStackTrace();
+				System.exit(1);
```
liulikun: I'm wondering if we should exit on errors from monitoring.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -186,6 +186,9 @@ private RowMapBuffer getTransactionRows() throws Exception {
 				case XID:
 					buffer.setXid(event.xidData().getXid());
 
+					// feed metric gauge.
+					replicationLag = event.getEvent().getHeader().getTimestamp() - (System.currentTimeMillis() / 1000);
```
liulikun: The lag is normally quite low. Can we use milliseconds here? So we don't need to `/ 1000`. Otherwise we may find the lag is 0 most of time.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -302,4 +305,8 @@ public Schema getSchema() throws SchemaStoreException {
 	public void setFilter(MaxwellFilter filter) {
 		this.filter = filter;
 	}
+
+	public Long getReplicationLag() {
+		return this.replicationLag;
+	}
```
liulikun: I think we can move this to `AbstractReplicator`

src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
```diff
@@ -283,6 +283,7 @@ private RowMapBuffer getTransactionRows() throws Exception {
 
 					buffer.setXid(xe.getXid());
 
+					replicationLag = (xe.getHeader().getTimestamp() / 1000) - (System.currentTimeMillis() / 1000);
```
liulikun: Use milliseconds here?

src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
```diff
@@ -409,4 +410,7 @@ public OpenReplicator getOpenReplicator() {
 		return replicator;
 	}
 
+	public Long getReplicationLag() {
+		return this.replicationLag;
+	}
```
liulikun: This can be implemented in `AbstractReplicator`?

liulikun:
@smferguson Very nice work! Just a couple of minor issues. Thanks very much.

src/main/java/com/zendesk/maxwell/metrics/MaxwellHTTPServer.java
smferguson: i wondered the same. i can see both sides. i ended up here b/c if monitoring crashes you have significantly degraded visibility into what maxwell is doing.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
smferguson: good point. the issue is that `event.getEvent().getHeader().getTimestamp()` is in seconds though. the doc i created actually says it's millis. i'll update it.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
smferguson: yeah. when i went down that road i ran into issues with the classes hanging on init during tests. i suppose that makes this the tail wagging the dog a bit. i'm not thrilled with it.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
liulikun: I tried this and it's milliseconds for me. Can you double check on this? 

screen shot 2017-04-20 at 9 19 36 am

smferguson: checked. you are correct. thanks for keeping me honest.

smferguson:
calling this out since it's late in the game (in theory). i flipped the logic around for replicationLag... it's (now - db_txn_time) instead of (db_txn_time - now). makes more sense in my head.

liulikun:
👍

PR #584, opened 2017-03-06
add IPC use case

smferguson:
bit of a lol. calling this out matters to some parties.


osheroff:
does it really?  jesus.  ok, whatever.  you get a merge because you get a merge.

PR #585, opened 2017-03-06
use seconds instead of milliseconds for DDL message timestamp

liulikun:
### Description

use seconds instead of milliseconds for DDL message timestamp

/cc @osheroff @timbertson


osheroff:
👍

Issue #587, opened 2017-03-08
"int(11) COLLATE utf8_unicode_ci    "  when i run the sql  ,,the maxwell is error ,how can i modify that ? my maxwell's version is 1.5.0


rongyousu:
ALTER TABLE `you` ADD  `feee` int(11) COLLATE utf8_unicode_ci  NOT NULL DEFAULT '0' COMMENT 'eee' AFTER `id`
ALTER TABLE `you` ADD  `ddd` int COLLATE utf8_unicode_ci  NOT NULL DEFAULT '0' COMMENT 'eee' AFTER `id`

14:36:22,471 ERROR SchemaChange - Error parsing SQL: 'ALTER TABLE `you` ADD  `ddd` int COLLATE utf8_unicode_ci  NOT NULL DEFAULT '0' COMMENT 'eee' AFTER `id`'
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: COLLATE
	at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:84)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parseSQL(SchemaChange.java:79)
	at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:91)
[...]

PR #588, opened 2017-03-08
All columns will silently accept an irrelevant COLLATION

osheroff:
Addresses #587

@zendesk/rules @zendesk/mel-inf


osheroff:
yeah, we're just aping the whole parser, not just the functional parts of it

PR #595, opened 2017-03-13
Adding encryption

BradyPT:
Not including the trimString and removeNonAscii features that Dave and I had for Vertica purposes in this branch so a lot less needed changing. Also improved the testing efforts for encryption. This is just 1.8.1 with our encryption and the requirements are off from that they wanted still because they wanted the whole payload encrypted with an option to prefix that encrypted data with the kafka key so I wouldn't submit this still.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,11 @@
       <artifactId>lz4</artifactId>
       <version>1.3.0</version>
     </dependency>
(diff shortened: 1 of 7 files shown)


BradyPT:
This pull request went to the wrong place and I don't think it's ready yet, sorry!

PR #598, opened 2017-03-15
Fix boolean cli args not using false

PaulSchnau:
Currently, if you pass a flag that takes a Boolean argument, the argument is ignored and the flag is always set to true. This fixes that.

For instance:
`--gtid_mode=false` would being ignored and gtid_mode would be set to ON anyway.


osheroff:
ok 👍

Issue #604, opened 2017-03-23
Missing Default Constructor in RowMap


komibajaj:
I am trying to use Dozer for transforming from entity into RowMap and received the following exception, as RowMap is missing default no arg constructor.

[ERROR] [03/23/2017 11:23:59.482] [etl-akka-akka.actor.default-dispatcher-3] [akka://etl-akka/user/transformActor] java.lang.NoSuchMethodException: com.zendesk.maxwell.row.RowMap.<init>()
org.dozer.MappingException: java.lang.NoSuchMethodException: com.zendesk.maxwell.row.RowMap.<init>()
	at org.dozer.util.MappingUtils.throwMappingException(MappingUtils.java:82)
	at org.dozer.factory.ConstructionStrategies$ByConstructor.newInstance(ConstructionStrategies.java:261)
	at org.dozer.factory.ConstructionStrategies$ByConstructor.create(ConstructionStrategies.java:245)
	at org.dozer.factory.DestBeanCreator.create(DestBeanCreator.java:65)
	at org.dozer.MappingProcessor.map(MappingProcessor.java:177)
	at org.dozer.MappingProcessor.map(MappingProcessor.java:124)
	at org.dozer.MappingProcessor.map(MappingProcessor.java:119)
	at org.dozer.DozerBeanMapper.map(DozerBeanMapper.java:120)
	at com.webpt.actor.TransformActor.onReceive(TransformActor.java:32)
	at com.webpt.actor.TransformActor$$FastClassBySpringCGLIB$$7bab0b85.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:721)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:157)
	at org.springframework.aop.framework.adapter.MethodBeforeAdviceInterceptor.invoke(MethodBeforeAdviceInterceptor.java:52)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.aspectj.AspectJAfterThrowingAdvice.invoke(AspectJAfterThrowingAdvice.java:62)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
[...]

PR #605, opened 2017-03-23
Adding default constructor to RowMap pertaining to issue #604

komibajaj:
I am trying to use Dozer for auto transforming objects, and the absence of default constructor is throwing exception as specified in #604 . Please let me know if this solution works for everyone.


komibajaj:
Closing PR as most of the attributes of RowMap are final, and that will be big design change if I change that. I will keep issue open if that becomes more required.

PR #612, opened 2017-04-06
Add detailed fix instructions when master recovery fails

timbertson:
It's hopefully paranoia since this _shouldn't_ happen any more, but can if you have bad data left in the `positions` table (or maybe something else goes horribly wrong at just the right moment).

Log the full details when we find multiple `RecoveryInfo`s, as well as manual remediation steps (delete all `positions` rows _except_ for the most recent master), assuming the one with the most recent heartbeat is The Right One:tm:

/cc @zendesk/goanna


liulikun:
👍

osheroff:
hokay dokay 👍

Issue #614, opened 2017-04-10
Broker errors can cause message loss


nzwalker:
Hey folks,

Noticed a problem with the async callback completer in 

https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java#L54

In particular, an exception thrown by Kafka causes the message to be marked as completed. It turns out the producer API can throw exceptions during weird network partitions or broker issues - which is not handled correctly by the Java producer library.

In our case we require Maxwell outputs to be ultra durable, so I've patched it by causing an exception to crash the producer and deliberately restart Maxwell. Its pretty hacky:

```java
public class MaxwellKafkaProducer extends AbstractProducer {
	private final ArrayBlockingQueue<RowMap> queue;
	private final MaxwellKafkaProducerWorker worker;
	private Exception lastException;

	public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties, String kafkaTopic) {
		super(context);
		this.queue = new ArrayBlockingQueue<>(100);
		this.worker = new MaxwellKafkaProducerWorker(this, context, kafkaProperties, kafkaTopic, this.queue);
		new Thread(this.worker, "maxwell-kafka-worker").start();

	}

	public synchronized void stop(Exception e) {
		this.lastException = e;
	}

	public synchronized Exception getLastException() {
		return this.lastException;
	}

	@Override
	public void push(RowMap r) throws Exception {
		Exception e = this.getLastException();
		if (e != null) {
			throw e;
		}

		this.queue.put(r);
	}
}
```

And in the callback:

```java
@Override
	public void onCompletion(RecordMetadata md, Exception e) {
		if ( e != null ) {
			// We've seen some kind of Kafka exception
			// Since we want producing to be ultra-durable, we cannot mark this message as completed
			// The only safe thing to do is to stop processing any new messages and exit!
			LOGGER.error(e.getClass().getSimpleName() + " @ " + position + " -- " + key);
			LOGGER.error(e.getLocalizedMessage());
			if ( e instanceof RecordTooLargeException ) {
				LOGGER.error("Considering raising max.request.size broker-side.");
			}

			this.producer.stop(e);
		} else {
			if ( LOGGER.isDebugEnabled()) {
				LOGGER.debug("->  key:" + key + ", partition:" +md.partition() + ", offset:" + md.offset());
				LOGGER.debug("   " + this.json);
				LOGGER.debug("   " + position);
[...]

PR #615, opened 2017-04-11
Add task manager class to terminate all background tasks on exit / exception

timbertson:
I'd appreciate review @osheroff if you've got some time. (also @zendesk/goanna )

Key points:

 - new TaskManager class attached to MaxwellConfig which collects tasks (and eventually stops them). Tasks are:
   - replicator
   - positionStoreThread
   - (possibly) producer, for async Kafka producer
 - MaxwellContext.terminate tracks whether termination was clean or due to an exception
 - Replicator no longer checks for PositionStoreThread exceptions; the parent (RunLoopProcess takes care of being stopped by the task manager)

~A failed kafka publish callback now terminates the associated maxwell context (issue #614)~, as does exception in the MaxwellKafkaProducerWorker's run() thread. This should cause maxwell to actually terminate on error, instead of skipping messages or bravely forging ahead without a producer thread.

TODO:

- [x] add tests
- [x] mark all non-main threads as daemons so they can't accidentally prevent shutdown

### DIFF
--- a/src/main/java/com/zendesk/maxwell/util/StoppableTaskState.java
+++ b/src/main/java/com/zendesk/maxwell/util/StoppableTaskState.java
@@ -1,0 +1,51 @@
+package com.zendesk.maxwell.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeoutException;
+
+public class StoppableTaskState {
+	private static Logger LOGGER = LoggerFactory.getLogger(StoppableTaskState.class);
+	private volatile RunState state;
+	private final String description;
+
+	public StoppableTaskState(String description) {
+		state = RunState.RUNNING;
+		this.description = description;
+	}
+
+	public boolean keepGoing() {
+		return state == RunState.RUNNING;
+	}
+
+	public synchronized void requestStop() {
+		LOGGER.info(description + " requestStop() called (in state: " + state + ")");
+		if (state == RunState.RUNNING) {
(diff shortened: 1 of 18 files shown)


timbertson:
As @osheroff mentioned in #614 , people may want to run maxwell with less than 100% durability. I could back off this part of the change (this PR still fixes bugs when other threads don't terminate properly), but we'd still want it. Perhaps behind configuration?

```
parser.accepts("producer_failure", "terminate|log, action to take on producer write failure");
```

Note to self: need to make this affect the kinesis producer, too.

timbertson:
OK, I've reverted the "producer error is fatal" part of this PR. I'd like to add this back behind a config flag but that can happen in another PR. For the time being, this still cleans up the awkward code where the replicator is watching the PositionStoreThread, and terminates when the MaxwellKafkaProducerWorker thread dies.

src/main/java/com/zendesk/maxwell/util/StoppableTaskState.java
```diff
@@ -0,0 +1,55 @@
...
+		this.description = description;
+	}
+
+	public boolean keepGoing() {
+		return state == RunState.RUNNING;
+	}
+
+	public synchronized void requestStop() {
+		LOGGER.info(description + " requestStop() called (in state: " + state + ")");
+		if (state == RunState.RUNNING) {
```
liulikun: `if (keepGoing())`?

liulikun:
👍

PR #618, opened 2017-04-13
findSchema: determine newest schema based on binlog position, not insertion order

timbertson:
OK, this is a bit of a hairy one (with a conveniently simple fix). This is how we think it went down:

(/cc @zendesk/goanna @osheroff @hawknewton @vanchi-zendesk )

### A tale of two maxwells

Two (or more) maxwells are replicating the same DB to different kafka clusters. call them MAXWELL_A and MAXWELL_B.

MAXWELL_A (spuriously?) detects a master failover:
 - gets new position on the new master (and stores it in schemas)
 - immediately afterwards it falls over because that binlog file doesn't exist:

```
WARN  Recovery - attempting to recover from master-change: old-server-id: 175113239, file: account1-bin-log.002480, position: 711569117, heartbeat: 1490725038684
INFO  OpenReplicator - starting replication at account4-bin-log.001571:4
WARN  Recovery - recovered new master position: BinlogPosition[account4-bin-log.001571:684320609]
( ... )
INFO  OpenReplicator - starting replication at account4-bin-log.001571:684320609
ERROR MaxwellReplicator - Missing binlog 'account4-bin-log.001571' on account-master
ERROR MaxwellReplicator - Transport exception #1236
com.google.code.or.net.TransportException: Could not find first log file name in binary log index file
 at com.google.code.or.OpenReplicator.dumpBinlog(OpenReplicator.java:343)
 at com.google.code.or.OpenReplicator.start(OpenReplicator.java:115)
 at com.zendesk.maxwell.replication.MaxwellReplicator.startReplicator(MaxwellReplicator.java:127)
 at com.zendesk.maxwell.replication.MaxwellReplicator.beforeStart(MaxwellReplicator.java:142)
 at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:29)
 at com.zendesk.maxwell.Maxwell.start(Maxwell.java:178)
 at com.zendesk.maxwell.Maxwell.main(Maxwell.java:199)

# ( note that this still happens on shutdown, which possibly poisons the well for future restarts: )

INFO  PositionStoreThread - Storing final position: BinlogPosition[account4-bin-log.001571:684320609]
 ```

**Question: How did it recover this binlog position if it doesn't exist?**
If it the heartbeat was the last entry of that particular binlog, maybe there's an edge case where we start at the "next position", which doesn't exist (what we actually want is the beginning of the next binlog file)?

Whatever the reason, MAXWELL_A is out of action for a bit after this - it keeps trying to resume from this faulty binlog position.

Meanwhile, node MAXWELL_B has been following the DB just fine. It still sees `account1` as the master (never saw a failover to `account4`). A migration occurs. MAXWELL_B sees this and dutifully writes the schema diff into the `schemas` table. That occurred at account1-bin-log.002486, which is a position after MAXWELL_A dropped off.

MAXWELL_A comes back - it sees `account1` as the master, and recovers from its last known binlog position (account4-bin-log.001571:684320609). Since this is associated with a valid heartbeat, it doesn't matter that the binlog position isn't valid - it finds the new position on `account1` and actually recovers successfully. It writes its recovered schema to `schemas`, at binlog position `account1-bin-log.002480:711569117`.

This is correct behaviour, until MAXWELL_B gets restarted. It finds the latest schema which was _written_ for the active server (by sorting by ID), which is MAXWELL_A's recovery. MAXWELL_B now thinks that this schema (for an earlier binlog position than the schema change it recorded) is the latest schema, and proceeds to fall over when data comes for the modified table.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
@@ -114,8 +114,11 @@ public class BinlogPosition implements Serializable {
 			return false;
 		BinlogPosition otherPosition = (BinlogPosition) other;
 
-		return this.file.equals(otherPosition.file) && this.offset == otherPosition.offset
-			&& (gtidSetStr == null) ? otherPosition.gtidSetStr == null
-				: gtidSetStr.equals(otherPosition.gtidSetStr);
+		return this.file.equals(otherPosition.file)
+			&& this.offset == otherPosition.offset
+			&& (gtidSetStr == null
+					? otherPosition.gtidSetStr == null
+					: gtidSetStr.equals(otherPosition.gtidSetStr)
+				);
 	}
 }
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -543,11 +543,13 @@ public class MysqlSavedSchema {
 			}
 			return null;
 		} else {
+			// Only consider binlog positions before the target position on the current server.
+			// Within those, sort for the latest binlog file, then the latest binlog position.
 			PreparedStatement s = connection.prepareStatement(
 				"SELECT id from `schemas` "
 				+ "WHERE deleted = 0 "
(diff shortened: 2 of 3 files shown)


liulikun:
👍

Issue #626, opened 2017-05-01
maxwell fails in parsing create table statement which has int default -1 column.


ajitkumar91:
I have observed this on using maxwell with mysql 5.5.46, works well if i change default value to 0.

com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: -
        at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:84)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:41)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:52)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parseSQL(SchemaChange.java:88)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:100)
        at com.zendesk.maxwell.schema.AbstractSchemaStore.resolveSQL(AbstractSchemaStore.java:49)
        at com.zendesk.maxwell.schema.MysqlSchemaStore.processSQL(MysqlSchemaStore.java:91)
        at com.zendesk.maxwell.replication.AbstractReplicator.processQueryEvent(AbstractReplicator.java:71)
        at com.zendesk.maxwell.replication.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:376)
[...]

PR #628, opened 2017-05-02
allow for negative default numbers.

osheroff:
addresses #626

### DIFF
--- a/src/main/antlr4/imports/mysql_idents.g4
+++ b/src/main/antlr4/imports/mysql_idents.g4
@@ -19,8 +19,8 @@ name: ( id | tokens_available_for_names | INTEGER_LITERAL);
 id: ( IDENT | QUOTED_IDENT );
 literal: (float_literal | integer_literal | string_literal | NULL | TRUE | FALSE);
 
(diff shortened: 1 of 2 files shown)


src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -19,8 +19,8 @@ name: ( id | tokens_available_for_names | INTEGER_LITERAL);
 id: ( IDENT | QUOTED_IDENT );
 literal: (float_literal | integer_literal | string_literal | NULL | TRUE | FALSE);
 
-float_literal: INTEGER_LITERAL? '.' INTEGER_LITERAL;
-integer_literal: INTEGER_LITERAL;
+float_literal: '-'? INTEGER_LITERAL? '.' INTEGER_LITERAL;
```
timbertson: Looks like this also allows stuff like `-1.-123` - is that a problem or won't it matter because we'll never actually get that from mysql?

timbertson: (nevermind, @liulikun pointed out `INTEGER_LITERAL` vs `integer_literal`)

PR #630, opened 2017-05-04
MysqlSchemaStore: track last_heartbeat_read to protect against binlog number reuse

timbertson:
With the changes in 7b50dfcc0a37081d5e1ebbef37ad7aed531d2e00 to order by binlog position instead of just insertion order, maxwell might treat old schema records as new if the mysql binlog numbering ever goes backwards. This can occur e.g. if a server is rebuilt from backups, keeping the same server_id but resetting its binlog numbering.

By storing the last heartbeat, we can accurately detect the "newest" schema for a given position by sorting by last_heartbeat_read. We still sort by position as a secondary ordering, both for backwards-compatibility step (all existing schemas have a `0` timestamp) and because there could be multiple schemas stored within a single heartbeat interval.

/cc @zendesk/goanna 

### Deployment note (for the changelog):

if you deploy this version and then rollback, future schemas created by an older version of maxwell will receive a default `0` value for `last_heartbeat_read`. Before rolling forward again, you should manually reset _all_ `maxwell.schema` rows to have a `0` `last_heartbeat_read` column so that the schemas created with this version are not considered newer than those created after the rollback.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
@@ -9,6 +9,7 @@ import java.io.Serializable;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.Objects;
 
 public class BinlogPosition implements Serializable {
 	static final Logger LOGGER = LoggerFactory.getLogger(BinlogPosition.class);
@@ -21,18 +22,22 @@ public class BinlogPosition implements Serializable {
 	private final String gtid;
 	private final long offset;
 	private final String file;
-	private final Long heartbeat;
 
-	public BinlogPosition(String gtidSetStr, String gtid, long l, String file, Long heartbeat) {
+	// LastHeartbeat is the most recent heartbeat seen prior to this position.
+	// For a HeartbeatRow, it is the exact (new) heartbeat value for this position.
+	// It must be set for this position to be stored or used as a restore target position.
+	private final Long lastHeartbeat;
+
+	public BinlogPosition(String gtidSetStr, String gtid, long l, String file, Long lastHeartbeat) {
 		this.gtidSetStr = gtidSetStr;
 		this.gtid = gtid;
 		this.offset = l;
 		this.file = file;
-		this.heartbeat = heartbeat;
+		this.lastHeartbeat = lastHeartbeat;
 	}
 
-	public BinlogPosition(long l, String file, Long heartbeat) {
-		this(null, null, l, file, heartbeat);
+	public BinlogPosition(long l, String file, Long lastHeartbeat) {
+		this(null, null, l, file, lastHeartbeat);
 	}
 
 	public BinlogPosition(long l, String file) {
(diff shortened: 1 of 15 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
```diff
@@ -86,8 +105,11 @@ public GtidSet getGtidSet() {
 
 	@Override
 	public String toString() {
-		return "BinlogPosition[" +
-			(gtidSetStr == null ? file + ":" + offset : gtidSetStr) + "]";
+		return "BinlogPosition["
+			+ (gtidSetStr == null ? file + ":" + offset : gtidSetStr)
+			+ ", lastHeartbeat=" + String.valueOf(lastHeartbeat)
+			+ ", lastHeartbeat=" + String.valueOf(lastHeartbeat)
```
liulikun: duplicated?

src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
```diff
@@ -68,8 +81,14 @@ public String getFile() {
...
 
-	public Long getHeartbeat() {
-		return heartbeat;
+	public Long getLastHeartbeat() {
+		return lastHeartbeat;
+	}
+
+	public void requireLastHeartbeat() {
+		if (lastHeartbeat == null) {
+			throw new RuntimeException("BinlogPosition.lastHeartbeat must be set");
```
liulikun: Should we use a checked exception here?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -548,13 +553,14 @@ private static Long findSchema(Connection connection, BinlogPosition targetPosit
 			PreparedStatement s = connection.prepareStatement(
 				"SELECT id from `schemas` "
 				+ "WHERE deleted = 0 "
-				+ "AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position <= ?)) AND server_id = ? "
-				+ "ORDER BY binlog_file DESC, binlog_position DESC limit 1");
+				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position <= ?)) AND server_id = ? "
+				+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1");
```
liulikun: Should we just order by `last_heartbeat_read DESC, id DESC` since binlog_file is not indicative of ordering if DB is restored from backup?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -167,6 +168,11 @@ public static void upgradeSchemaStoreSchema(Connection c) throws SQLException, I
 			InputStream is = MysqlSavedSchema.class.getResourceAsStream("/sql/maxwell_schema_heartbeats.sql");
 			executeSQLInputStream(c, is, null);
 		}
+
+		if ( !getTableColumns("schemas", c).containsKey("last_heartbeat_read") ) {
+			// default 0 makes sorting easier (rows before this migration are older than those after)
+			performAlter(c, "alter table `schemas` add column last_heartbeat_read bigint null default 0");
```
liulikun: nit: quote `last_heartbeat_read`?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -167,6 +168,11 @@ public static void upgradeSchemaStoreSchema(Connection c) throws SQLException, I
 			InputStream is = MysqlSavedSchema.class.getResourceAsStream("/sql/maxwell_schema_heartbeats.sql");
 			executeSQLInputStream(c, is, null);
 		}
+
+		if ( !getTableColumns("schemas", c).containsKey("last_heartbeat_read") ) {
```
liulikun: nit: `if ( !schemaColumns.containsKey("last_heartbeat_read") ) {`

src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java
timbertson: If we ever hit this code path it's a programmer error - adding a checked exception would just add noise at the callsite since there's nothing code can actually do to recover. How about an AssertionError, since that's unchecked and less likely to be accidentally caught?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
timbertson: It only really comes into it if you have multiple schemas saved within a single heartbeat interval (typically a second or two). In that case I think the binlog position should be the best ordering - old recycled binlogs should be well in the past, but it could be possible for two maxwell clients to create schema snapshots at roughly the same time (in the wrong order), if one is slightly lagged. Both cases do seem very unlikely though 🤷‍♂️

liulikun:
👍

PR #631, opened 2017-05-05
add more logging for schema changes

liulikun:
more logging for schema changes
some minor cleanup

/cc @zendesk/goanna

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEvent.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEvent.java
@@ -62,7 +62,7 @@ public class BinlogConnectorEvent {
 
 	public void setLastHeartbeat(Long lastHeartbeat) {
 		this.position     = new BinlogPosition(gtidSetStr, gtid, this.position.getOffset(), this.position.getFile(), lastHeartbeat);
-		this.nextPosition = new BinlogPosition(gtidSetStr, gtid, this.position.getOffset(), this.position.getFile(), lastHeartbeat);
+		this.nextPosition = new BinlogPosition(gtidSetStr, gtid, this.nextPosition.getOffset(), this.nextPosition.getFile(), lastHeartbeat);
 	}
 
 	public Long getTableID() {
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEvent.java
```diff
@@ -62,7 +62,7 @@ public EventType getType() {
 
 	public void setLastHeartbeat(Long lastHeartbeat) {
 		this.position     = new BinlogPosition(gtidSetStr, gtid, this.position.getOffset(), this.position.getFile(), lastHeartbeat);
-		this.nextPosition = new BinlogPosition(gtidSetStr, gtid, this.position.getOffset(), this.position.getFile(), lastHeartbeat);
+		this.nextPosition = new BinlogPosition(gtidSetStr, gtid, this.nextPosition.getOffset(), this.nextPosition.getFile(), lastHeartbeat);
```
timbertson: This looks like it changes behaviour - the RowMap now ends up with `getPosition()` returning the _next_ position, not the position of the final event (which is used for marking a position as completed). Are we sure this is what we want?

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaStore.java
```diff
@@ -119,7 +122,13 @@ public void clone(Long serverID, BinlogPosition position) throws SchemaStoreExce
...
 
 			MysqlSavedSchema cloned = new MysqlSavedSchema(serverID, caseSensitivity, getSchema(), position, savedSchema.getSchemaID(), empty);
-			cloned.save(c);
+			Long schemaId = cloned.save(c);
+
+			String logMessage = "clone schema @" + position + " based on id " + savedSchema.getSchemaID();
+			if (schemaId != null) {
+				logMessage += ", new schema id is " + schemaId;
+			}
+			LOGGER.info(logMessage);
```
timbertson: Could just use `String.valueOf(schemaId)` here (and above) if all we care about is avoiding NPE

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEvent.java
liulikun: I think this was a mistake to set current position to `nextPosition`. OpenReplicator is using nextPosition for position store.

PR #635, opened 2017-05-10
ignore kafka/kinesis producer errors based on new configuration ignore_producer_error

liulikun:
We Kafka/Kinesis fails to publish the message, we only log the error currently. When setting the new config `ignore_producer_error` to false, it will enable us to shutdown Maxwell and reprocess the message again when Maxwell is running again (hopefully Kafka/Kinesis can publish the message by then).

/cc @timbertson @osheroff

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -138,7 +138,7 @@ client_id                      | STRING                              | unique te
 replica_server_id              | LONG                                | unique numeric identifier for this maxwell instance | 6379 (see notes)
 master_recovery                | BOOLEAN                             | enable experimental master recovery code            | false
 gtid_mode                      | BOOLEAN                             | enable GTID-based replication                       | false
-ignore_producer_error          | BOOLEAN                             | Maxwell will be terminated on kafka/kinesis errors when set it to false. Otherwise, those producer errors are only logged. | true
+ignore_producer_error          | BOOLEAN                             | Maxwell will be terminated on kafka/kinesis errors when false. Otherwise, those producer errors are only logged. | true
 &nbsp;
 replication_host               | STRING                              | mysql host to replicate from.  Only specify if different from `host` (see notes) | *schema-store host*
 replication_password           | STRING                              | password on replication server | (none)
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -138,6 +138,7 @@ client_id                      | STRING                              | unique te
 replica_server_id              | LONG                                | unique numeric identifier for this maxwell instance | 6379 (see notes)
 master_recovery                | BOOLEAN                             | enable experimental master recovery code            | false
 gtid_mode                      | BOOLEAN                             | enable GTID-based replication                       | false
+ignore_producer_error          | BOOLEAN                             | terminate Maxwell on Kafka/Kinesis errors if false  | true
 &nbsp;
 replication_host               | STRING                              | mysql host to replicate from.  Only specify if different from `host` (see notes) | *schema-store host*
 replication_password           | STRING                              | password on replication server | (none)
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -138,7 +138,7 @@ client_id                      | STRING                              | unique te
 replica_server_id              | LONG                                | unique numeric identifier for this maxwell instance | 6379 (see notes)
 master_recovery                | BOOLEAN                             | enable experimental master recovery code            | false
 gtid_mode                      | BOOLEAN                             | enable GTID-based replication                       | false
-ignore_producer_error          | BOOLEAN                             | terminate Maxwell on Kafka/Kinesis errors if false  | true
+ignore_producer_error          | BOOLEAN                             | Maxwell will be terminated on kafka/kinesis errors when set it to false. Otherwise, those producer errors are only logged. | true
 &nbsp;
 replication_host               | STRING                              | mysql host to replicate from.  Only specify if different from `host` (see notes) | *schema-store host*
 replication_password           | STRING                              | password on replication server | (none)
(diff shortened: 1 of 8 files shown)


docs/docs/config.md
```diff
@@ -138,6 +138,7 @@ client_id                      | STRING                              | unique te
 replica_server_id              | LONG                                | unique numeric identifier for this maxwell instance | 6379 (see notes)
 master_recovery                | BOOLEAN                             | enable experimental master recovery code            | false
 gtid_mode                      | BOOLEAN                             | enable GTID-based replication                       | false
+ignore_producer_error          | BOOLEAN                             | terminate Maxwell on Kafka/Kinesis errors if false  | true
```
timbertson: Describing the `false` case is a bit confusing. Perhaps "log producer errors instead of terminating maxwell"?

Alternatively, we could do something more explicit like `producer_error_action=log|terminate`, but it might not be worth the effort - we probably won't ever have a third action.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -165,6 +166,7 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "replay", "replay mode, don't store any information to the server").withOptionalArg();
 		parser.accepts( "master_recovery", "(experimental) enable master position recovery code").withOptionalArg();
 		parser.accepts( "gtid_mode", "(experimental) enable gtid mode").withOptionalArg();
+		parser.accepts( "ignore_producer_error", "maxwell is terminated on Kafka/Kinesis errors when set it to false, default to true").withOptionalArg();
```
timbertson: Make this the same wording as in the docs?

src/test/java/com/zendesk/maxwell/producer/KafkaCallbackTest.java
```diff
@@ -0,0 +1,49 @@
...
+			new BinlogPosition(1, "binlog-1"), "key", "value",
+			new Timer(), new Counter(), new Counter(), new Meter(), new Meter(),
+			context);
+		NotEnoughReplicasException error = new NotEnoughReplicasException("blah");
+		callback.onCompletion(new RecordMetadata(new TopicPartition("topic", 1), 1, 1), error);
+		verify(cc).markCompleted();
+	}
+
+	@Test
+	public void shouldTerminateWhenNotIgnoreProuderError() {
```
timbertson: typo - `prouder`

src/test/java/com/zendesk/maxwell/producer/KinesisCallbackTest.java
```diff
@@ -0,0 +1,42 @@
...
+		KinesisCallback callback = new KinesisCallback(cc,
+			new BinlogPosition(1, "binlog-1"), "key", "value",
+			context);
+		IrrecoverableError error = new IrrecoverableError("blah");
+		callback.onFailure(error);
+		verify(cc).markCompleted();
+	}
+
+	@Test
+	public void shouldTerminateWhenNotIgnoreProuderError() {
```
timbertson: copy-pasted typo ;)

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -51,7 +53,11 @@ public void onFailure(Throwable t) {
 
 		logger.error("Exception during put", t);
 
-		cc.markCompleted();
+		if (!context.getConfig().ignoreProducerError) {
+			context.terminate(new RuntimeException(t));
```
timbertson: Let's just make `terminate` take a throwable instead of having to wrap this.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -136,7 +142,7 @@ public MaxwellKafkaProducerWorker(MaxwellContext context, Properties kafkaProper
 			this.topic = "maxwell";
 		}
 
-		this.interpolateTopic = kafkaTopic.contains("%{");
+		this.interpolateTopic = this.topic.contains("%{");
```
timbertson: cheeky bug - might be worth a test if it's easy?

docs/docs/config.md
liulikun: The `true` case behaviour is self explanatory to some degree. The `false` behaviour needs clarification and attention since it causes service disruption. I'll explain both cases then.  Agree with what you said about `producer_error_action=log|terminate`.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
liulikun: It's not easy unless we expose a getter method just for this test.

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
liulikun: Hmm, this will change all the major methods in Maxwell to throw `Throwable` instead of `Exception`. Doesn't seem right.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson: Could we just instantiate with a null topic and make sure it doesn't explode? I'm assuming the previous code would...

src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
timbertson: Ugh, true. This'll do then - not like anyone's going to be catching it anyway 🤷‍♂️

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
liulikun: Can do. The test case may seem weird down the track.

docs/docs/config.md
```diff
@@ -138,6 +138,7 @@ client_id                      | STRING                              | unique te
 replica_server_id              | LONG                                | unique numeric identifier for this maxwell instance | 6379 (see notes)
 master_recovery                | BOOLEAN                             | enable experimental master recovery code            | false
 gtid_mode                      | BOOLEAN                             | enable GTID-based replication                       | false
+ignore_producer_error          | BOOLEAN                             | Maxwell will be terminated on kafka/kinesis errors when set it to false. Otherwise, those producer errors are only logged. | true
```
timbertson: nit: "when set it to false" can just be "when false"

PR #641, opened 2017-05-11
clean restart open replicator on missing heartbeats

liulikun:
When we restart open replicator due to missing heartbeats, maxwell could end up with a bad state:
the queue used by `MaxwellReplicator` contains messages with positions later than restarting position of the replicator, so the messages in the queue are out of order, so we cannot group them into clean mysql transactions anymore.

The fix here is to only mark replicator position at the end of transaction so we can restart at the start of the transaction. Also add a parser listener to the replicator so we can clear the queue on restart.

/cc @timbertson @osheroff

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
+++ b/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
@@ -35,7 +35,10 @@ public class InflightMessageList {
 	/* returns the position that stuff is complete up to, or null if there were no changes */
 	public synchronized BinlogPosition completeMessage(BinlogPosition p) {
 		InflightMessage m = this.linkedMap.get(p.toString());
-		assert(m != null);
+
+		if (m == null) {
+			return null;
+		}
 
 		m.isComplete = true;
 
--- a/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
@@ -2,6 +2,8 @@ package com.zendesk.maxwell.replication;
 
 import com.google.code.or.OpenReplicator;
 import com.google.code.or.binlog.BinlogEventV4;
+import com.google.code.or.binlog.BinlogParser;
+import com.google.code.or.binlog.BinlogParserListener;
 import com.google.code.or.binlog.impl.event.*;
 import com.google.code.or.common.util.MySQLConstants;
 import com.google.code.or.net.TransportException;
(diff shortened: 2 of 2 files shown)


src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -35,7 +35,10 @@ public synchronized void addMessage(BinlogPosition p) {
 	/* returns the position that stuff is complete up to, or null if there were no changes */
 	public synchronized BinlogPosition completeMessage(BinlogPosition p) {
 		InflightMessage m = this.linkedMap.get(p.toString());
-		assert(m != null);
+
+		if (m == null) {
+			return null;
+		}
```
timbertson: This shouldn't happen after this patch, right? In which case I'd probably leave this change out so we notice similar issues in the future.

src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
```diff
@@ -114,6 +116,28 @@ private void ensureReplicatorThread() throws Exception {
...
+					queue.clear();
+					try {
+						//wait a bit here for new messages to be pushed to the queue since the queue might be blocked before
+						Thread.sleep(2000);
+					} catch (InterruptedException e) {
+					}
+					//clear again in case some messages were pushed just now
+					queue.clear();
+				}
+			});
```
timbertson: This sleeping business isn't ideal, but it's better than what we've got. Hopefully we can deprecate this connector soon anyway.

src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java
```diff
@@ -214,7 +238,6 @@ private RowMapBuffer getTransactionRows() throws Exception {
 				continue;
 			}
 
-			setReplicatorPosition((AbstractBinlogEventV4) v4Event);
```
timbertson: Remind me why this went away - doesn't it mean that on restart, we'll jump back to the beginning of the most recent transaction? If we're in the middle of `getTransactionRows()` when that happens, we'll get very confused.

liulikun: With this code, the replicator can restart in the middle of a transaction. Since we clear the queue on restart, this is not ideal. I changed it to only set position at the end of transaction, so we can have a clean restart.

timbertson: But a replicator restart could happen within `getTransactionRows`, which would mean:

 - we're in a transaction, processing rows
 - we reset the replicator position to the end of the previous transaction
 - we continue to try and process the `BEGIN` or other non-transactional events inside a transaction
 - 😿

timbertson:
OR deprecated; closing

PR #650, opened 2017-05-16
move lastHeartbeat out of BinlogPosition and into an explicit Position class

### LINKED ISSUE #636: lastHeartbeat checks in BinlogPosition cause blowups in RDS
I think the changes in https://github.com/zendesk/maxwell/pull/630 break maxwell in the case where we're running in RDS and receive a RDS heartbeat.
Following the codepath  starting at https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/replication/MaxwellReplicator.java#L265 ends in calling the constructor here: https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java#L43 which results in the lastHeartbeat check failing here https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/replication/BinlogPosition.java#L88.

Maybe I'm just missing something?

edit:
I should note that maxwell doesn't die completely (at least in my debugger), but it does stacktrace and look pretty angry:

> 15:04:21,704 INFO  TaskManager - stopped all tasks
> Exception in thread "main" java.lang.AssertionError: BinlogPosition.lastHeartbeat must be set
> 	at com.zendesk.maxwell.replication.BinlogPosition.requireLastHeartbeat(BinlogPosition.java:90)
> 	at com.zendesk.maxwell.row.HeartbeatRowMap.<init>(HeartbeatRowMap.java:15)
> 	at com.zendesk.maxwell.replication.AbstractReplicator.processRDSHeartbeatInsertEvent(AbstractReplicator.java:95)
> 	at com.zendesk.maxwell.replication.MaxwellReplicator.processRDSHeartbeatInsertEvent(MaxwellReplicator.java:377)
> 	at com.zendesk.maxwell.replication.MaxwellReplicator.getTransactionRows(MaxwellReplicator.java:265)
> 	at com.zendesk.maxwell.replication.MaxwellReplicator.getRow(MaxwellReplicator.java:337)
> 	at com.zendesk.maxwell.replication.AbstractReplicator.work(AbstractReplicator.java:141)
> 	at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:30)
> 	at com.zendesk.maxwell.Maxwell.startInner(Maxwell.java:212)
> 	at com.zendesk.maxwell.Maxwell.start(Maxwell.java:151)
> 	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:236)

timbertson:
Compatibility note: the --init_position config format has changed to include
":HEARTBEAT" - i.e. it is now "BINLOG_FILE:POSITION:HEARTBEAT"

Benefits: the compiler enforces the presence of `lastHeartbeatRead` instead of relying on assertions at runtime. This makes a clear distinction between DB events which only have a binlog position, and positions used in the replicator which also have heartbeat information.

This also changes the position sha calculation to include lastHeartbeat. This has a small chance of allowing multiple concurrent maxwell instances to store duplicate schema changes if they are using different versions. That shouldn't break anything, and they will stop creating duplicates once all clients are upgraded to the new version.

Fixes #636

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -11,6 +11,7 @@ import com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationExceptio
 import com.fasterxml.jackson.databind.JavaType;
 import com.zendesk.maxwell.CaseSensitivity;
 import com.zendesk.maxwell.MaxwellContext;
+import com.zendesk.maxwell.replication.Position;
 import com.zendesk.maxwell.schema.columndef.*;
 
 import org.apache.commons.codec.digest.DigestUtils;
@@ -33,7 +34,7 @@ public class MysqlSavedSchema {
 	static int SchemaStoreVersion = 4;
 
 	private Schema schema;
-	private BinlogPosition position;
+	private Position position;
 	private Long schemaID;
 	private int schemaVersion;
 
@@ -58,17 +59,17 @@ public class MysqlSavedSchema {
 		this.sensitivity = sensitivity;
 	}
 
-	public MysqlSavedSchema(Long serverID, CaseSensitivity sensitivity, Schema schema, BinlogPosition position) throws SQLException {
+	public MysqlSavedSchema(Long serverID, CaseSensitivity sensitivity, Schema schema, Position position) throws SQLException {
 		this(serverID, sensitivity);
 		this.schema = schema;
 		setPosition(position);
 	}
 
-	public MysqlSavedSchema(MaxwellContext context, Schema schema, BinlogPosition position) throws SQLException {
+	public MysqlSavedSchema(MaxwellContext context, Schema schema, Position position) throws SQLException {
 		this(context.getServerID(), context.getCaseSensitivity(), schema, position);
 	}
 
-	public MysqlSavedSchema(Long serverID, CaseSensitivity sensitivity, Schema schema, BinlogPosition position,
+	public MysqlSavedSchema(Long serverID, CaseSensitivity sensitivity, Schema schema, Position position,
 							long baseSchemaID, List<ResolvedSchemaChange> deltas) throws SQLException {
 		this(serverID, sensitivity, schema, position);
 
@@ -76,7 +77,7 @@ public class MysqlSavedSchema {
 		this.deltas = deltas;
 	}
 
-	public MysqlSavedSchema createDerivedSchema(Schema newSchema, BinlogPosition position, List<ResolvedSchemaChange> deltas) throws SQLException {
+	public MysqlSavedSchema createDerivedSchema(Schema newSchema, Position position, List<ResolvedSchemaChange> deltas) throws SQLException {
 		if ( this.shouldSnapshotNextSchema )
 			return new MysqlSavedSchema(this.serverID, this.sensitivity, newSchema, position);
 		else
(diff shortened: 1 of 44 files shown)


ericnilo:
+1

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -746,11 +756,17 @@ protected void handleVersionUpgrades(Connection conn) throws SQLException, Inval
...
-	public static String getSchemaPositionSHA(Long serverID, String binlogFilename, Long binlogPosition) {
-		String shaString = String.format("%d/%s/%d", serverID, binlogFilename, binlogPosition);
+	public static String getSchemaPositionSHA(Long serverID, Position position) {
+		BinlogPosition binlogPosition = position.getBinlogPosition();
+		String shaString = String.format("%d/%s/%d/%d",
+			serverID,
+			binlogPosition.getFile(),
+			binlogPosition.getOffset(),
+			position.getLastHeartbeatRead()
+		);
```
liulikun: This could break `findSchemaForPositionSHA` since the existing rows in `schemas` table have the old sha?

src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
```diff
@@ -69,22 +73,24 @@ public void heartbeat() throws Exception {
...
 		if ( heartbeatPosition == null )
 			return true;
-		if ( !heartbeatPosition.getFile().equals(currentPosition.getFile()) )
+
+		BinlogPosition heartbeatBinlog = heartbeatPosition.getBinlogPosition();
+		BinlogPosition currentBinlog = currentPosition.getBinlogPosition();
+		if ( !heartbeatBinlog.getFile().equals(currentBinlog.getFile()) )
 			return true;
-		if ( currentPosition.getOffset() - heartbeatPosition.getOffset() > 1000 ) {
+		if ( currentBinlog.getOffset() - heartbeatBinlog.getOffset() > 1000 ) {
```
liulikun: Don't we want heartbeat all the time?

src/main/java/com/zendesk/maxwell/schema/SchemaStoreSchema.java
```diff
@@ -179,7 +181,11 @@ private static void backfillPositionSHAs(Connection c) throws SQLException {
 		ResultSet rs = c.createStatement().executeQuery("select * from `schemas`");
 		while (rs.next()) {
 			Long id = rs.getLong("id");
-			String sha = MysqlSavedSchema.getSchemaPositionSHA(rs.getLong("server_id"), rs.getString("binlog_file"), rs.getLong("binlog_position"));
+			Position position = new Position(
+				new BinlogPosition(rs.getLong("binlog_position"), rs.getString("binlog_file")),
+				rs.getLong("last_heartbeat_read")
+			);
+			String sha = MysqlSavedSchema.getSchemaPositionSHA(rs.getLong("server_id"), position);
```
liulikun: Also this may not work with the old records.

liulikun:
👍

src/main/java/com/zendesk/maxwell/schema/PositionStoreThread.java
timbertson: That should probably happen too (and remove RDS heartbeat support at the same time), but this PR is already big enough.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
timbertson: Yup, explained in the PR comment - I'm pretty sure that if this has an effect, it'll be short-lived and will fix itself once all clients are running the new version 🤞

smferguson:
thanks @timbertson !

osheroff:
this one slipped by me.  @timbertson, I'm really not sure: 

a) why init_position should include a heartbeat, or what the user is supposed to pass here (see #908), or
b) why we'd add it into the SHA -- the whole notion of the SHA there was that different clients shouldn't store two different copies of the schema changes, but if you add last_heartbeat, they very likely each will store their own copy of the change. 

In general, please avoid PRs that are like:
- This is a bug fix!
- This is a giant refactor!
- Oh, btw, I changed some other stuff, why not.

PR #651, opened 2017-05-16
support for kafka 0.10.2.1

liulikun:
/cc @timbertson


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -136,8 +136,6 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "kafka_partition_hash", "default|murmur3, hash function for partitioning").withRequiredArg();
 		parser.accepts( "kafka_topic", "optionally provide a topic name to push to. default: maxwell").withOptionalArg();
 		parser.accepts( "kafka_key_format", "how to format the kafka key; array|hash").withOptionalArg();
-		parser.accepts("kafka_version",
-		    "switch to kafka 0.8, 0.10 or 0.10.1 producer (from 0.9)");
```
liulikun: This is redundant, can't see any reference to it.

timbertson: I think this is needed so it shows up when you run with `--help` - the option itself is implemented by the shell wrapper (`/bin/maxwell`).

liulikun: Good point!  Though this option only works as an argument, doesn't work in config file.

PR #653, opened 2017-05-19
Polling bootstrap table for bootstrapping task.

### LINKED ISSUE #645: Cannot trigger bootstrap if use different 'schema storage host' and 'replica host'
If schema storage host is not the same as replica host,  `maxwell-bootstrap` command line report `Table does not exist` since it try to find table on schema host.

After insert a tuple in to bootstrap table `insert into maxwell.bootstrap (database_name, table_name) values ('xxx', 'xxx');`. Bootstrapping is not triggered. It seems that bootstrapping  is only triggered by an event on table `bootstrap`. But maxwell does not listen on schema storage host.

wangzw:
This pull request fix #645

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -41,6 +41,7 @@ public class MaxwellConfig extends AbstractConfig {
 	public String kafkaPartitionColumns;
 	public String kafkaPartitionFallback;
 	public String bootstrapperType;
+	public long bootstrapPollerInterval;
 	public int bufferedProducerSize;
 
 	public String producerPartitionKey;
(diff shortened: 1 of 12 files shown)


osheroff:
this approach needs some work, I think.  Having an external loop poll for events and then tunnel them into the existing bootstrap code is not ideal -- there's a bunch of possible race conditions, and it's also quite a bit hacky.  

Much better would be for the polling loop to directly control the bootstrapper -- it should be the one calling `startBootstrap`, `cancelBootstrap`, etc, and it shouldn't have to bother creating `RowMap` events -- it should just pass the database and table into the bootstrapper.

It'd also be nice (although not 100% necessary) if the polling code was the only code responsible for the `maxwell.bootstrap` table, and if it periodically updated the status of the bootstrapping and stuff.

wangzw:
@osheroff I agree with you. It is also the first design into my head. But to me, it is much important to make it work than make it perfect. The ideal approach needs more code change and refactor and I'm currently working on it.

If you think it is ok, I suggest to merge current code and refactor it in following patch.

osheroff:
The great engineer @stauugard once said to me "if you don't fix it now, you'll never fix it."  If I get the time I can try to pitch in on the refactor, but to me the feature isn't so key as to warrant leaving the code base in an in-between state

Issue #634, opened 2017-05-09
Feature request / discussion: embedded mode


geoff-addepar:
This is the sort of thing I would normally post on a mailing list, but I can't find one, so it's going in the bug tracker. :)

I'm wondering if it's possible to run Maxwell embedded inside another java application. And if that's not currently an option, if me and my team developed that ability, we would like to know whether or not it might be accepted upstream. Besides being able to run embedded, we would also like to be able to customize (at least) the producer's behavior, and I believe an embedded mode would allow for other useful extension points like more sophisticated message filtering, topic name transformations, custom metrics reporters, etc.

At a code level, I'm thinking this would involve some refactoring of Maxwell, MaxwellContext, and MaxwellConfig, but nothing would end up being too invasive. The code is already fairly modular, so I don't think it would be all that much work. In the end, I picture it being like the Apache HttpComponents' HttpClientBuilder, where you use a builder pattern to tie all of Maxwell's components together and allow custom code to override/configure the various pieces. Where this might fall apart is if Maxwell uses any static state, system properties, or anything like that. (I do see some logger configuration in a static init block; that might need a small tweak.)

We have two main motivating factors. One is a simplified development process for our application, where we won't have to make all our developers install and configure Maxwell and Kafka. People can just run a single application JAR and have everything come up in an embedded mode (we would also use an embedded message queue), though we would probably separate things out onto different machines when running on production. The other main factor is that we want to introduce custom behavior into the message producer (e.g. a standard message header for all messages in our environment) and we think the easiest way would be to pull Maxwell in as a maven dependency and configure it in our code.

Does the Maxwell team have any thoughts on this approach? Has it been considered or done before?

timbertson:
Seems reasonable to me - it's not a high priority for us, but as long as it's not invasive then making sure maxwell is modular is a good goal. 

@osheroff do you have strong opinions on forcing config to happen via properties (e.g. specifying names of classes) vs runtime-only config (allowing people to pass actual classes / instances)?

I expect it could get painful to have to maintain two configuration mechanisms, so perhaps try to deviate as little as possible from the existing MaxwellConfig (i.e. prefer putting things in there rather than separate APIs which are only available in embedded mode).

osheroff:
Maxwell should be *mostly* usable as a JAR right now.  The build pipeline isn't setup to auto-publish the jar artifacts to maven, but there's some there.  

[...]

PR #665, opened 2017-05-29
Pass around MaxwellMetrics instance instead of using static methods

timbertson:
Mainly this allows us to drop the awkward "only add a gauge if there isn't one with this name already" code whenever we add a custom gauge, but it should help with #634 too.

/cc @zendesk/goanna @smferguson

### DIFF
--- a/src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
+++ b/src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
@@ -1,26 +1,21 @@
 package com.zendesk.maxwell.metrics;
 
-import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Meter;
 import com.codahale.metrics.health.HealthCheck;
-import com.zendesk.maxwell.producer.AbstractAsyncProducer;
-import org.slf4j.Logger;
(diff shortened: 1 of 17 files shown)


src/main/java/com/zendesk/maxwell/metrics/MaxwellHealthCheck.java
```diff
@@ -1,29 +1,24 @@
...
+	public MaxwellHealthCheck(AbstractProducer producer) {
+		this.failedMessageMeter = producer.getFailedMessageMeter();
 	}
 
 	@Override
 	protected Result check() throws Exception {
 		// TODO: this should be configurable.
-		if (this.metricRegistry.getMeters().get(AbstractAsyncProducer.failedMessageMeterName).getFifteenMinuteRate() > 0) {
+		if (failedMessageMeter != null && failedMessageMeter.getFiveMinuteRate() > 0) {
 			return Result.unhealthy(">1 messages failed to be sent to Kafka in the past 15minutes");
```
smferguson: a little late to the game, but i think the text here needs updating ("15minutes" -> "5minutes")

timbertson: Thanks, that was unintentional - reverted in #672

PR #666, opened 2017-05-30
additional metrics

timbertson:
New metrics:

 - `replication.queue.time`: time taken to enqueue a DB event
 - `replication.queue.count`: size of the DB event queue
 - `inflightmessages.count`: number of inflight messages (waiting to be ACKed)
 - `time.overall` renamed to `publish.time` since it tracks the time between constructing a publishable message to that message being completed (ACKed)

The new replication metrics are only implemented for the binlog_connector mode. The inflight messages metric is implemented for async producers.

I refactored `rowMap` to store (and accept) millis instead of seconds so that `replicationLag` ban just be based on this (and because it's the only `seconds` timestamp we have), but aside from the constructor the interface remains unchanged.

/cc @zendesk/goanna @smferguson

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -78,16 +78,6 @@ public class BinlogConnectorReplicator extends AbstractReplicator implements Rep
 		this.client.setServerId(replicaServerID.intValue());
 
 		this.stopOnEOF = stopOnEOF;
-
-		metrics.getRegistry().register(
-				metrics.metricName("replication", "queue", "count"),
-				new Gauge<Long>() {
-					@Override
-					public Long getValue() {
-						return (long) queue.size();
-					}
-				}
-		);
 	}
 
 	public BinlogConnectorReplicator(SchemaStore schemaStore, AbstractProducer producer, AbstractBootstrapper bootstrapper, MaxwellContext ctx, Position start) throws SQLException {
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -80,16 +80,6 @@ public class BinlogConnectorReplicator extends AbstractReplicator implements Rep
 		this.stopOnEOF = stopOnEOF;
 
 		metrics.getRegistry().register(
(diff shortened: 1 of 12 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -68,12 +70,24 @@ public BinlogConnectorReplicator(
...
 		this.client.setServerId(replicaServerID.intValue());
 
 		this.stopOnEOF = stopOnEOF;
+
+		metrics.getRegistry().register(
+				metrics.metricName("replication", "queue", "count"),
+				new Gauge<Long>() {
+					@Override
+					public Long getValue() {
+						return (long) queue.size();
```
liulikun: `queue` is blocked queue with capacity of 20, I'm feeling this metrics might not give us much value. Can we instrument the blocking time?

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -67,12 +68,13 @@ protected JsonGenerator initialValue() {
...
 
-	public RowMap(String type, String database, String table, Long timestamp, List<String> pkColumns,
+	public RowMap(String type, String database, String table, Long timestampMillis, List<String> pkColumns,
 			Position nextPosition) {
 		this.rowType = type;
 		this.database = database;
 		this.table = table;
-		this.timestamp = timestamp;
+		this.timestampMillis = timestampMillis;
+		this.timestampSeconds = timestampMillis / 1000;
```
liulikun: Good naming!

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
timbertson: It doesn't tell us much about the DB, the important thing is whether maxwell is keeping up. i.e. when we see lag, is it because the DB is not sending binlog events quickly or because maxwell is too busy to read them?

I'm imagining this metric would dither between "almost 0" and "20", so maybe the actual number isn't important. It probably doesn't add anything over the wait time metric, so I'll nix it.

liulikun:
How about we measure the binlog processing rate (the speed we progress on binlog positions)? Then we can use it to monitor db traffic and maxwell capacity.

timbertson:
@liulikun Good idea, perhaps in a future PR.

PR #652, opened 2017-05-18
Stop processing RDS heartbeat events; improve shutdown heartbeat logic

timbertson:
Instead, we just cause maxwell to heartbeat at least every 10s.

Follow-on to #650 , and related to #636 .

Support for recognising RDS heartbeats was initially added (in #523) to ensure that quiet DBs still update their positions, this achieves the same result for non-RDS DBs as well.

In addition, for a clean shutdown we now always send a heartbeat and wait up to 10s until that heartbeat row has been fully processed. This ensures we actually see the final heartbeat (as long as we're not 10s lagged), and won't publish any events beyond this heartbeat. This means there'll be no duplicate events when maxwell starts up again.

Sample shutdown logs:

```
^C06:23:27,725 INFO  MaxwellContext - sending final heartbeat: 1495088607725
06:23:27,728 INFO  AbstractReplicator - received final heartbeat 1495088607725; stopping replicator
06:23:27,730 INFO  TaskManager - stopping 3 tasks
06:23:27,730 INFO  StoppableTaskState - MaxwellKafkaProducerWorker requestStop() called (in state: RUNNING)
06:23:27,735 INFO  KafkaProducer - Closing the Kafka producer with timeoutMillis = 9223372036854775807 ms.
06:23:27,743 INFO  StoppableTaskState - com.zendesk.maxwell.schema.PositionStoreThread requestStop() called (in state: RUNNING)
06:23:27,744 INFO  PositionStoreThread - Storing final position: Position[BinlogPosition[master.000004:66843351], lastHeartbeat=1495088607725]
06:23:27,749 INFO  StoppableTaskState - com.zendesk.maxwell.replication.MaxwellReplicator requestStop() called (in state: STOPPED)
06:23:28,592 INFO  TaskManager - stopped all tasks
```

/cc @zendesk/goanna @smferguson

PR #674, opened 2017-06-05
BufferedProducer: set position immediately for heartbeat rows

timbertson:
Test times increased dramatically after #652, because MaxwellTestSupport's use of a BufferedProducer in a way that never consumes the final heartbeat row, leading to timeouts at the end of every test which uses it. This PR sends heartbeats to the position store immediately.

There's a slim chance of false positives/negatives since now the position thread's position will be updated when a heartbeat row is _queued_, not _processed_. But I've eyeballed all the tests which make use of the position and they don't seem to be affected.

Another alternative would be to not bother with the heartbeat dance when we're using a buffered producer, but I like that we exercise this termination logic in the tests.

/cc @zendesk/goanna

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ config/*.yml
 .DS_Store
 /test.log
 .sw?
+/config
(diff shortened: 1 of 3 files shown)


liulikun:
👍

PR #679, opened 2017-06-16
Shutdown hardening

timbertson:
When `ignore_producer_error` is `false`, a kafka producer in certain bad states can prevent shutting down (even when closed with a timeout), leaving a maxwell process which is neither replicating nor publishing events.

This branch attempts to ensure the shutdown of maxwell under all circumstances, by force if necessary. The basic flow is:

 - entire shutdown logic (in MaxwellContext) has been moved to a thread. That thread gets 10 seconds to shutdown successfully. If an exception is thrown, or the thread doesn't complete in 10 seconds, the entire process is forcefully halted.

/cc @zendesk/goanna

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,7 @@ before_install:
   - sudo apt-get install pastebinit
 install: true
 script:
-  - echo travis_fold:start:compile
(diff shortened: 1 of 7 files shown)


liulikun:
👍

Issue #681, opened 2017-06-22
Inconsistent date and time handling


geoff-addepar:
I'm having trouble wrapping my head around how dates and times are handled in Maxwell, and I can't seem to get things to work correctly for me. I think I have at least 3 separate bugs here.

To summarize the differences:
- TIMESTAMP fields show up in the local time zone when `--binlog_connector=false`, and in UTC when `--binlog_connector=true`. I don't know which is correct, but I think they should at least be the same. Or document otherwise. (I prefer local, I think, but either is fine.)
- When `--binlog_connector=false`, DATETIME fields in the daylight savings gap get adjusted to a valid local time, but this means the string in Kafka doesn't match the string in the database. If DATETIME fields are being used to store UTC times (as they are in my case) then your data is basically being corrupted before being placed in the queue.
- DATETIMEs prior to 1970 that have a fractional seconds part cause a crash.

Reproduction steps (I'm in California, which is UTC-8 in this example):
```
create table test (dt datetime, ts timestamp null default null, fdt datetime(6));
insert into test (dt, ts) values('2017-01-01 01:00:00', '2017-01-01 01:00:00');
insert into test (dt) values('2017-03-12 02:30:00'); -- note that this is in the daylight savings gap in my local time zone
insert into test (fdt) values('1950-01-01 01:00:00.999999');
```
With --binlog_connector=false:
```
{"database":"test","table":"test","type":"insert","ts":1498089571,"xid":248502,"commit":true,"data":{"dt":"2017-01-01 01:00:00","ts":"2017-01-01 01:00:00","fdt":null,"fts":null}}
{"database":"test","table":"test","type":"insert","ts":1498089571,"xid":248507,"commit":true,"data":{"dt":"2017-03-12 03:30:00","ts":null,"fdt":null,"fts":null}}
{"database":"test","table":"test","type":"insert","ts":1498089572,"xid":248511,"commit":true,"data":{"dt":null,"ts":null,"fdt":"1950-01-01 01:00:00.999999","fts":null}}
```
With --binlog_connector=true:
```
{"database":"test","table":"test","type":"insert","ts":1498089667,"xid":248612,"commit":true,"data":{"dt":"2017-01-01 01:00:00","ts":"2017-01-01 09:00:00","fdt":null,"fts":null}}
[...]

PR #690, opened 2017-07-03
fix incorrect division / modulus operations on negative (pre-epoch) dates

timbertson:
> DATETIMEs prior to 1970 that have a fractional seconds part cause a crash.

Reported in https://github.com/zendesk/maxwell/issues/681

Java's truncation and modulus operations for negative numbers are mechanical, they don't quite match mathematics when dealing with negative numbers. This causes bounds errors and then rounding errors, fixed by using (backported versions of) `java.Math.{floorDiv,floorMod}` from java 8 instead of the native operators.

/cc @zendesk/goanna

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/columndef/DateFormatter.java
+++ b/src/main/java/com/zendesk/maxwell/schema/columndef/DateFormatter.java
@@ -22,9 +22,10 @@ public class DateFormatter {
 	public static Timestamp extractTimestamp(Object value) {
 		if (value instanceof Long) {
 			Long micros = (Long) value;
-			Timestamp t = new Timestamp(micros / 1000);
-
-			t.setNanos((int) (micros % 1000000) * 1000);
+			long millis = floorDiv(micros, 1000L);
+			Timestamp t = new Timestamp(millis);
+			long microsOnly = floorMod(micros, (long) 1000000);
+			t.setNanos((int) microsOnly * 1000);
 			return t;
 		} else if (value instanceof Timestamp) {
 			return (Timestamp) value;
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/schema/columndef/DateFormatter.java
```diff
@@ -33,17 +34,17 @@ public static Timestamp extractTimestamp(Object value) {
...
-			Timestamp t = extractTimestamp(value);
-			if ( t.getTime() < MIN_DATE )
-				return null;
-			else
-				return formatter.format(t);
+	private static Timestamp MIN_DATE = Timestamp.valueOf("1000-01-01 00:00:00");
+
+	private static String format(SimpleDateFormat formatter, Timestamp ts) {
+		if ( ts.before(MIN_DATE) ) {
+			return null;
```
liulikun: Should we throw errors here?

timbertson: An exception would stop maxwell from processing, which is probably not what we want (the data won't get any better; we have to do something and continue).

From 089d76bf19e77dee67097a7b64dd2a8381e3b80c (when this check was added) it sounds like such dates are invalid, so `null` ought to be fine:

> The only non-refactoring change here deals with how we deal with
"invalid" dates; those before '1000-01-01'.  Mysql's documentation seems
to indicate that dates like '0000-00-00' should be considered invalid,
although it accepts the damn things anyway.  Maxwell has had highly
differing policies on the data in the past; invalid dates coming from
the bootstrap code paths were null, dates from 5.5 were reproduced
faithfully, and dates from 5.6 were off by a few years.
binlog-connector makes 'em null, and that's good enough for me.

PR #692, opened 2017-07-10
RowMap minor fixes, optimizations and test cases

tariq1890:
- Added a null check for pkColumns. I felt this was required as pkColumns was being passed as null in other RowMap test cases. Thus, we must account for the possibility of null values of pkColumns. In the current code passing a null pkColumn value will result in a NullpointerException when calling pkToJsonHash() method.

- Convert String concatenation operation to append operation of StringBuilder. String concatenation would be a huge overhead, especially in loops.

- I've written JUnit test cases for the code around these changes.

### DIFF
--- a/src/test/java/com/zendesk/maxwell/row/RowMapTest.java
+++ b/src/test/java/com/zendesk/maxwell/row/RowMapTest.java
@@ -12,9 +12,15 @@ import java.util.*;
 import java.util.regex.Pattern;
 
 public class RowMapTest {
+
+	private static final long TIMESTAMP_MILLISECONDS = 1496712943447L;
+
+	private static final Position POSITION = new Position(new BinlogPosition(1L, "binlog-0001"), 0L);
+
+
 	@Test
 	public void testGetDataMaps() throws Exception {
-		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", 1234567890L, null, null);
+		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", 1234567890L, new ArrayList<String>(), null);
 		rowMap.putData("foo", "bar");
 		rowMap.putOldData("fiz", "buz");
 
@@ -43,14 +49,12 @@ public class RowMapTest {
 
 	@Test
 	public void testTimestampConversion() throws Exception {
-		long timestampMillis = 1496712943447L;
 		long timestampSeconds = 1496712943;
 
-		Position position = new Position(new BinlogPosition(1L, "binlog-0001"), 0L);
-		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", timestampMillis, null, position);
+		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", TIMESTAMP_MILLISECONDS, new ArrayList<String>(), POSITION);
 
 		Assert.assertEquals(timestampSeconds, rowMap.getTimestamp().longValue());
-		Assert.assertEquals(timestampMillis, rowMap.getTimestampMillis().longValue());
+		Assert.assertEquals(TIMESTAMP_MILLISECONDS, rowMap.getTimestampMillis().longValue());
 		Map<String, Object> output = MaxwellTestJSON.parseJSON(rowMap.toJSON());
 
 		int ts = (int) output.get("ts");
@@ -60,17 +64,13 @@ public class RowMapTest {
 	@Test
 	public void testPkToJsonHash() throws IOException {
 
-		long timestampMillis = 1496712943447L;
-
 		List<String> pKeys = new ArrayList<>();
 
 		pKeys.add("id");
 
 		pKeys.add("name");
 
-		Position position = new Position(new BinlogPosition(1L, "binlog-0001"), 0L);
-
-		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", timestampMillis, pKeys, position);
+		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", TIMESTAMP_MILLISECONDS, pKeys, POSITION);
 
 		rowMap.putData("id", "9001");
 		rowMap.putData("name", "example");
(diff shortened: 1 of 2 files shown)


tariq1890:
That's odd. I thought these changes were pretty minor. It seems to fail when the environment variable is set to MySQL version 5.6

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -55,4 +56,198 @@ public void testTimestampConversion() throws Exception {
...
+		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", timestampMillis,
+				null, position);
+
+		rowMap.putData("id", "9001");
+		rowMap.putData("first_name", "foo");
+		rowMap.putData("middle_name", "buzz");
+		rowMap.putData("last_name", "bar");
+		rowMap.putData("salary", "4000");
+		rowMap.putData("department", "science");
+        rowMap.putData("rawJSON", new RawJSONString("{\"database\":\"MyDatabase\",\"table\":\"User\",\"type\":\"insert\",\"ts\":1486439516,\"xid\":5694,\"commit\":true,\"data\":{\"UserID\":20,\"FirstName\":\"Fiz\",\"MiddleName\":null,\"LastName\":\"Buz\",\"Version\":1486703131000}}\n"));
```
timbertson: You can probably make this shorter - it's effectively string concat, right? It could probably be pulled out into a variable and reused, too.

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -55,4 +56,198 @@ public void testTimestampConversion() throws Exception {
...
+		rowMap.putData("id", "9001");
+		rowMap.putData("first_name", "foo");
+		rowMap.putData("middle_name", "buzz");
+		rowMap.putData("last_name", "bar");
+		rowMap.putData("salary", "4000");
+		rowMap.putData("department", "science");
+		rowMap.putData("interests", Arrays.asList("classical music", "hiking", "biking", "programming"));
+		MaxwellOutputConfig outputConfig = getMaxwellOutputConfig();
+
+		Assert.assertEquals("{\"database\":\"MyDatabase\",\"table\":\"MyTable\",\"type\":\"insert\",\"ts\":1496712943,\"position\":\"binlog-0001:1\",\"gtid\":null,\"data\":{\"id\":\"9001\",\"salary\":\"4000\",\"department\":\"science\",\"interests\":[\"classical music\",\"hiking\",\"biking\",\"programming\"]}}", rowMap.toJSON(outputConfig));
```
timbertson: Again, we probably don't need this much sample data - if it works with one & two fields, chances are it's going to work with 3+.

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -55,4 +56,198 @@ public void testTimestampConversion() throws Exception {
...
+		Pattern pattern = Pattern.compile("^.*name.*$");
+		patterns.add(pattern);
+
+		outputConfig.includesBinlogPosition = true;
+		outputConfig.includesCommitInfo = true;
+		outputConfig.includesGtidPosition = true;
+		outputConfig.includesServerId = true;
+		outputConfig.includesThreadId = true;
+		outputConfig.includesNulls = true;
+		outputConfig.excludeColumns = patterns;
```
timbertson: Do these tests actually care about all of these settings? It's fine to use the defaults unless you actually need different behaviour.

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -55,4 +56,198 @@ public void testTimestampConversion() throws Exception {
...
+		Assert.assertEquals("{\"database\":\"MyDatabase\",\"table\":\"MyTable\",\"type\":\"insert\",\"ts\":1496712943,\"position\":\"binlog-0001:1\",\"gtid\":null,\"data\":{\"id\":\"9001\",\"salary\":\"4000\",\"department\":\"science\",\"interests\":[\"classical music\",\"hiking\",\"biking\",\"programming\"]}}", rowMap.toJSON(outputConfig));
+	}
+
+	private MaxwellOutputConfig getMaxwellOutputConfig() {
+		MaxwellOutputConfig outputConfig = new MaxwellOutputConfig();
+
+
+
+		List<Pattern> patterns = new ArrayList<>();
+		Pattern pattern = Pattern.compile("^.*name.*$");
```
timbertson: If this is important for one or more tests, can we make it explicit? i.e. pass in a `Pattern` to this builder method and only actually use it for a test which is explicitly testing this filtering.

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -55,4 +56,198 @@ public void testTimestampConversion() throws Exception {
...
+
+		long timestampMillis = 1496712943447L;
+
+		List<String> pKeys = new ArrayList<>();
+
+		pKeys.add("id");
+
+		pKeys.add("name");
+
+		Position position = new Position(new BinlogPosition(1L, "binlog-0001"), 0L);
```
timbertson: It seems like timestampMillis and position are used in each test, and it doesn't look like you care about their values. In which case you might as well hoist them to a field instead of setting them in each test.

tariq1890: Sure. I can do that

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
tariq1890: I'll refine these tests. I'll set the serverID and threadID values to the rowMap object. I am doing this to achieve more line coverage. I hope that is okay

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
tariq1890: I am doing this to add more line coverage.

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
tariq1890: Sure. I'll do that

tariq1890:
@timbertson I've made the changes. Requesting your review :)

PR #698, opened 2017-07-15
bugfix, new table is not recognized if only dName changes

copart-sujit:
While restoring the schema, a new "currentDatabase.buildTable" is not recognized when the same table name is present in more than one databases

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -463,6 +463,7 @@ public class MysqlSavedSchema {
 
 		Database currentDatabase = null;
 		Table currentTable = null;
+		String currentTableDName = null;
 		int columnIndex = 0;
 
 		while (rs.next()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -492,7 +493,9 @@ private void restoreFullSchema(Connection conn, Long schemaID) throws SQLExcepti
 			if (tName == null) {
 				// if tName is null, there are no tables connected to this database
 				continue;
-			} else if (currentTable == null || !currentTable.getName().equals(tName)) {
+			} else if (currentTable == null || !currentTable.getName().equals(tName) || 
+					currentTableDName == null || !currentTableDName.equals(dbName)) {
```
liulikun: `currentTableDName == null` is always true here?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -492,7 +493,9 @@ private void restoreFullSchema(Connection conn, Long schemaID) throws SQLExcepti
...
+				currentTableDBName = dbName;
```
liulikun: `currentTableDBName` is a typo?

osheroff:
yeah, I see the bug here, but would like a test to reproduce the fix.  Also it's probably easier to resent `currenTable` to NULL when the database changes, yeah?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
copart-sujit: It's a typo. It should be `currentTableDName = dbName;`

copart-sujit: sorry :(

osheroff:
work re-done in #853

Issue #697, opened 2017-07-14
Crash when create database


niuhf0452:
Maxwell crashes when I create an new database.

Steps to reproduce this error:

* Prepare an empty mysql.
* Setup maxwell which watches on database 'A'.
* Run mysql, maxwell, zookeeper and kafka.
* Execute sql on mysql like this: `CREATE DATABASE A`

Then maxwell crash.

Logs:

```
java.lang.NullPointerException: null
 	at java.util.regex.Matcher.getTextLength(Matcher.java:1283) ~[?:1.8.0_131]
 	at java.util.regex.Matcher.reset(Matcher.java:309) ~[?:1.8.0_131]
 	at java.util.regex.Matcher.<init>(Matcher.java:229) ~[?:1.8.0_131]
 	at java.util.regex.Pattern.matcher(Pattern.java:1093) ~[?:1.8.0_131]
 	at com.zendesk.maxwell.MaxwellFilter.filterListsInclude(MaxwellFilter.java:111) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.MaxwellFilter.matchesTable(MaxwellFilter.java:123) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.MaxwellFilter.matches(MaxwellFilter.java:127) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.MaxwellFilter.matches(MaxwellFilter.java:149) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.schema.ddl.ResolvedSchemaChange.shouldOutput(ResolvedSchemaChange.java:26) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.replication.AbstractReplicator.processQueryEvent(AbstractReplicator.java:106) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.replication.MaxwellReplicator.processQueryEvent(MaxwellReplicator.java:367) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.replication.MaxwellReplicator.getRow(MaxwellReplicator.java:341) ~[maxwell-1.10.4.jar:1.10.4]
 	at com.zendesk.maxwell.replication.AbstractReplicator.work(AbstractReplicator.java:157) ~[maxwell-1.10.4.jar:1.10.4]
[...]

PR #699, opened 2017-07-15
Bugfix, Filter fails to parse alter database when "exclude_tables" option is enabled

copart-sujit:
Fixed NPE when com.zendesk.maxwell.schema.ddl.ResolvedSchemaChange.shouldOutput is invoked.


liulikun:
Can you give a concrete example on how to reproduce this error?

ericwush:
It might be related to #697 by the look of NPE.

ericwush:
Close this PR as it is not a correct fix. See #720 .

PR #706, opened 2017-07-22
default to binlog-connector mode

osheroff:
it's time enough that we just turned on this option by default.


liulikun:
Looks good! We can also cleanup `.travis.yml` a bit I think.

osheroff:
@liulikun good point, done

liulikun:
Thanks! 🎉 👍

adhiemsumitro:
How to know if this has been implemented on the latest binary ? I just realize that  the repository doesn't include a compiled binary (newbie here, 3 weeks usage)

liulikun:
You can check the releases: https://github.com/zendesk/maxwell/releases, is that what you are asking?

Issue #705, opened 2017-07-21
bin/maxwell-bootstrap error


elipoke:
[root@hadoopnn-01 maxwell-1.10.3]# bin/maxwell-bootstrap --user root --password 12345678 --database test --table ttt --log_level info
10:41:06,905 WARN  MaxwellBootstrapUtilityConfig - mysql host not specified, defaulting to localhost
10:41:06,920 INFO  MaxwellBootstrapConnectionPool - MaxwellBootstrapConnectionPool: Getting connection (user/password): jdbc:mysql://localhost:null/null
10:41:06,969 INFO  MaxwellBootstrapConnectionPool - MaxwellBootstrapConnectionPool: Can't create a new connection for jdbc:mysql://localhost:null/null
com.mysql.jdbc.exceptions.jdbc4.MySQLNonTransientConnectionException: Cannot load connection class because of underlying exception: 'java.lang.NumberFormatException: For input string: "null"'.
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:1.7.0_67]
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57) ~[?:1.7.0_67]
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:1.7.0_67]
        at java.lang.reflect.Constructor.newInstance(Constructor.java:526) ~[?:1.7.0_67]
        at com.mysql.jdbc.Util.handleNewInstance(Util.java:404) ~[mysql-connector-java-5.1.39.jar:5.1.39]
        at com.mysql.jdbc.Util.getInstance(Util.java:387) ~[mysql-connector-java-5.1.39.jar:5.1.39]
        at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:917) ~[mysql-connector-java-5.1.39.jar:5.1.39]
        at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:896) ~[mysql-connector-java-5.1.39.jar:5.1.39]
        at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:885) ~[mysql-connector-java-5.1.39.jar:5.1.39]
        at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:860) ~[mysql-connector-java-5.1.39.jar:5.1.39]
[...]

PR #712, opened 2017-07-26
setup proper defaults for the bootstrap utility

osheroff:
addresses #705, had to refactor some to use common code.


ericwush:
@osheroff I can reproduce the original issue #705 by not providing any `config.properties`. By looking at your PR I'm almost sure it has fixed the issue. Can you please confirm if that's the case?

osheroff:
it should.  just needs a code-review from some folks and should be mergable

PR #713, opened 2017-07-27
Add compatibility notes for BinlogConnector, fixes #681

### LINKED ISSUE #681: Inconsistent date and time handling

timbertson:
/cc @geoff-addepar @zendesk/goanna

### DIFF
--- a/docs/docs/compat.md
+++ b/docs/docs/compat.md
@@ -36,4 +36,15 @@ Notes:
 - on highly active servers, as much as 1 second of data may be duplicated.
 - master recovery is not available in GTID-mode.
 
+### MySQL binlog connector
+***
 
+As of 1.10.5, maxwell uses [shyiko/mysql-binlog-connector-java][] as its underlying
+replication library (previously it was opt-in via `--binlog_connector`). This is
+largely compatible with the previous OpenReplicator implementation, but there are some differences:
+
+ - TIMESTAMP columns are always treated as UTC, regardless of your timezone. See
+   [issue #681][issue-681] for more details.
+
+[shyiko/mysql-binlog-connector-java]: https://github.com/shyiko/mysql-binlog-connector-java
+[issue-681]: https://github.com/zendesk/maxwell/issues/681


geoff-addepar:
I'm not sure this is quite the right way to explain it. Specifically, it's TIMESTAMP columns that have different behavior, not DATETIME columns.

timbertson:
Good point @geoff-addepar, updated to only mention `TIMESTAMP` columns.

geoff-addepar:
Thanks, looks good!

PR #720, opened 2017-08-01
Skip table filters for any db schema change if output ddl

ericwush:
@zendesk/goanna 
Replace #699 addressing #697


timbertson:
Looks good. Can you add a test to cover it? Looks like we have some in MaxwellIntegrationTest you could add to, or feel free to create a more targeted suite if you prefer.

ericwush:
@timbertson test added

PR #723, opened 2017-08-02
cleanup old recovery infos before maxwell starts repliction

liulikun:
### Description

We can only have one recovery info per client id when `masterRecovery` is enabled to prevent pollution of start position when failing over to the previous server. 

/cc @timbertson @ericwush

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
@@ -251,14 +251,19 @@ public class MysqlPositionStore {
 		return result;
 	}
 
-	public void cleanupOldPositions() throws SQLException {
-		try ( Connection c = connectionPool.getConnection()) {
-			PreparedStatement s = c.prepareStatement(
-				"DELETE FROM `positions` WHERE server_id <> ? AND client_id = ?"
-			);
-			s.setLong(1, serverID);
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
```diff
@@ -246,23 +246,24 @@ protected RecoveryInfo getRecoveryInfo(MaxwellConfig config, Connection c) throw
...
-				"DELETE from `positions` where server_id = ? and client_id = ? and binlog_file = ? and binlog_position = ?"
-			);
-			BinlogPosition binlogPosition = position.getBinlogPosition();
-			s.setLong(1, serverID);
-			s.setString(2, clientID);
-			s.setString(3, binlogPosition.getFile());
-			s.setLong(4, binlogPosition.getOffset());
-			s.execute();
-			return s.getUpdateCount();
+	public void cleanupOldRecoveryInfos() throws SQLException {
```
timbertson: Not sure about the name, as these aren't recovery infos, they're just old positions. perhaps `deleteInactiveServerPositions`?

liulikun: I was trying to keep the consistent naming here since we get those positions by calling `getAllRecoveryInfos`. And those positions were used for recovery at one stage in the past.

Issue #725, opened 2017-08-04
Maxwell RowMapDeserializer getValue() method doesn't consider floating point number


rrlichwa:
In the data and old fields, I have a DoubleNode with value "30224.5" but getValue method converts it to 30224.

Is there any reason why there is no support for floating point numbers?

PR #726, opened 2017-08-06
Update RowMapDeserializer to return native number type of ValueNode a…

rrlichwa:
…nd store commit value.

https://github.com/zendesk/maxwell/issues/725


rrlichwa:
Hello, anyone have thoughts on this PR?  Sorry, I'm new to contributing to maxwell.

src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
```diff
@@ -99,12 +105,32 @@ public RowMap deserialize(JsonParser jsonParser, DeserializationContext deserial
 		return rowMap;
 	}
 
-	private Object getValue(ValueNode value)
+	@VisibleForTesting
+	final Object getValue(ValueNode value)
```
ericwush: I wouldn't recommend changing the method accessibility only for testing purpose. In this case it means we can't unit test the method directly but it can still be tested on a higher level.

src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
```diff
@@ -99,12 +105,32 @@ public RowMap deserialize(JsonParser jsonParser, DeserializationContext deserial
...
+					return value.intValue();
+				case BIG_DECIMAL:
+					return value.decimalValue();
+				case BIG_INTEGER:
+					return value.bigIntegerValue();
+				default:
+					return value.asText();
+			}
+		}
+
```
ericwush: You can remove the if `value.canConvertToInt()` and `value.canConvertToLong()` blocks below.

src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
rrlichwa: Changes made

src/main/java/com/zendesk/maxwell/row/RowMapDeserializer.java
rrlichwa: Changes made

ericwush:
@rrlichwa Sweet. If you can remove the import `import com.google.common.annotations.VisibleForTesting;` in `RowMapDeserializer.java` it's ready to go.

rrlichwa:
Sorry about that - all set.  Thanks!

Issue #722, opened 2017-08-01
Crash with EOFException on certain valid JSON column operations


danmactough:
maxwell v1.10.5
mysql v5.7.15

If I try to `INSERT` into a table with a JSON column but don't specify any value for the JSON column, maxwell crashes with an EOFException:

```
19:18:14,714 INFO  TaskManager - Stopping 2 tasks
19:18:14,714 ERROR TaskManager - cause:
java.lang.RuntimeException: java.io.EOFException
```

<details>
<summary>Click for complete stack trace</summary>

```
19:18:14,714 INFO  TaskManager - Stopping 2 tasks
19:18:14,714 ERROR TaskManager - cause:
java.lang.RuntimeException: java.io.EOFException
	at com.zendesk.maxwell.schema.columndef.JsonColumnDef.asJSON(JsonColumnDef.java:26) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.replication.BinlogConnectorEvent.writeData(BinlogConnectorEvent.java:88) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.replication.BinlogConnectorEvent.buildRowMap(BinlogConnectorEvent.java:136) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.replication.BinlogConnectorEvent.jsonMaps(BinlogConnectorEvent.java:148) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getTransactionRows(BinlogConnectorReplicator.java:160) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getRow(BinlogConnectorReplicator.java:271) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.replication.AbstractReplicator.work(AbstractReplicator.java:157) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:27) ~[maxwell-1.10.5.jar:1.10.5]
	at com.zendesk.maxwell.Maxwell.startInner(Maxwell.java:183) ~[maxwell-1.10.5.jar:1.10.5]
[...]

PR #728, opened 2017-08-09
Fix EOFException on empty JSON column operations

ericwush:
@zendesk/goanna 
Fix is based on scenario described in #722 regarding returning `null` for empty JSON column value.

`INSERT INTO test_table (props) VALUES (null);` -> `{"database":"A","table":"test_table","type":"insert","ts":1502241788,"xid":1435,"commit":true,"data":{"id":9,"props":null}}`

`INSERT INTO test_table () VALUES ();` -> 
`{"database":"A","table":"test_table","type":"insert","ts":1502241791,"xid":1442,"commit":true,"data":{"id":10,"props":null}}`

### DIFF
--- a/src/test/java/com/zendesk/maxwell/schema/columndef/ColumnDefTest.java
+++ b/src/test/java/com/zendesk/maxwell/schema/columndef/ColumnDefTest.java
@@ -14,6 +14,7 @@ import java.util.GregorianCalendar;
 import java.util.TimeZone;
 
 import com.zendesk.maxwell.TestWithNameLogging;
+import com.zendesk.maxwell.row.RawJSONString;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


src/test/java/com/zendesk/maxwell/schema/columndef/ColumnDefTest.java
```diff
@@ -143,6 +144,40 @@ public void TestStringAsJSON() {
...
+		input[4] = Byte.valueOf((byte) 0);
+		input[5] = Byte.valueOf((byte) 11);
+		input[6] = Byte.valueOf((byte) 0);
+		input[7] = Byte.valueOf((byte) 2);
+		input[8] = Byte.valueOf((byte) 0);
+		input[9] = Byte.valueOf((byte) 5);
+		input[10] = Byte.valueOf((byte) 3);
+		input[11] = Byte.valueOf((byte) 0);
+		input[12] = Byte.valueOf((byte) 105);
+		input[13] = Byte.valueOf((byte) 100);
```
timbertson: Seems a little verbose, can you use `new byte[] { ... }` instead of setting each individual byte?

danmactough:
Awesome. Thanks @ericwush!

PR #735, opened 2017-08-21
add metrics for mysql transaction execution time

liulikun:
/cc @timbertson @ericwush

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
@@ -1,17 +1,17 @@
 package com.zendesk.maxwell.replication;
 
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import com.codahale.metrics.Timer;
 import com.github.shyiko.mysql.binlog.BinaryLogClient;
 import com.github.shyiko.mysql.binlog.event.Event;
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
```diff
@@ -1,36 +1,38 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 class BinlogConnectorEventListener implements BinaryLogClient.EventListener,
 					      BinaryLogClient.LifecycleListener {
 	private static final Logger LOGGER = LoggerFactory.getLogger(BinlogConnectorEventListener.class);
 
 	private final BlockingQueue<BinlogConnectorEvent> queue;
 	private Timer queueTimer;
+	private Timer mysqlTxExecutionTimer;
```
ericwush: It's nice to have those private parameters `final`, but not mandatory. Can setup IDE to do so.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
```diff
@@ -41,6 +43,11 @@ public void stop() {
 	public void onEvent(Event event) {
 		long eventSeenAt = 0L;
 
+		if (event.getHeader().getEventType() == EventType.QUERY) {
```
timbertson: We already have a local `trackMetrics` below, for XID events. I would assume execution time is per transaction, not per update - can we track this only for XIDs, or is that not how it works?

PR #740, opened 2017-08-25
report lag metrics only on commits events

liulikun:
### Description

Only report lag on commit events as when MySQL is slow on transactions, the lag will be incorrect based on the timestamp of every event (row change).

Also measure the execution time internally instead of using the value in the header.

/cc @timbertson @ericwush

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
@@ -23,6 +23,7 @@ class BinlogConnectorEventListener implements BinaryLogClient.EventListener,
 	protected final AtomicBoolean mustStop = new AtomicBoolean(false);
 	private final BinaryLogClient client;
 	private long replicationLag;
+	private String gtid;
 
 	public BinlogConnectorEventListener(
 		BinaryLogClient client,
@@ -50,7 +51,6 @@ class BinlogConnectorEventListener implements BinaryLogClient.EventListener,
 
 	@Override
 	public void onEvent(Event event) {
-		String gtid = null;
 		long eventSeenAt = 0;
 		boolean trackMetrics = false;
 
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
@@ -1,11 +1,13 @@
 package com.zendesk.maxwell.replication;
 
+import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Timer;
 import com.github.shyiko.mysql.binlog.BinaryLogClient;
 import com.github.shyiko.mysql.binlog.event.Event;
 import com.github.shyiko.mysql.binlog.event.EventType;
 import com.github.shyiko.mysql.binlog.event.GtidEventData;
 import com.github.shyiko.mysql.binlog.event.QueryEventData;
+import com.zendesk.maxwell.metrics.Metrics;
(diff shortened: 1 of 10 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
```diff
@@ -44,19 +70,23 @@ public void onEvent(Event event) {
...
+		if (eventType == EventType.QUERY) {
+			transactionStartTime = event.getHeader().getTimestamp();
+		} else if (eventType == EventType.XID) {
 			trackMetrics = true;
 			eventSeenAt = System.currentTimeMillis();
-			QueryEventData queryEventData = event.getData();
-			mysqlTxExecutionTimer.update(queryEventData.getExecutionTime(), TimeUnit.SECONDS);
+			long transactionCommitTime = event.getHeader().getTimestamp();
+			replicationLag = eventSeenAt - transactionCommitTime;
+			executionTime = transactionCommitTime - transactionStartTime;
```
timbertson: Do we need some sanity checking here in case it gets triggered without seeing a prior QUERY event type? (i.e. if `transactionStartTime` == 0 it's probably not legitimate). Or are we guaranteed to always see QUERY before XID?

liulikun: Good point. Will move this metric into replicator after transaction rollup.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
```diff
@@ -41,22 +44,23 @@ public void stop() {
 
 	@Override
 	public void onEvent(Event event) {
-		long eventSeenAt = 0L;
-		boolean trackMetrics = false;
+		EventType eventType = event.getHeader().getEventType();
```
timbertson: TODO: put `eventSeenAt`, `trackMetrics` and `gtid` into locals here.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
```diff
@@ -19,20 +21,21 @@
...
 		BlockingQueue<BinlogConnectorEvent> q,
-		Timer queueTimer,
-		Timer mysqlTxExecutionTimer) {
+		Metrics metrics) {
 		this.client = client;
 		this.queue = q;
-		this.queueTimer = queueTimer;
-		this.mysqlTxExecutionTimer = mysqlTxExecutionTimer;
+		this.queueTimer =  metrics.getRegistry().timer(metrics.metricName("replication", "queue", "time"));
+		this.replicationLag = metrics.getRegistry().histogram(metrics.metricName("replication", "lag"));
```
timbertson: TODO: gauge?

src/main/java/com/zendesk/maxwell/util/ListWithDiskBuffer.java
```diff
@@ -51,6 +51,7 @@ public boolean isEmpty() {
 	public T getLast() {
 		return list.getLast();
 	}
+	public T getFirst() { return list.getFirst(); }
```
timbertson: I don't think this is correct in the case where the list has overflowed onto disk - can you add a test to confirm? If it's expensive to grab the element from disk, we could potentially work around the need for this by capturing the first element's timestamp when we add it, rather than at the end of the transaction.

liulikun: Good point. Will do that.

PR #742, opened 2017-08-30
Diagnostic health checks

ericwush:
@zendesk/goanna


src/main/java/com/zendesk/maxwell/monitoring/MaxwellHTTPServer.java
```diff
@@ -0,0 +1,143 @@
...
+class MaxwellHTTPServerWorker implements StoppableTask, Runnable {
+
+	private int port;
+	private final String pathPrefix;
+	private final MaxwellMetrics.Registries metricsRegistries;
+	private final MaxwellDiagnosticContext diagnosticContext;
+	private Server server;
+
+	public MaxwellHTTPServerWorker(int port, String pathPrefix, MaxwellMetrics.Registries metricsRegistries,
+																 MaxwellDiagnosticContext diagnosticContext) {
```
timbertson: indentation

src/main/java/com/zendesk/maxwell/monitoring/MaxwellHTTPServer.java
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	@Override
+	public void requestStop() {
+		try {
+			this.server.stop();
+		} catch (Exception e) {
+			e.printStackTrace();
+		}
```
timbertson: Don't swallow exceptions here; requestStop() should just throw Exception (we catch them in maxwellContext.shutdown)

src/main/java/com/zendesk/maxwell/monitoring/MaxwellHTTPServer.java
```diff
@@ -0,0 +1,143 @@
...
+	private String genPath(String path) {
+		if (StringUtils.isNotBlank(path)) {
+			if (!path.startsWith("/")) {
+				return "/" + path;
+			}
+			return path;
+		} else {
+			return null;
+		}
+	}
```
timbertson: These two methods allow for a lot of leeway (nulls, empty, slashes in various places). Could we instead normalize the input configuration, so that:

 - `pathPrefix` is always a non-null string which starts and ends with a slash, ie.. `"/"` or `"/some/path/"`
 - `path` never starts with a slash

That way string concat will give you nice paths without needing to think about the possible variations.

I haven't checked, but it's also possible we could put the prefix in the call to `new ServletContextHandler`, and then we just make sure that it handles prefixes sanely.

src/main/java/com/zendesk/maxwell/monitoring/MaxwellMetrics.java
```diff
@@ -123,16 +118,21 @@ public String metricName(String... names) {
...
 
 	public void startBackgroundTasks(MaxwellContext context) throws IOException {
-		String metricsReportingType = config.metricsReportingType;
-		if (metricsReportingType != null && metricsReportingType.contains(reportingTypeHttp)) {
-			LOGGER.info("Metrics http server starting");
-			new MaxwellHTTPServer(config.metricsHTTPPort, metricRegistry, healthCheckRegistry, context);
-			healthCheckRegistry.register("MaxwellHealth", new MaxwellHealthCheck(context.getProducer()));
-			LOGGER.info("Metrics http server started on port " + config.metricsHTTPPort);
+		context.getConfig().healthCheckRegistry.register("MaxwellHealth", new MaxwellHealthCheck(context.getProducer()));
+		MaxwellHTTPServer.startIfRequired(context);
```
timbertson: This feels misplaced now, since we may be starting the HTTP server even if HTTP metrics are disabled. Perhaps move this `startIfRequired` call up to MaxwellContext's `start()` since it might be enabled by both metrics and diagnostics?

src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
```diff
@@ -61,15 +60,7 @@ public AbstractAsyncProducer(MaxwellContext context) {
...
-		metricRegistry.register(
-			gaugeName,
-			new Gauge<Long>() {
-				@Override
-				public Long getValue() {
-					return (long) inflightMessages.size();
-				}
-			}
-		);
+		metricRegistry.register(gaugeName, (Gauge<Long>) () -> (long) inflightMessages.size());
```
timbertson: Ahh, brevity 😀

src/test/java/com/zendesk/maxwell/monitoring/DiagnosticMaxwellTest.java
```diff
@@ -0,0 +1,92 @@
...
+		new Thread(maxwell).start();
+		latch.await();
+
+		// When
+		healthCheck.doGet(request, response);
+		writer.flush();
+
+		// Then
+		String result = new String(outputStream.toByteArray());
+		assertTrue(result.contains("\"name\":\"binlog-connector\",\"success\":true"));
```
timbertson: If it's not horribly verbose (I haven't tried), it'd be nice to parse this into a map and check the actual object instead of relying on string comparison.

src/test/java/com/zendesk/maxwell/monitoring/DiagnosticMaxwellTest.java
```diff
@@ -0,0 +1,92 @@
...
+		healthCheck.doGet(request, response);
+		writer.flush();
+
+		// Then
+		String result = new String(outputStream.toByteArray());
+		assertTrue(result.contains("\"name\":\"binlog-connector\",\"success\":false"));
+	}
+
+	private DiagnosticHealthCheck getDiagnosticHealthCheck(MaxwellDiagnosticContext.Config config,
+																												 MaxwellContext maxwellContext) throws ServletException {
```
timbertson: indentation

config.properties.example
```diff
@@ -209,6 +212,18 @@ kinesis_stream=maxwell
 #metrics_datadog_host=localhost # default localhost
 #metrics_datadog_port=8125 # default 8125
 
+# Maxwell exposes http diagnostic endpoint to check below in parallel:
+# 1. binlog replication lag
+# 2. producer (currently kafka) lag
+
+# To enable Maxwell diagnostic
+#diagnostic=true # default false
```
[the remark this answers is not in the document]
ericwush: 👍

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -317,14 +321,21 @@ private void setup(OptionSet options, Properties properties) {
...
 		this.metricsDatadogType = fetchOption("metrics_datadog_type", options, properties, "udp");
 		this.metricsDatadogTags = fetchOption("metrics_datadog_tags", options, properties, "");
 		this.metricsDatadogAPIKey = fetchOption("metrics_datadog_apikey", options, properties, "");
 		this.metricsDatadogHost = fetchOption("metrics_datadog_host", options, properties, "localhost");
 		this.metricsDatadogPort = Integer.parseInt(fetchOption("metrics_datadog_port", options, properties, "8125"));
 		this.metricsDatadogInterval = fetchLongOption("metrics_datadog_interval", options, properties, 60L);
 
+		this.diagnosticConfig = new MaxwellDiagnosticContext.Config();
+		this.diagnosticConfig.enable = fetchBooleanOption("diagnostic", options, properties, false);
+		this.diagnosticConfig.port = this.monitoringHTTPPort;
```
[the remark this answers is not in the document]
ericwush: This port is unused after refactoring, will be removed.

src/main/java/com/zendesk/maxwell/monitoring/DiagnosticHealthCheckModule.java
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	@Override
+	public String getModuleName() {
+		return "diagnostic";
+	}
+
+	@Override
+	public Version version() {
+		return new Version(0, 1, 0, "", "com.zendesk", "maxwell");
```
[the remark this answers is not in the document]
ericwush: TBH I have no idea. But it can't be null so I've made one up. See `MetricsModule.VERSION`.

src/main/java/com/zendesk/maxwell/monitoring/MaxwellMetrics.java
ericwush: Good catch, I promise I had this thought, until I forgot to do it -_-||

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -197,12 +198,15 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "metrics_type", "how maxwell metrics will be reported, at least one of slf4j|jmx|http|datadog" ).withOptionalArg();
 		parser.accepts( "metrics_slf4j_interval", "the frequency metrics are emitted to the log, in seconds, when slf4j reporting is configured" ).withOptionalArg();
 		parser.accepts( "metrics_http_port", "the port the server will bind to when http reporting is configured" ).withOptionalArg();
```
timbertson: Should change help text here to "[deprecated]"

src/main/java/com/zendesk/maxwell/monitoring/DiagnosticHealthCheck.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+
+public class DiagnosticHealthCheck extends HttpServlet {
+
+	private static final String CONTENT_TYPE = "text/plain";
```
liulikun: `application/json`?

ericwush: Good catch! 👍

PR #745, opened 2017-09-04
MELINF-114: make inflight message list bounded to back pressure binlog connector listener

ericwush:
@zendesk/goanna

### DIFF
--- a/src/test/java/com/zendesk/maxwell/producer/InflightMessageListTest.java
+++ b/src/test/java/com/zendesk/maxwell/producer/InflightMessageListTest.java
@@ -5,20 +5,25 @@ import com.zendesk.maxwell.replication.Position;
 import org.junit.Before;
 import org.junit.Test;
 
-import static org.junit.Assert.*;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
 
 /**
  * Created by ben on 5/25/16.
  */
 public class InflightMessageListTest {
+	static int capacity = 3;
 	static Position p1 = new Position(BinlogPosition.at(1, "f"), 0L);
 	static Position p2 = new Position(BinlogPosition.at(2, "f"), 0L);
 	static Position p3 = new Position(BinlogPosition.at(3, "f"), 0L);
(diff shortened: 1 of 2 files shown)


src/test/java/com/zendesk/maxwell/producer/InflightMessageListTest.java
```diff
@@ -54,4 +59,36 @@ public void testOutOfOrderComplete() {
...
+	public void testAddMessageWillWaitWhenCapacityIsFull() throws InterruptedException {
+		AddMessage addMessage = new AddMessage();
+		new Thread(addMessage).start();
+		assertThat("Should never exceed capacity", list.size(), lessThanOrEqualTo(capacity));
+
+		long wait = 500;
+		Thread.sleep(wait);
+		list.completeMessage(p1);
+
+		Thread.sleep(100);
```
timbertson: The first `sleep()` is unavoidable, but you could replace this one with a `join()` on the `AddMessage` thread.

src/test/java/com/zendesk/maxwell/producer/InflightMessageListTest.java
```diff
@@ -54,4 +59,36 @@ public void testOutOfOrderComplete() {
...
+		long start;
+		long end;
+
+		@Override
+		public void run() {
+			start = System.currentTimeMillis();
+			try {
+				list.addMessage(p4);
+			} catch (InterruptedException e) {
+				e.printStackTrace();
```
timbertson: Can you confirm that the test fails if this thread dies with (say) a RuntimeExecption? And if so, maybe just raise one here instead of logging, because this should never happen.

src/test/java/com/zendesk/maxwell/producer/InflightMessageListTest.java
ericwush: 👍 nice, good to know

src/test/java/com/zendesk/maxwell/producer/InflightMessageListTest.java
ericwush: If `InterruptedException` is thrown during `wait()`, the rest behaviour will be aborted which means it won't add the message. I've changed the assertion to check the list size should be exactly equal to the capacity to make sure the test will fail for this case. Any other exceptions will blow up the test anyways.

PR #746, opened 2017-09-04
Encryption

timbertson:
Building on #683


config.properties.example
```diff
@@ -173,6 +173,18 @@ kinesis_stream=maxwell
...
+
+# convert the data field to a json string and encrypt it
+#encrypt_data=true
+
+# convert the whole payload to a json string and encrypt it
+#encrypt_all=true
+
+# specify the encryption key and secret key
+#encryption_key=aaaaaaaaaaaaaaaa
+#secret_key=RandomInitVector
```
liulikun: outdated here

docs/docs/config.md
```diff
@@ -182,6 +182,10 @@ exclude_tables                 | PATTERN                             | ignore up
 blacklist_dbs                  | PATTERN                             | ignore updates AND schema changes from databases (see warnings below) |
 blacklist_tables               | PATTERN                             | ignore updates AND schema changes from tables named like PATTERN (see warnings below) |
 &nbsp;
+**encryption**
+encrypt_data                   | BOOLEAN                             | encrypt the data field of the payload               | false
+encrypt_all                    | BOOLEAN                             | encrypt the entire payload                          | false
```
liulikun: Another way to configure this: `encrypt=[data|all|none]`, default is `none`.

src/main/java/com/zendesk/maxwell/producer/MaxwellOutputConfig.java
```diff
@@ -4,7 +4,14 @@
 import java.util.ArrayList;
 import java.util.regex.Pattern;
 
+
 public class MaxwellOutputConfig {
+	public enum Encryption {
```
liulikun: `EncryptionMode`?

liulikun:
Looks good!

docs/docs/config.md
timbertson: Yeah that would probably be better, since you're not allowed to specify both anyway

liulikun:
👍

PR #747, opened 2017-09-05
MELINF-114: terminate maxwell if message is stuck in inflight message list

ericwush:
@zendesk/goanna 
- As discussed, we will only track publish+ack time for commit rows. 
- Please give feedback on the description in config example for `inflight_ack_timeout`. I want it to be neat and clear without exposing too many implementation details.

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -15,6 +15,12 @@ password=maxwell
 # choose where to produce data to. currently this is one of: stdout|file|kafka
 #producer=kafka
 
+# set delivery timeout (including acknowledgement) in milliseconds for a COMMIT row message that has been sent by the producer,
+# where producer is either kafka or kinesis
+# maxwell may be terminated if the timeout has expired, when majority of the COMMIT row messages have been acknowledged
+# set it to 0 to turn off this check
+#producer_ack_timeout=120000 # default 0
+
 # set the log level.  note that you can configure things further in log4j2.xml
 #log_level=DEBUG # [DEBUG, INFO, WARN, ERROR]
 
--- a/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
+++ b/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
@@ -6,6 +6,7 @@ package com.zendesk.maxwell.producer;
    Assumes .addInflight(position) will be call monotonically.
    */
 
+import com.zendesk.maxwell.MaxwellContext;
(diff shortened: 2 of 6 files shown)


config.properties.example
```diff
@@ -15,6 +15,12 @@ password=maxwell
 # choose where to produce data to. currently this is one of: stdout|file|kafka
 #producer=kafka
 
+# set delivery timeout (including acknowledgement) in milliseconds for a COMMIT row message that has been sent by the producer,
+# where producer is either kafka or kinesis
+# maxwell may be terminated if the timeout has expired, when majority of the COMMIT row messages have been acknowledged
+# set it to 0 to turn off this check
+#inflight_ack_timeout=120000 # default 0
```
liulikun: what about `producer_ack_timeout`?

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -86,4 +117,13 @@ public Position completeMessage(Position p) {
...
 		return linkedMap.size();
 	}
+
+	private double completePercentage() {
+		int completed = 0;
+		for (InflightMessage m : linkedMap.values()) {
+			if (m.isComplete)
+				completed ++;
+		}
+		return ((double) completed) / ((double) linkedMap.size());
```
liulikun: Only cast one number to `double` will do.

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -47,36 +62,52 @@ public void addMessage(Position p) throws InterruptedException {
...
+				Iterator<InflightMessage> iterator = this.linkedMap.values().iterator();
 
-			while ( iterator.hasNext() ) {
-				InflightMessage msg = iterator.next();
-				if ( !msg.isComplete ) {
-					break;
+				while ( iterator.hasNext() ) {
+					InflightMessage msg = iterator.next();
+					if ( !msg.isComplete ) {
+						head = msg;
```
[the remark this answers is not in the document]
ericwush: You are correct, if all messages are completed in a single go, the `head` isn't reset at the moment. But it can be fixed by having `if (linkedMap.size() == 0)` check.
Without tracking the `head`, we won't have `if (p.equals(head.position))` any more so the 1st `linkedMap.values().iterator()` will run every time. I dislike repeating the code/behaviour that's why having the `head` in the first place. What do you think?

PR #752, opened 2017-09-07
Add hook "onReplicatorEnd"

nikhilunni:
We've been using Maxwell in the "embedded" mode. We'd like a hook for when the
replicator thread dies, for whatever reason, to allow us to clean up our own
resources as well as alert necessary parties that Maxwell is no longer
processing the binlog.


liulikun:
👍

PR #754, opened 2017-09-07
Add initial Google Cloud Pub/Sub Producer

k4y3ff:
Dearest Zendesk friends,

I was hoping to use Maxwell with Pub/Sub! I noticed https://github.com/zendesk/maxwell/pull/296 from last year, which was never accepted; here's a PR that's largely copy-pasted from the current Kafka producer and uses the new Java library for Pub/Sub.

I've been able to compile/test locally, but I'm having trouble actually building the `.tar.gz` and such so I can run it in dev.

Have any feedback or hot takes?

 dog waving from car window

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
@@ -1,0 +1,208 @@
+package com.zendesk.maxwell.producer;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Meter;
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutureCallback;
+import com.google.cloud.pubsub.v1.Publisher;
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.PubsubMessage;
+import com.google.pubsub.v1.TopicName;
+import com.zendesk.maxwell.MaxwellContext;
+import com.zendesk.maxwell.monitoring.Metrics;
+import com.zendesk.maxwell.replication.Position;
+import com.zendesk.maxwell.row.RowMap;
+import com.zendesk.maxwell.schema.ddl.DDLMap;
+import com.zendesk.maxwell.util.Logging;
+import com.zendesk.maxwell.util.StoppableTask;
+import com.zendesk.maxwell.util.StoppableTaskState;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Properties;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+class PubsubCallback implements ApiFutureCallback<String> {
+  public static final Logger LOGGER = LoggerFactory.getLogger(PubsubCallback.class);
+
+  private final AbstractAsyncProducer.CallbackCompleter cc;
+  private final Position position;
+  private final String json;
+  private MaxwellContext context;
+
+  private Counter succeededMessageCount;
+  private Counter failedMessageCount;
+  private Meter succeededMessageMeter;
+  private Meter failedMessageMeter;
+
+  public PubsubCallback(AbstractAsyncProducer.CallbackCompleter cc,
+                        Position position, String json,
+                        Counter producedMessageCount, Counter failedMessageCount,
+                        Meter producedMessageMeter, Meter failedMessageMeter,
+                        MaxwellContext context) {
+    this.cc = cc;
+    this.position = position;
+    this.json = json;
+    this.succeededMessageCount = producedMessageCount;
+    this.failedMessageCount = failedMessageCount;
+    this.succeededMessageMeter = succeededMessageMeter;
+    this.failedMessageMeter = failedMessageMeter;
+    this.context = context;
+  }
+
+  @Override
+  public void onSuccess(String messageId) {
+    this.succeededMessageCount.inc();
+    this.succeededMessageMeter.mark();
+
+    if ( LOGGER.isDebugEnabled() ) {
+      LOGGER.debug("->  " + this.json);
+      LOGGER.debug("    " + this.position);
+      LOGGER.debug("");
+    }
+
+    cc.markCompleted();
+  }
+
+  @Override
+  public void onFailure(Throwable t) {
+    this.failedMessageCount.inc();
+    this.failedMessageMeter.mark();
+
+    LOGGER.error(t.getClass().getSimpleName() + " @ " + position);
+    LOGGER.error(t.getLocalizedMessage());
+
+    if ( !this.context.getConfig().ignoreProducerError ) {
+      this.context.terminate(new RuntimeException(t));
+      return;
+    }
+
+    cc.markCompleted();
+  }
+}
+
+public class MaxwellPubsubProducer extends AbstractProducer {
+  public static final Logger LOGGER = LoggerFactory.getLogger(MaxwellPubsubProducer.class);
+
+  private final ArrayBlockingQueue<RowMap> queue;
+  private final MaxwellPubsubProducerWorker worker;
+
+  public MaxwellPubsubProducer(MaxwellContext context, String pubsubProjectId,
+                               String pubsubTopic, String ddlPubsubTopic)
+                               throws IOException {
+    super(context);
+    this.queue = new ArrayBlockingQueue<>(100);
+    this.worker = new MaxwellPubsubProducerWorker(context, pubsubProjectId,
+                                                  pubsubTopic, ddlPubsubTopic,
+                                                  this.queue);
+    Thread thread = new Thread(this.worker, "maxwell-pubsub-worker");
+    thread.setDaemon(true);
+    thread.start();
+  }
+
+  @Override
+  public void push(RowMap r) throws Exception {
+    this.queue.put(r);
+  }
+
+  @Override
+  public StoppableTask getStoppableTask() {
+    return this.worker;
+  }
+}
+
+class MaxwellPubsubProducerWorker
+    extends AbstractAsyncProducer implements Runnable, StoppableTask {
+  static final Logger LOGGER = LoggerFactory.getLogger(MaxwellPubsubProducerWorker.class);
+
+  private final String projectId;
+  private Publisher pubsub;
+  private final TopicName topic;
+  private final TopicName ddlTopic;
+  private Publisher ddlPubsub;
+  private final ArrayBlockingQueue<RowMap> queue;
+  private Thread thread;
+  private StoppableTaskState taskState;
+
+  public MaxwellPubsubProducerWorker(MaxwellContext context,
+                                     String pubsubProjectId, String pubsubTopic,
+                                     String ddlPubsubTopic,
+                                     ArrayBlockingQueue<RowMap> queue)
+                                     throws IOException {
+    super(context);
+
+    this.projectId = pubsubProjectId;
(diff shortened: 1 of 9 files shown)


ryankennedy:
Looks like an issue with the `grpc` dependency (you can see it in [the build failure](https://travis-ci.org/zendesk/maxwell/jobs/273062559) on Travis):

> build/maven_fetcher:202:in `get': couldn't fetch io/grpc/grpc-core/[1.4.0]/grpc-core-[1.4.0].pom (RuntimeError)
> 	from build/maven_fetcher:300:in `get_file_from_repo'
> 	from build/maven_fetcher:295:in `get_pom'
> 	from build/maven_fetcher:224:in `block in resolve'
> 	from /home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/set.rb:324:in `each_key'
> 	from /home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/set.rb:324:in `each'
> 	from build/maven_fetcher:223:in `map'
> 	from build/maven_fetcher:223:in `resolve'
> 	from build/maven_fetcher:241:in `resolve'
> 	from build/maven_fetcher:241:in `resolve'
> 	from build/maven_fetcher:269:in `fetch_pom_deps'
> 	from build/maven_fetcher:358:in `<main\>'

Building master isn't currently throwing this error. So I'm guessing this dependency has been added as of your changes. That makes sense since your changes include the pubsub client dependency, which is going to be using grpc. Running `mvn dependency:tree` verifies that this is the case…

> [INFO] +- com.google.cloud:google-cloud-pubsub:jar:0.22.0-beta:compile
> [INFO] |  +- io.grpc:grpc-netty:jar:1.4.0:compile
> [INFO] |  |  +- io.grpc:grpc-core:jar:1.4.0:compile (version selected from constraint [1.4.0,1.4.0])

`io.grpc:grpc-core:1.4.0` [is definitely on Maven Central](http://search.maven.org/#artifactdetails%7Cio.grpc%7Cgrpc-core%7C1.4.0%7Cjar), though.

Rereading the output, I strongly suspect what's happening is that [build/maven_fetcher](https://github.com/zendesk/maxwell/blob/f36c53f31b83eedcd7fb88c65791203f7e728ebc/build/maven_fetcher) isn't handling the version constraint (`\[1.4.0\]`) properly when generating the URL to download from Maven:

> build/maven_fetcher:202:in `get': couldn't fetch io/grpc/grpc-core/[1.4.0]/grpc-core-[1.4.0].pom (RuntimeError)

ryankennedy:
I wonder why we use this instead of `mvn dependency:copy-dependencies`.

🤔

k4y3ff:
@ryankennedy: Noticed https://github.com/zendesk/maxwell/issues/756--anything I can help with in the meantime?

ryankennedy:
@k4y3ff resolving #756 should resolve the build problem you're encountering. So if you have any interest in taking that on, have at it. Although, I started pulling the thread a little bit and noticed that it's really tied up in how builds get produced so they can support multiple versions of Kafka (it's actually quite dangerous, IMO, how it's being done). So taking it on may be biting off more than you'd like.

Alternatively, if you want to get the build passing on this PR without addressing the entirety of #756 you could try modifying the POM file to exclude the io.grpc:grpc-core sub-dependency from the com.google.cloud:google-cloud-pubsub dependency and adding a direct dependency on io.grpc:grpc-core:1.4.0:

```xml
<dependencies>
    <dependency>
        <groupId>com.google.cloud</groupId>
        <artifactId>google-cloud-pubsub</artifactId>
        <version>0.22.0-beta</version>
        <!-- temporary until #756 is resolved -->
        <exclusions>
            <exclusion>
                <groupId>io.grpc</groupId>
                <artifactId>grpc-core</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
    <!-- temporary until #756 is resolved -->
    <dependency>
        <groupId>io.grpc</groupId>
        <artifactId>grpc-core</artifactId>
        <version>1.4.0</version>
    </dependency>
</dependencies>
```

I think that would sidestep the build/maven_fetcher problem in this case, which might at least get this PR to turn green in TravisCI to get to the next step of evaluation.

Thoughts?

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
```diff
@@ -0,0 +1,225 @@
...
+      try {
+        ddlPubsub.shutdown();
+      } catch (Exception e) {
+        LOGGER.error("Failed to shutdown Pubsub publisher: " + e.getLocalizedMessage());
+        LOGGER.error("    projectId: " + projectId);
+        LOGGER.error("    topic: " + ddlTopic.getTopic());
+      }
+    }
+  }
+
```
timbertson: `requestStop` will be allowed to throw an exception once #742 is merged - are you happy if we later take out these catches and just propagate the error?

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
```diff
@@ -0,0 +1,225 @@
...
+    this.worker.setDDLPubsub();
+    Thread thread = new Thread(this.worker, "maxwell-pubsub-worker");
+    thread.setDaemon(true);
+    thread.start();
+  }
+
+  @Override
+  public void push(RowMap r) throws Exception {
+    this.queue.put(r);
+  }
```
timbertson: The main reason we use an intermediate queue in the kafka producer is that kafka's `send()` method can block the current thread (performing compression I believe) even though it's otherwise async.

If pubsub producer is properly async and doesn't block the sending thread at all, you could probably make this producer a bit simpler by avoiding the queue (and the worker thread), like the Kinesis producer.

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
```diff
@@ -0,0 +1,225 @@
...
+  private StoppableTaskState taskState;
+
+  public MaxwellPubsubProducerWorker(MaxwellContext context, Properties pubsubProperties, String pubsubProjectId, String pubsubTopic, ArrayBlockingQueue<RowMap> queue) {
+    super(context);
+
+    if ( pubsubTopic == null ) {
+      this.topic = TopicName.create(pubsubProjectId, "maxwell");
+    } else {
+      this.topic = TopicName.create(pubsubProjectId, pubsubTopic);
+    }
```
liulikun: `pubsubTopic`  won't be `null` if we set the default topic to `maxwell`.

k4y3ff: Ah! So,

1. Your comment has pointed out to me that I left the `fetchOption` lines out of this PR, just updated accordingly.

2. The Kafka producer has a default topic of `maxwell` but still checks for a `null` value, anyway--do you still want me to remove this? If so, should probably remove it from the Kafka producer, too.

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
```diff
@@ -0,0 +1,225 @@
...
+
+    this.projectId = pubsubProjectId;
+    this.ddlTopic = TopicName.create(pubsubProjectId, context.getConfig().ddlPubsubTopic);
+
+    Metrics metrics = context.getMetrics();
+    this.metricsTimer = metrics.getRegistry().timer(metrics.metricName("message", "publish", "time"));
```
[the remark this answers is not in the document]
k4y3ff: Done!

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
k4y3ff: Yeah! Also just noticed you've since merged #742, so removed the catches now.

k4y3ff:
@ryankennedy: Can I take the diplomatic stance, which is that Zendesk owns the project, y'all might have introduced internal Maven project style guides since this project was written, and I'm inclined to let y'all (or us, collectively) fix it in a different PR so that those style guides can be respected? 😂 

Also: got the build mostly passing, but the `GTID_MODE=1 MYSQL_VERSION=5.7` job [failed with](https://travis-ci.org/zendesk/maxwell/jobs/274689382#L643)

```
java.lang.AssertionError: Should have waited message to be completed
Expected: a value equal to or greater than <500L>
     but: <497L> was less than <500L>
```

Given that it's pretty close, only in one of the jobs, and referencing a part of the code I didn't touch: do you know whether the test is nondeterministic and fine, or a thing I should also go poke at? ¯\_(ツ)_/¯

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
timbertson: Yup, either is fine - we can always simplify later if someone does the testing to figure out whether it affects throughput 👍

timbertson:
> I'm inclined to let y'all (or us, collectively) fix it in a different PR

Yeah, if we can workaround the maven fetcher for this PR (looks like you've done this already) then we'll fix that separately :). If you want to help that's great, but it's certainly not a requirement.

> do you know whether the test is nondeterministic and fine, or a thing I should also go poke at

It's a known flakey test, unfortunately. I've kicked it off again and it's passing now.

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
```diff
@@ -0,0 +1,212 @@
...
+    this.taskState = new StoppableTaskState("MaxwellPubsubProducerWorker");
+  }
+
+  public void setPubsub() throws IOException {
+    this.pubsub = Publisher.defaultBuilder(this.topic).build();
+  }
+
+  public void setDDLPubsub() throws IOException {
+    this.ddlPubsub = Publisher.defaultBuilder(this.ddlTopic).build();
+  }
```
timbertson: Can these be private / inlined now? (I'm not sure I understand the intent here, as they don't allow any control over what instance gets created)

timbertson: Also, if the creation of a publisher causes some external action (e.g. connecting to a broker or validating the topic name exists), we shouldn't set one up unless `context.getConfig().outputDDL == true`

k4y3ff: Welp! Yep, updated.

src/main/java/com/zendesk/maxwell/producer/MaxwellPubsubProducer.java
liulikun: I think Kafka producer code is also having the same issue. If you can clean up here that'll be great. We can cleanup Kafka producer code later or you can clean it up if you don't mind.

k4y3ff:
> Yeah, if we can workaround the maven fetcher for this PR (looks like you've done this already) then we'll fix that separately :). If you want to help that's great, but it's certainly not a requirement.

Well, we might have to fix it first, after all. 😂 

Apparently need to add an explicit dependency on `google-auth-library-java` based on the current packaging process, but unfortunately `maven_fetcher` doesn't like that, either:

```
build/maven_fetcher:102:in `block in subst': unknown variable: project.google.http.version (RuntimeError)
```

timbertson:
🎉 thanks for your assistance sorting out the maven issues - looks like it needs a rebase, then we should be good to go :)

k4y3ff:
So, uh... in true open-source PR fashion, I went traveling! And then to a conference! And now I have returned, with a PR that has been rebased, passed CI, and run with some success. 😂 💞

heart hands

Care to take another look?

PR #761, opened 2017-09-13
Fix NPE for tracking publish time

ericwush:
/cc @zendesk/goanna

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
@@ -1,13 +1,12 @@
 package com.zendesk.maxwell.producer;
 
-import com.codahale.metrics.Counter;
-import com.codahale.metrics.Gauge;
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
```diff
@@ -36,14 +34,12 @@ public void markCompleted() {
 
 				if (message != null) {
 					context.setPosition(message.position);
-					timeSinceSendMS = message.timeSinceSendMS();
+					Metrics metrics = context.getMetrics();
+					Timer timer = metrics.getRegistry().timer(metrics.metricName("message", "publish", "time"));
```
liulikun: We'd be able to reuse the same `timer` here.

ericwush: What do you mean? Can you elaborate pls? The code is actually moved from `MaxwellKafkaProducer`.

Issue #756, opened 2017-09-08
build/maven_fetcher can't handle maven version range references


ryankennedy:
As seen in #754, build/maven_fetcher is failing when one of the resolved maven dependencies includes a maven [version range reference](https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN402).

```
build/maven_fetcher:202:in `get': couldn't fetch io/grpc/grpc-core/[1.4.0]/grpc-core-[1.4.0].pom (RuntimeError)
	from build/maven_fetcher:300:in `get_file_from_repo'
	from build/maven_fetcher:295:in `get_pom'
	from build/maven_fetcher:224:in `block in resolve'
	from /home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/set.rb:324:in `each_key'
	from /home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/set.rb:324:in `each'
	from build/maven_fetcher:223:in `map'
	from build/maven_fetcher:223:in `resolve'
	from build/maven_fetcher:241:in `resolve'
	from build/maven_fetcher:241:in `resolve'
	from build/maven_fetcher:269:in `fetch_pom_deps'
	from build/maven_fetcher:358:in `<main>'
```

It appears that the maven download URL isn't resolving the version range reference to an individual version number and simply dumps the raw version range reference into the URL.

Should we switch to letting maven itself download the dependencies? [mvn dependency:copy-dependencies](https://maven.apache.org/plugins/maven-dependency-plugin/examples/copying-project-dependencies.html) was built precisely to do this and will handle resolving version range references without duplicating the logic in build/maven_fetcher.

timbertson:
Seems reasonable to let maven do its thing...

@osheroff, do you know if there are particular features `maven_fetcher` was built for that aren't supported by mvn?

osheroff:
Nah I wrote it in a fit of rage trying to get compile times under control; maven adds like 7 seconds to each damn step.  If you wanna rip it out feel free.

timbertson:
Fixed in #765

PR #763, opened 2017-09-18
Switch to using Maven for builds

k4y3ff:
ref: #756 

Removes all the `maven_fetcher` logic I could find and replaces it with Maven.

This Works on My Machine™. ¯\_(ツ)_/¯ ᕕ( ᐛ )ᕗ


k4y3ff:
The build has passed! 🎉

timbertson:
Thanks @k4y3ff ! I took a look, it's a great start. The main issue I can see is that the packaged tar only includes one version of kafka-clients, whereas our existing release tars bundle multiple versions and switches which one's included on the classpath via a commandline flag in `bin/maxwell`.

I'm happy to have a go fixing that on top of this PR if you like?

k4y3ff:
Sure, go to town!

PR #764, opened 2017-09-19
Add virtual hosts to RabbitMQ producer

nebez:
This adds the ability to set virtual hosts to the RabbitMQ producer other than the default of `/`.
Also add ability to make the exchange durable or not. Defaults to false.

Both of these options use the defaults so if you don't use them this won't change your producers behaviour at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 </div>
 
 This is Maxwell's daemon, an application that reads MySQL binlogs and writes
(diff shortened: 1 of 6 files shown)


nebez:
Also, is there any chance to get maxwell tagged after this? Currently using my own dockerhub repository to get it going.

ericwush:
@nebez thanks for this. You'd probably want to update `config.properties.example` file to reflect the changes.

nebez:
@ericwush - updated the PR. Does this look better?

PR #765, opened 2017-09-19
Switch to using Maven for builds (#2)

timbertson:
Builds on https://github.com/zendesk/maxwell/pull/763. I've fixed up the archiver stage to output the same files as the current release tarball, and reinstated the `--kafka_version` flag in bin/maxwell with some minor changes (kafka-clients jars live in their own directory now).

It's still a bit dodgy to smush together the dependencies for multiple kafka-clients versions and hope that nothing goes wrong at runtime, we will need to do something different if versions of this jar ever differ significantly in their dependencies.


Makefile
```diff
@@ -1,133 +1,24 @@
...
-
-define get_test_parameters
-	$(eval filter=$(shell echo $(1) | perl -ne '/#(.*)/; print "--filter=com.zendesk.maxwell.JUnitNameFilterFactory=$$1" if $$1')) \
-	$(eval class_pattern=$(shell echo $(1) | perl -ne '/test\.(\w+)/; print $$1 if $$1')) \
-	$(eval classes=$(if $(class_pattern),$(filter %$(class_pattern),$(TEST_CLASSES)),$(TEST_CLASSES))) \
-	$(filter) $(classes)
-endef
-
-JAVA_TEST_PARAMS=-Djava.util.logging.config.file=./src/test/resources/logging.properties
+all: compile
```
liulikun: Does this include antlr compilation?

Makefile
```diff
@@ -1,133 +1,24 @@
...
-	cp $(DISTFILES) $(TARDIR)
-	cp -a bin $(TARDIR)
-	mkdir $(TARDIR)/lib
-	cp -a $(MAXWELL_JARFILE) target/dependency-build/* $(TARDIR)/lib
-	tar czvf $(TARFILE) -C target $(PKGNAME)
-
-package: depclean package-jar package-tar
 
+package: depclean
+	@# TODO: this is inefficient, we really just want to copy the jars...
```
liulikun: What if we depend on these versions in `dependencies`?

timbertson: I tried that - maven complains if you depend on multiple versions of a dependency, and picks the latest. Which is usually what you want, but we're doing something weird here on purpose.

Makefile
timbertson: Yep - I did a full (git) clean, and compile still works. Looks like we already have the antlr maven plugin, we were just replicating it in the Makefile...

> [INFO] --- antlr4-maven-plugin:4.5:antlr4 (antlr) @ maxwell ---
[INFO] ANTLR 4: Processing source directory /Users/tcuthbertson/Code/zendesk/maxwell/src/main/antlr4
[INFO] Processing grammar: com/zendesk/maxwell/schema/ddl/mysql.g4

liulikun: 👍

liulikun:
Looks good! 👍

PR #773, opened 2017-10-05
Add Timeout binary log client into configuration

kukuhpro:
I add this feature because in my country, internet connection is not good and i ended timeout binary client after 5000ms, so i make connection timeout for binary client to configuration, it will give default value for 5000ms, but it can be configured by set those value on config.properties.

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -4,6 +4,9 @@ log_level=info
 producer=kafka
 kafka.bootstrap.servers=localhost:9092
 
+Set Timeout for connect to binary client, default is 5000ms
+#timeout_binary_log_client=50000
+
 # mysql login info
 host=localhost
 user=maxwell
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -40,6 +40,8 @@ public class BinlogConnectorReplicator extends AbstractReplicator implements Rep
 	private boolean hitEOF = false;
 	private Histogram transactionRowCount;
 	private Histogram transactionExecutionTime;
+	private MaxwellMysqlConfig mysqlConfig;
+	private long timeoutBinaryLogClient;
 
 	public BinlogConnectorReplicator(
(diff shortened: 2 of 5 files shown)


config.properties.example
```diff
@@ -4,6 +4,9 @@ log_level=info
 producer=kafka
 kafka.bootstrap.servers=localhost:9092
 
+Set Timeout for connect to binary client, default is 5000ms
+#timeout_binary_log_client=50000
```
liulikun: To be exact, the timeout is used to connect to MySQL server to establish replication stream. I'm thinking `replication_connection_timeout` might be clearer.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -40,6 +40,8 @@
 	private boolean hitEOF = false;
 	private Histogram transactionRowCount;
 	private Histogram transactionExecutionTime;
+	private MaxwellMysqlConfig mysqlConfig;
```
liulikun: Don't think we need this field.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -56,6 +58,7 @@ public BinlogConnectorReplicator(
 	) {
 		super(clientID, bootstrapper, maxwellSchemaDatabaseName, producer, metrics, start, heartbeatNotifier);
 		this.schemaStore = schemaStore;
+		this.mysqlConfig = mysqlConfig;
```
liulikun: We can just set the timeout field here.

liulikun:
CI failed due to the introduction of this config. The PR looks good apart from the nitpicks.

kukuhpro:
@liulikun I commit a new change based on your comment, please do a review, just comment is there is anything need to change about this pull request. Thanks.

liulikun:
Cool. Can you take a look at failed tests? Let me know if you need any help.

ericwush:
The build still failed. You can check the travis logs or rerun the tests locally.

PR #777, opened 2017-10-11
Fix docker build

danmactough:
The docker build is broken. 😭

image

This fixes the Dockerfile. The main problem was that maven was missing.

It also slightly modifies the Makefile to enable separately building the different packages for different targeted kafka versions. It's not needed to fix the docker build, and I can pull it out if you prefer.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,18 @@
 FROM java:openjdk-8
-ENV MAXWELL_VERSION 1.10.7
-ENV KAFKA_VERSION 0.10.1.0
+ENV MAXWELL_VERSION=1.10.7 KAFKA_VERSION=0.10.1.0
 
-RUN apt-get update && apt-get -y upgrade
-
-RUN apt-get install build-essential ruby -y
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get -y install build-essential maven
 
 COPY . /workspace
-WORKDIR /workspace
 
-RUN KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION
-
-WORKDIR /workspace/target/maxwell-$MAXWELL_VERSION/
-
-RUN mkdir /app \
-  && mv ./* /app/
+RUN cd /workspace \
+    && KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION \
+    && mkdir /app \
+    && mv /workspace/target/maxwell-$MAXWELL_VERSION/maxwell-$MAXWELL_VERSION/* /app/ \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/
 
 WORKDIR /app
 
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,9 @@ clean:
 depclean: clean
 	rm -f $(CLASSPATH)
 
-package: depclean
+package: depclean kafka-0.8.2.2 kafka-0.9.0.1 kafka-0.10.0.1 kafka-0.10.2.1
 	@# TODO: this is inefficient, we really just want to copy the jars...
-	mvn compile -P kafka-0.8.2.2
-	mvn compile -P kafka-0.9.0.1
-	mvn compile -P kafka-0.10.0.1
-	mvn compile -P kafka-0.10.2.1
 	mvn package -DskipTests=true
+
+kafka-%:
+	mvn compile -P package-kafka-$(*)


Dockerfile
```diff
@@ -2,21 +2,18 @@ FROM java:openjdk-8
 ENV MAXWELL_VERSION 1.10.7
 ENV KAFKA_VERSION 0.10.1.0
 
-RUN apt-get update && apt-get -y upgrade
-
-RUN apt-get install build-essential ruby -y
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get install build-essential maven ruby -y
```
nebez: Is `ruby` necessary anymore with the introduction of `maven`? I've removed it from my forked maxwell Dockerfile entirely.

nebez:
Thanks for fixing it!
I've managed to reduce the steps even further by combining the related ones together + removing unnecessary steps in my own fork.

This is what I'm working with now:
```Dockerfile
FROM java:openjdk-8
ENV MAXWELL_VERSION 1.10.6
ENV KAFKA_VERSION 0.10.1.0

RUN apt-get update \
    && apt-get -y upgrade \
    && apt-get install -y build-essential maven

COPY . /workspace

RUN cd /workspace \
    && KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION \
    && mkdir /app \
    && mv /workspace/target/maxwell-$MAXWELL_VERSION/maxwell-$MAXWELL_VERSION/* /app/ \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/

WORKDIR /app

RUN echo "$MAXWELL_VERSION" > /REVISION
CMD bin/maxwell --user=$MYSQL_USERNAME --password=$MYSQL_PASSWORD --host=$MYSQL_HOST --producer=kafka --kafka.bootstrap.servers=$KAFKA_HOST:$KAFKA_PORT $MAXWELL_OPTIONS
```

What do you think?

Dockerfile
danmactough: @nebez Good call. I think ruby is only needed for the release script. Running a build now without it 🤞

danmactough:
> What do you think?

@nebez Took your suggestions. Fewer layers is 👍  although sadly all the cleanup didn't reduce the image size very much: 1.47GB -> 1.31GB.

nebez:
I couldn't think of much else to shrink the image, but to be honest I didn't put a lot of effort into investigating further. If you have a better cleanup, I'd love to incorporate it into my images.

danmactough:
No, I agree! There are a ton of jars!

Dockerfile
```diff
@@ -1,20 +1,18 @@
...
-
-WORKDIR /workspace/target/maxwell-$MAXWELL_VERSION/
 
-RUN mkdir /app \
-  && mv ./* /app/
+RUN cd /workspace \
+    && KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION \
+    && mkdir /app \
+    && mv /workspace/target/maxwell-$MAXWELL_VERSION/maxwell-$MAXWELL_VERSION/* /app/ \
+    && apt-get clean \
```
xuebingli: `apt-get clean` here doesn't reduce size, since the files have been generated (and cached) in the previous layer.

Dockerfile
```diff
@@ -1,20 +1,18 @@
...
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/
```
xuebingli: Ditto.

nebez:
Good catch, @xuebingli - you're absolutely right.

I've got it down to ~960MB with this Dockerfile instead:
```Dockerfile
FROM maven:3.5-jdk-8
ENV MAXWELL_VERSION 1.10.7
ENV KAFKA_VERSION 0.10.1.0

COPY . /workspace

RUN apt-get update \
    && apt-get install -y build-essential \
    && cd /workspace \
    && KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION \
    && mkdir /app \
    && mv /workspace/target/maxwell-$MAXWELL_VERSION/maxwell-$MAXWELL_VERSION/* /app/ \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/

WORKDIR /app

RUN echo "$MAXWELL_VERSION" > /REVISION
CMD bin/maxwell --user=$MYSQL_USERNAME --password=$MYSQL_PASSWORD --host=$MYSQL_HOST --producer=kafka --kafka.bootstrap.servers=$KAFKA_HOST:$KAFKA_PORT $MAXWELL_OPTIONS
```

Still rather large but ~400MB less is better than nothing.

danmactough:
@xuebingli makes sense. Let's restrict this to fixing. I can pull out the ineffective attempts to diminish image size to keep this focused.

xuebingli:
@nebez great! To be honest I didn't realize Maxwell's docker image was at ~1.5GB until today.

@danmactough agreed. Thanks again for the fix!

danmactough:
image

PR #784, opened 2017-10-13
Adding autodelete behavior for rabbitmq exchanges

mortensen:
Adding possibility to configure autodelete behavior for rabbitmq exchanges.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/RabbitmqProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/RabbitmqProducer.java
@@ -26,7 +26,7 @@ public class RabbitmqProducer extends AbstractProducer {
 		factory.setVirtualHost(context.getConfig().rabbitmqVirtualHost);
 		try {
 			this.channel = factory.newConnection().createChannel();
-			this.channel.exchangeDeclare(exchangeName, context.getConfig().rabbitmqExchangeType, context.getConfig().rabbitMqExchangeDurable);
+			this.channel.exchangeDeclare(exchangeName, context.getConfig().rabbitmqExchangeType, context.getConfig().rabbitMqExchangeDurable, context.getConfig().rabbitMqExchangeAutoDelete, null);
 		} catch (IOException | TimeoutException e) {
 			throw new RuntimeException(e);
 		}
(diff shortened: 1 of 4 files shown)


mortensen:
wtf? is master broke right now? i just added a property.

nebez:
This is a backwards incompatible change.

If we look at the overload that is currently being used:
https://github.com/rabbitmq/rabbitmq-java-client/blob/8327c6aa4d58780c18a8d56e90a64ca837846560/src/main/java/com/rabbitmq/client/impl/ChannelN.java#L787-L794
The `autodelete` argument is set to false.

If you change the default from `true` to `false`, it'll work fine.

mortensen:
ok thanks for the hint. i chose true as the default, because the rabbitmq interface by default uses true. but as long as its configurable it can sure also be false, thanks. fixed it.

src/main/java/com/zendesk/maxwell/producer/RabbitmqProducer.java
```diff
@@ -26,7 +26,7 @@ public RabbitmqProducer(MaxwellContext context) {
 		factory.setVirtualHost(context.getConfig().rabbitmqVirtualHost);
 		try {
 			this.channel = factory.newConnection().createChannel();
-			this.channel.exchangeDeclare(exchangeName, context.getConfig().rabbitmqExchangeType, context.getConfig().rabbitMqExchangeDurable);
+			this.channel.exchangeDeclare(exchangeName, context.getConfig().rabbitmqExchangeType, context.getConfig().rabbitMqExchangeDurable, context.getConfig().rabbitMqExchangeAutoDelete, null);
```
liulikun: Have you checked if `null` is safe here? Should we use an empty `Map`?

mortensen: The classes ChannelN and AutorecoveringChannel have several overloaded variants of this method exchangeDeclare. Before my change the arguments were not given. In case the map was not given as parameter, they always call another overload with null. Also when you check AMQImpl, there is following snippet:

 this.arguments = arguments==null ? null : Collections.unmodifiableMap(new HashMap<String,Object>(arguments));

To me it seems quite safe to use null here.

liulikun:
Looks good! 👍

PR #788, opened 2017-10-23
Reduce Docker image size

nebez:
This reduces the image size from ~1.47GB to ~960MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,19 @@
-FROM java:openjdk-8
+FROM maven:3.5-jdk-8
 ENV MAXWELL_VERSION=1.10.8 KAFKA_VERSION=0.10.1.0
-
-RUN apt-get update \
-    && apt-get -y upgrade \
-    && apt-get -y install build-essential maven
 
 COPY . /workspace
 
-RUN cd /workspace \
+RUN apt-get update \
+    && apt-get upgrade \
+    && apt-get install -y build-essential \
+    && cd /workspace \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,20 +1,19 @@
...
+    && apt-get upgrade \
+    && apt-get install -y build-essential \
+    && cd /workspace \
     && KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION \
     && mkdir /app \
     && mv /workspace/target/maxwell-$MAXWELL_VERSION/maxwell-$MAXWELL_VERSION/* /app/ \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/ \
+    && echo "$MAXWELL_VERSION" > /REVISION
```
xuebingli: I reckon `echo "$MAXWELL_VERSION" > /REVISION` need to be behind `WORKDIR /app`, otherwise `REVISION ` would be created in `/workspace` instead of `/app`.

nebez: @xuebingli as it's an absolute path, the current working directory doesn't impact the destination. It will always be written at the root `/` directory.

xuebingli: My bad. I should've read it more closely.

PR #790, opened 2017-10-26
[MELINF-157] Listen to binlog connector lifecycle events

xuebingli:
@zendesk/goanna 

- `BinlogConnectorLifecycleListener` is splitted out of `BinlogConnectorEventListener`, and is registered to the binlog connector to listen for lifecycle events.
- `binlogEventListener` is changed from `protected` to `private`

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorLifecycleListener.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorLifecycleListener.java
@@ -4,16 +4,8 @@ import com.github.shyiko.mysql.binlog.BinaryLogClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
 class BinlogConnectorLifecycleListener implements BinaryLogClient.LifecycleListener {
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorLifecycleListener.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+class BinlogConnectorLifecycleListener implements BinaryLogClient.LifecycleListener {
+	private static final Logger LOGGER = LoggerFactory.getLogger(BinlogConnectorLifecycleListener.class);
+
+	protected final AtomicBoolean mustStop = new AtomicBoolean(false);
+
+	public void stop() {
+		mustStop.set(true);
+	}
```
liulikun: Don't think we need this.

xuebingli: This method is used [here](https://github.com/zendesk/maxwell/pull/790/files#diff-6c0364725d8f72f2d70f68c6c6115747R128).

☝️  Can the usage be safely remove?

liulikun: Yup, this listener doesn't need to deal with `stop` request .

PR #796, opened 2017-10-30
Allow LOCALTIME and LOCALTIMESTAMP as default values

hexene:


hexene:
`LOCALTIME` and `LOCALTIMESTAMP` are synonyms for `NOW()`: https://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_localtime

hexene:
Assuming this change is OK, will add a test once #797 is merged to avoid a merge-conflict.

liulikun:
Yup, looks good!

timbertson:
#797 merged, fire away with tests :)

osheroff:
`mysql_literal_tokens.g4` is a file generated from the ruby script in the directory.  need to edit that file instead.

osheroff:
I added some tests and merged a rebased version of this.

PR #797, opened 2017-10-30
Allow default values with an explicit plus sign

hexene:


src/main/antlr4/imports/mysql_idents.g4
```diff
@@ -19,8 +19,8 @@ name: ( id | tokens_available_for_names | INTEGER_LITERAL);
 id: ( IDENT | QUOTED_IDENT );
 literal: (float_literal | integer_literal | string_literal | NULL | TRUE | FALSE);
 
-float_literal: '-'? INTEGER_LITERAL? '.' INTEGER_LITERAL;
-integer_literal: '-'? INTEGER_LITERAL;
+float_literal: ('+'|'-')? INTEGER_LITERAL? '.' INTEGER_LITERAL;
+integer_literal: ('+'|'-')? INTEGER_LITERAL;
```
liulikun: looks fine, can you add a test for this case? maybe in this file: https://github.com/zendesk/maxwell/blob/master/src/test/resources/sql/ddl/mysql-test.sql

hexene: I've added a test to mysql-test-fixed.sql, since those in mysql-test.sql aren't being executed directly. Hope this will do.

timbertson: Actually both `-fixed` and `-errors` are generated by `DDLParserTest.generateTestFiles`, but it's not automated (or particularly obvious). I've moved this example and regenerated in #818

hexene: Thank you, I'll close this PR.

PR #801, opened 2017-11-03
Add rabbitmq config option for persistent messages

d-babiak:
Adds RabbitMQ config option to publish messages using `MINIMAL_PERSISTENT_BASIC` properties instead of the `MINIMAL_BASIC` which `null` props defaults to.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -104,6 +104,7 @@ public class MaxwellConfig extends AbstractConfig {
 	public boolean rabbitMqExchangeDurable;
 	public boolean rabbitMqExchangeAutoDelete;
 	public String rabbitmqRoutingKeyTemplate;
+	public boolean rabbitmqMessagePersistent;
 
 	public String redisHost;
 	public int redisPort;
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -234,6 +235,7 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "rabbitmq_exchange_durable", "Exchange durability. Default is disabled" ).withOptionalArg();
 		parser.accepts( "rabbitmq_exchange_autodelete", "If set, the exchange is deleted when all queues have finished using it. Defaults to false" ).withOptionalArg();
 		parser.accepts( "rabbitmq_routing_key_template", "A string template for the routing key, '%db%' and '%table%' will be substituted. Default is '%db%.%table%'." ).withRequiredArg();
+		parser.accepts( "rabbitmq_message_persistent", "Message persistence. Defaults to false" ).withOptionalArg();
```
liulikun: here should be `withRequiredArg();`, meaning if you provide `--rabbitmq_message_persistent` option, you need to have a value for this option.

timbertson: Actually this is OK, we do usually use withOptionalArg for booleans

PR #806, opened 2017-11-14
Supported row query when binlog option "binlog_rows_query_log_events" is enable

saimon7:
Add full row query to output format when binlog option "binlog_rows_query_log_events" is enable and properties output_row_query is enable.

Example:
mysql> insert into test.maxwell set id = 1, daemon = 'Stanislaw Lem' /* comment */;

maxwell: {
"database": "test",
"table": "maxwell",
"type": "insert",
"ts": 1449786310,
"xid": 940752,
"commit": true,
"data": { "id":1, "daemon": "Stanislaw Lem" },
"query": "insert into test.maxwell set id = 1, daemon = 'Stanislaw Lem' /* comment */"
}


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -147,6 +145,8 @@ private RowMapBuffer getTransactionRows(BinlogConnectorEvent beginEvent) throws
 		BinlogConnectorEvent event;
 		RowMapBuffer buffer = new RowMapBuffer(MAX_TX_ELEMENTS);
 
+		String temporaryQuery = "";
```
timbertson: nit: maybe rename this `currentQuery`, since it's holding the query for the current update?

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -179,15 +179,19 @@ private RowMapBuffer getTransactionRows(BinlogConnectorEvent beginEvent) throws
...
+					temporaryQuery = "";
 					break;
 				case TABLE_MAP:
 					TableMapEventData data = event.tableMapData();
 					tableCache.processEvent(getSchema(), this.filter, data.getTableId(), data.getDatabase(), data.getTable());
 					break;
+				case ROWS_QUERY:
+					RowsQueryEventData rqed = event.getEvent().getData();
+					temporaryQuery = rqed.getQuery();
+					break;
```
timbertson: It looks like this code is in a loop which runs for every update in a transaction. If there are multiple queries in a transaction, does that mean we only end with the last query?

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -254,6 +261,11 @@ public String toJSON(MaxwellOutputConfig outputConfig) throws Exception {
 
 		g.writeStringField("database", this.database);
 		g.writeStringField("table", this.table);
+
+		if ( outputConfig.includesRowQuery ) {
+			g.writeStringField("query", this.rowQuery);
```
timbertson: I'm not sure about your use case, but barring a good reason I'd rather see `rowQuery` be `null` instead of an empty string when we don't have one, and this property just wouldn't be set in that case. Emitting `query: ""` seems a bit unnecessary.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
saimon7: You have right, `currentQuery` is much better.

src/main/java/com/zendesk/maxwell/row/RowMap.java
saimon7: Right, my mistake.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
saimon7: I suppose you mean the situation

<pre>
START TRANSACTION;
UPDATE `test` SET `field`='foo' WHERE `id`='1';
UPDATE `test` SET `field`='bar' WHERE `id`='1';
COMMIT;
</pre>

and in this case, there are two separate event in the binlog, not one.

src/main/java/com/zendesk/maxwell/row/RowMap.java
timbertson: You're still passing around `""` as the default rowQuery in a few places, this would need to be `null` instead to have it not appear in the output.

saimon7: Fixed.

src/test/java/com/zendesk/maxwell/MaxwellIntegrationTest.java
```diff
@@ -562,4 +563,51 @@ public void testSchemaServerNotSet() throws Exception {
...
+
+		expectedJSON.add(MaxwellTestJSON.parseJSON("{\"database\":\"test\",\"table\":\"row_query\",\"query\":\"" + sql[4]+ "\",\"type\":\"update\", \"data\":{id:11}, \"old\":{id:10}}"));
+		expectedJSON.add(MaxwellTestJSON.parseJSON("{\"database\":\"test\",\"table\":\"row_query\",\"query\":\"" + sql[5]+ "\",\"type\":\"update\", \"data\":{id:21}, \"old\":{id:20}}"));
+
+		expectedJSON.add(MaxwellTestJSON.parseJSON("{\"database\":\"test\",\"table\":\"row_query\",\"query\":\"" + sql[6]+ "\",\"type\":\"delete\", \"data\":{id:11}}"));
+		expectedJSON.add(MaxwellTestJSON.parseJSON("{\"database\":\"test\",\"table\":\"row_query\",\"query\":\"" + sql[7]+ "\",\"type\":\"delete\", \"data\":{id:21}}"));
+
+		MaxwellTestJSON.assertJSON(eventJSON, expectedJSON);
+
+	}
```
osheroff: this is a great place to use the integration test framework: see the files in `src/test/resources/sql/json`

saimon7: Changed as you propose

osheroff:
LGTM 👍  @timbertson if you want to take a look feel free, otherwise if I don't hear from you in a bit I'll merge

PR #808, opened 2017-11-15
Add Graylog producer

saimon7:
This is implementation of a producer for Graylog.

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -12,7 +12,7 @@ password=maxwell
 
 ######### general stuff #############
 
-# choose where to produce data to. currently this is one of: stdout|file|kafka|kinesis|pubsub|rabbitmq|redis
+# choose where to produce data to. currently this is one of: stdout|file|kafka|kinesis|pubsub|rabbitmq|redis|graylog
 #producer=kafka
 
 # set delivery timeout (including acknowledgement) in milliseconds for a COMMIT row message that has been sent by the producer,
@@ -168,6 +168,13 @@ kinesis_stream=maxwell
 #redis_auth=redis_auth
 #redis_database=0
 #redis_pub_channel=maxwell
+
+######### graylog ######################
+#graylog_host=graylog_host
+#graylog_port=12201
+#graylog_transport=udp
+#graylog_additional_field.field_foo=foo
+#graylog_additional_field.field_bar=bar
 
 ######### filter stuff ###############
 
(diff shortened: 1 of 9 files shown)


osheroff:
Code looks generally sane.  You may have to watch out for the case where people are running in binlog_row_image=minimal (for updates, we may not end up having both a before-and-after value).  

Could you provide an example of what the output format looks like?  I'm not familiar with the GELF format, but it looks like you're sending a new GELF message for each column?  would it maybe be better to send one message with each column as an additional field like:

``` 
  { ... timestamp, etc
    "_user_id": 553243, 
    "_user_id.old": 422231,
    "_foo_column": "bar"  
   }
```
 (I'm presuming GELF doesn't allow for nested data?)

saimon7:
I looked into Graylog documentation about additional fields and checked witch field name can be accepted. The problem is that field name allowed only: letter, number, underscore and also not supported nested field. 
I suppose that the elasticksearch which is used by Graylog can be a problem, but this is only my suspicion.

Example behavior:

| input          | output      | result |
|----------------|-------------|------------
| `column_id`: 12 | `column_id`: 12 | work fine 
| `_column_id`: 12 | `column_id`: 12 | work fine, but remove first sign undersore
| `__column_id`: 12 | `_column_id`: 12 | work, but replace name
| `column.id`: 12 | `_column_id`: 12 | replace name to: column_id 
| `column`: {`id`: 12} | `column`: {`id` : 12} | cann't search by nested field
| `column-id`: 12 | `column-id`: 12 | work fine
| `-column-id`: 12 | `-column-id`: 12 | cann't search by this field, is negate condition

When people running with the "minimalist" case, there is put empty string.


I also forgot to implement primary key (can be composite key) for log, so finally payload will be one send message as you suggested:

<pre>
</code>
{ 
	... timestamp, etc
	"primary_key_one": 1,
        "primary_key_two": 2,
        "column_user_id": 123, 
        "column_foo_column": "bar" 
        "column_old_user_id": 321
}
</code>
</pre>

`"primary_key_{column_name}"`, `"column_{column_name}"`, `"column_old_{column_name}"` must be prefix.

osheroff:
man, I gotta stop reviewing PRs and then going traveling.  I think it might be more effective here to just tunnel the data into a string that is then unwrapped, ie:

```
  "data": " { \"id\": 123 } "
```

otherwise there's too much weirdness around crashing field names and such.

@liulikun if I wander away here can you pick up my laziness?

liulikun:
Will keep an eye out.

PR #818, opened 2017-12-01
Allow default values with an explicit plus sign

timbertson:
This is the same as #797, I just fixed up the generated test files on top of hexene's work.


osheroff:
👍

Issue #819, opened 2017-12-07
Expose ColumnDef information in RowMap


sungjuly:
Hi team! Thank you for open sourcing a great project! Just wondering what do you think about exposing `ColumnDef` information in `RowMap`?  The current design assumes all data will be represented to JSON only eventually. 

It would be more flexible when `maxwell` runs as embedding. In my case, `ColumnDef` should be exposed in order to convert proper column types when transferring a POJO. I could contribute my small changes if you guys think ok. thank you again!

osheroff:
Makes sense to me, go ahead and PR it

sungjuly:
cool! I will make a PR soon!

PR #827, opened 2017-12-15
Expose ColumnDef information in RowMap

sungjuly:
#819

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowMap.java
@@ -5,6 +5,7 @@ import com.zendesk.maxwell.producer.EncryptionMode;
 import com.zendesk.maxwell.replication.BinlogPosition;
 import com.zendesk.maxwell.producer.MaxwellOutputConfig;
 import com.zendesk.maxwell.replication.Position;
+import com.zendesk.maxwell.schema.columndef.ColumnDef;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -39,7 +40,9 @@
 	private Long threadId;
 
 	private final LinkedHashMap<String, Object> data;
+	private final LinkedHashMap<String, ColumnDef> dataColumnDef;
 	private final LinkedHashMap<String, Object> oldData;
+	private final LinkedHashMap<String, ColumnDef> oldDataColumnDef;
```
osheroff: these can just be `HashMap`, no need to iterate them in order

osheroff:
otherwise lgtm 👍

sungjuly:
@osheroff thanks for reviewing quickly! I've changed to HashMap, would you please review this again?

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -39,7 +41,9 @@
 	private Long threadId;
 
 	private final LinkedHashMap<String, Object> data;
+	private final HashMap<String, ColumnDef> dataColumnDef;
 	private final LinkedHashMap<String, Object> oldData;
+	private final HashMap<String, ColumnDef> oldDataColumnDef;
```
timbertson: Is there any case where the column def in `old` and `data` would be different? If not, we could get rid of this.

sungjuly: Good a point. It seems there no cases where the column def in `old` and `data` would be different. I hope one thing it would be nice in terms of API consistency if we could define `getOldDataColumnDef` method in `RowMap` class. Do you want me to get rid of that method as well? Or just want me to reuse `dataColumnDef` variable in `getOldDataColumnDef`/`putOldData`?

timbertson: A `RowMap` can't represent data _across_ a schema change, so we can guarantee that old + new data must be the same column def. So I think it's fine to just update the same map from both `putData` and `putOldData`, and have a single `getColumnDef` which accesses this map.

If you want to be paranoid you could assert that whenever we put a column def in the map we're either adding a new entry to the map or replacing an identical column def, but that should already be the case.

mohamagdy:
@sungjuly, thanks for the effort you put here.

With this change, I assume the JSON event emitted will be changed because the column definition is added? If yes, can you please give an example?

Thanks,

PR #829, opened 2017-12-16
Stop metrics related threads when shutdown

sungjuly:
I've noticed maxwell doesn't stop metrics related threads when shutdown. This bug makes a trouble if maxwell is embedded to another application especially. All metrics keep reported as wrong numbers even if maxwell stopped successfully. This PR is for closing all threads, that is enabled in `MaxwellMetrics` class, when shutdown process.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/monitoring/MaxwellMetrics.java
+++ b/src/main/java/com/zendesk/maxwell/monitoring/MaxwellMetrics.java
@@ -27,6 +27,10 @@ public class MaxwellMetrics implements Metrics {
 	static final String reportingTypeHttp = "http";
 	static final String reportingTypeDataDog = "datadog";
 
+	private Slf4jReporter slf4jReporter;
+	private JmxReporter jmxReporter;
+	private DatadogReporter datadogReporter;
+
 	private static final Logger LOGGER = LoggerFactory.getLogger(MaxwellMetrics.class);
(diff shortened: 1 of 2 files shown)


timbertson:
Thanks, this looks good. For consistency, could you make MaxwellMetrics a StoppableTask and add it to MaxwellContext's `taskManager`? That keeps the shutdown logic somewhat consistent and gives us a bit more logging.

src/main/java/com/zendesk/maxwell/monitoring/MaxwellMetrics.java
```diff
@@ -89,17 +93,31 @@ private void setup(MaxwellConfig config) {
...
+
+		if(jmxReporter != null) {
+			jmxReporter.stop();
+		}
+
+		if(datadogReporter != null) {
+			datadogReporter.stop();
+		}
+	}
+
```
timbertson: Is it safe to call `stop` multiple times on these? Or should you set each of these to `null` after stopping?

sungjuly: good catch! I should have marked as null for those variables.

sungjuly:
@timbertson sure, I will take a look whether I could implement the cleanup logic with `StoppableTask` or not. And then get back to you guys.

PR #835, opened 2017-12-27
Row based filter and fix to the position commit issue in Google Pub/Sub

finnplay:
Hello,

This is Yang from Finnplay.
We are using Maxwell to stream the events to database and publish it to Google Pub/Sub for our customer. In order to make it working with out system, we introduced one feature and fixed one issue with help from @marcusnilssonjickl

Row-based filter, and an issue of committing position in MaxwellPubsubProducerWorker.

1. Row-based filter
  With this change, Maxwell can filter the rows based on the given column(s) and value(s).

  New configuration parameter has been added,  for example --include_column_values="COLUMN1=val1,COLUMN2=val2", so Maxwell will only stream the row which has columns (COLUMN1 and COLUMN2) and their values match given values (val1 and val2), or the row which does not have these columns at all.

  This is useful if the user wants to filter the data on row level, not only on database or table level, since our platform is an multi-tenancy system, several clients share the same database and table, we need to use this filter to stream the data belong to the right client, not all data from the same table.

2. Committing position issue in MaxwellPubsubProducerWorker.
  When streaming the events to Google Pub/Sub, we found that it gets stuck after some time. It seems Maxwell stops sending the event after sent around 1000 messages to Google Pub/Sub, then we restart the Maxwell, found it started to replay from the beginning. Finally we found that in MaxwellPubsubProducerWorker.sendAsync(), the CallbackCompleter is not invoked in order to commit the position, so we update the sendAsync() method to use PubsubCallback to commit the position.

It would be better to merge these two changes to master branch, please have a look.

Thanks!

BR,
Yang

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -60,6 +60,8 @@ public class MaxwellConfig extends AbstractConfig {
 	public String kinesisStream;
 	public boolean kinesisMd5Keys;
 
+	public String sqsQueueUri;
+	
 	public String pubsubProjectId;
 	public String pubsubTopic;
 	public String ddlPubsubTopic;
@@ -160,7 +162,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 		parser.accepts("__separator_3");
 
-		parser.accepts( "producer", "producer type: stdout|file|kafka|kinesis|pubsub" ).withRequiredArg();
+		parser.accepts( "producer", "producer type: stdout|file|kafka|kinesis|pubsub|sqs" ).withRequiredArg();
 		parser.accepts( "producer_ack_timeout", "producer message acknowledgement timeout" ).withRequiredArg();
 		parser.accepts( "output_file", "output file for 'file' producer" ).withRequiredArg();
 
@@ -178,7 +180,8 @@ public class MaxwellConfig extends AbstractConfig {
 		parser.accepts( "kafka_topic", "optionally provide a topic name to push to. default: maxwell").withRequiredArg();
 		parser.accepts( "kafka_key_format", "how to format the kafka key; array|hash").withRequiredArg();
 
-		parser.accepts( "kinesis_stream", "kinesis stream name").withRequiredArg();
+		parser.accepts( "kinesis_stream", "kinesis stream name").withOptionalArg();
+		parser.accepts("sqs_queue_uri", "SQS Queue uri").withRequiredArg();
 
 		parser.accepts( "pubsub_project_id", "provide a google cloud platform project id associated with the pubsub topic").withRequiredArg();
 		parser.accepts( "pubsub_topic", "optionally provide a pubsub topic to push to. default: maxwell").withRequiredArg();
@@ -396,6 +399,8 @@ public class MaxwellConfig extends AbstractConfig {
 
 		this.kinesisStream  = fetchOption("kinesis_stream", options, properties, null);
 		this.kinesisMd5Keys = fetchBooleanOption("kinesis_md5_keys", options, properties, false);
+		
+		this.sqsQueueUri = fetchOption("sqs_queue_uri", options, properties, null);
 
 		this.outputFile = fetchOption("output_file", options, properties, null);
 
@@ -548,6 +553,8 @@ public class MaxwellConfig extends AbstractConfig {
 			usageForOptions("please specify --output_file=FILE to use the file producer", "--producer", "--output_file");
 		} else if ( this.producerType.equals("kinesis") && this.kinesisStream == null) {
 			usageForOptions("please specify a stream name for kinesis", "kinesis_stream");
+		} else if (this.producerType.equals("sqs") && this.sqsQueueUri == null) {
+			usageForOptions("please specify a queue uri for sqs", "sqs_queue_uri");
 		}
 
 		if ( !this.bootstrapperType.equals("async")
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -37,7 +37,7 @@ public class MaxwellConfig extends AbstractConfig {
 
 	public String databaseName;
 
-	public String includeDatabases, excludeDatabases, includeTables, excludeTables, excludeColumns, blacklistDatabases, blacklistTables;
+	public String includeDatabases, excludeDatabases, includeTables, excludeTables, excludeColumns, blacklistDatabases, blacklistTables, includeColumnValues;
 
 	public ProducerFactory producerFactory; // producerFactory has precedence over producerType
 	public String producerType;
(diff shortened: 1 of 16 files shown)


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -225,6 +225,7 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "exclude_columns", "exclude these columns, formatted as exclude_columns=col1,col2" ).withRequiredArg();
 		parser.accepts( "blacklist_dbs", "ignore data AND schema changes to these databases, formatted as blacklist_dbs=db1,db2. See the docs for details before setting this!").withRequiredArg();
 		parser.accepts( "blacklist_tables", "ignore data AND schema changes to these tables, formatted as blacklist_tables=tb1,tb2. See the docs for details before setting this!").withRequiredArg();
+		parser.accepts( "include_column_values", "include only rows with these values formatted as include_column_values=C=x,D=y").withRequiredArg();
```
osheroff: `c=X,d=y` seems tricky to parse properly (think values that include commas).  Can the options parser accept multiple arguments with the same name?  thinking `--include_column_values=c=X --include_column_values=d=Y` would be superior here if possible/easy.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -130,6 +149,24 @@ public boolean matches(String database, String table) {
...
+
+				Object value = data.get(column);
+				if (value == null) return false;
+
+				String valueString = value.toString();
+				if (!columnValue.equals(valueString)) return false;
+			}
+		}
+
+		return true;
```
osheroff: I guess the decision to return true if the row doesn't contain the column is fine, but we should be explicit about it in the documentation.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -130,6 +149,24 @@ public boolean matches(String database, String table) {
...
 
+	private boolean matchesValues(Map<String, Object> data) {
+		for (Map.Entry<String, String> entry : includeColumnValues.entrySet()) {
+			String column = entry.getKey();
+
+			if (data.containsKey(column)) {
+				String columnValue = entry.getValue();
+
+				Object value = data.get(column);
+				if (value == null) return false;
```
osheroff: I could definitely see value in including NULL columns; we should special case that (if the user passes in C=null, it matches both NULL and "null".  Obviously kinda odd and non-ideal but hey it's ok for government work)

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -130,6 +149,24 @@ public boolean matches(String database, String table) {
...
+
+				String valueString = value.toString();
+				if (!columnValue.equals(valueString)) return false;
```
osheroff: In the future; regexp matching on the values would be nice.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -152,4 +189,12 @@ public static boolean matches(MaxwellFilter filter, String database, String tabl
...
 		}
 	}
+
+	public static boolean matchesValues(MaxwellFilter filter, String database, String table, Map<String, Object> data) {
+		if (filter == null) {
+			return true;
+		} else {
+			return filter.matchesValues(data);
+		}
+	}
```
osheroff: nope, I think you just accidentally disabled all the rest of maxwell filtering.  Please have `MaxwellFilter#matches` call `MaxwellFilter#matchesValues` instead.  (you'll need to change the signature of `matches`)

finnplay: Hi,
Thanks for the comment.
Actually the MaxwellFilter.matchesValues() is invoked from AbstractReplicator.shouldOutputRowMap(), which is finally called from BinlogConnectorReplicator.getTransactionRows().

In method BinlogConnectorReplicator.getTransactionRows(), the shouldOutputEvent() is called first in order to filter the database and table, then in the for loop of the rows, shouldOutputRowMap() is called in order to filter the row base on the values.

If we do the row filter in MaxwellFilter.matches(), we have to change the signature of the method to have row passed to this method, and it will be invoked once for each row, and it should have less performance than the current solution.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
finnplay: Hi, did you mean that we should have a separated folk on maxwell, then we add the the fix for Google Pub/Sub, then submit a new pull request for it?

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
finnplay: Seems it does support multiple arguments with same name, will try to make the change.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
finnplay: Yes, if none of the columns is found in the row, it returns true.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
finnplay: We will add support for null value.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
finnplay: It is supported if we use argument, but how should we handle it when we use properties file? Since properties file does not support multiple properties with same name. Any good idea?

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
osheroff: ah, yeah, I missed that.  ok.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: hmmm, good point.  I guess leave it comma separated.

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
osheroff: I just mean remove the pub-sub fix from this PR and open it as a new PR.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
finnplay: OK, I keep this unchanged.

finnplay:
Hi @osheroff 
As discussed, the support of value NULL has been added, and the fix for Google Pub/Sub has been removed from this PR, I will create another branch for that and submit a new pull request.
Could you please take a look?

Thank you!
Br,
Yang

src/main/java/com/zendesk/maxwell/MaxwellFilter.java
```diff
@@ -1,26 +1,46 @@
...
+	private final Map<String, String> includeColumnValues = new HashMap<>();
 
 	public MaxwellFilter() { }
+	public MaxwellFilter(String includeDatabases,
+						 String excludeDatabases,
+						 String includeTables,
+						 String excludeTables,
+						 String blacklistDatabases,
+						 String blacklistTables,
+						 String includeColumnValues) throws MaxwellInvalidFilterException {
```
osheroff: nitpicky:  please reformat, the code style here should be:
```
	public MaxwellFilter(
                String includeData,
                String excludeDatabases
                ...
        )
```

finnplay:
Hi @osheroff 

I have pushed the formatted MaxwellFilter.
Have a look.

Thank you!
Br,
Yang

finnplay:
Hi @osheroff 
Any update on this pull request?

BR,
Yang

osheroff:
Ok, I reformatted a bit more and merged by hand.

finnplay:
Thank you very much!

Br,
Yang

PR #836, opened 2017-12-29
Adding support for including and configuring a custom ProducerFactory.

tomcollinsproject:
In order for our organization to integrate Maxwell, we need to be able batch and broadcast transactions as a whole. There are a handful of complicated or half solutions out there (Apache Flink, RabbitMQ's consistent hash exchange, etc.), but we figured that the easiest thing would be to buffer the raw RowMaps processed by Maxwell until all the rows in a transaction are received. Once there's a mechanism for collecting all the rows for a transaction, we can use a custom solution for broadcasting the transactions to any consumers interested in one or more of the tables contributing to the transaction.

The ProducerFactory interface is exactly what we needed to integrate our custom transaction batching producer, but there was no way to configure it and any properties needed by the custom producer. This pull request exposes a couple configuration properties that allow us to do just that.


osheroff:
Sure, the idea looks fine.  Is this a WIP?  I don't see `producerFactory` being used anywhere...

tomcollinsproject:
I noticed that the MaxwellContext.getProducer() method first delegates to the MaxwellConfig.producerFactory if present. So what I'm doing right now is setting the producer_factory property I introduced to a custom class I've created in my own artifact and then dumping that artifact in the $MAXWELL_HOME/lib directory. It all seems to be working as I had hoped. Thoughts on that?

BTW, I'm going to dig in to what I broke in this build now.

osheroff:
I'm gonna read deeper before I comment.  Too much has changed in this project since I first wrote it :)  You're right on `producerFactory`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -368,7 +372,10 @@ private void setup(OptionSet options, Properties properties) {
 		if ( properties != null ) {
 			for (Enumeration<Object> e = properties.keys(); e.hasMoreElements(); ) {
 				String k = (String) e.nextElement();
-				if (k.startsWith("kafka.")) {
+				if (k.startsWith("custom.")) {
+				    this.customProperties.setProperty(k.replace("custom.", ""), properties.getProperty(k));
```
osheroff: maybe `custom_producer.` here?  I dunno.  Something a bit more clear.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -368,7 +372,10 @@ private void setup(OptionSet options, Properties properties) {
...
+				}
+				else if (k.startsWith("kafka.")) {
```
osheroff: stlyistic nit here and other places: `else` on the same line as braces.

osheroff:
Code looks good other than a couple of nits.  I would like to see this documented before merging...

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
tomcollinsproject: I was wondering about that, actually. How about `custom_producer_factory` just to be clear as to the expected implementation?

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
tomcollinsproject: On it.

docs/docs/producers.md
```diff
@@ -198,3 +198,66 @@ Other configurable properties are:
...
+    	// filtering out DDL and heartbeat rows
+        if(!r.shouldOutput(outputConfig))
+        {
+            context.setPosition(r.getPosition());
+            return;
+        }
+
+        // your custom producer logic here
+    }
+```
```
osheroff: i think let's just put these two files in an `example` directory somewhere and link to them from the docs

tomcollinsproject: I'm not quite sure what you're wanting here. I've started to create an "examples" page in the docs with these code examples, but I suspect I'm on the wrong track. Would you mind being a bit more specific about the format and where I should put these examples? I could also just omit the example code if that's more in keeping with project structure. Thanks.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
tomcollinsproject: I just saw this and I'll make this tweak.

docs/docs/producers.md
osheroff: i mean just put the example custom producers in like `src/example` and then drop a link to the github source page for those files in the documentation

tomcollinsproject: Got it. Thanks!

config.properties.example
```diff
@@ -176,6 +176,12 @@ sqs_queue_uri=aws_sqs_queue_uri
 #redis_database=0
 #redis_pub_channel=maxwell
 
+######### custom producer ######################
+# the fully qualified class name for custom ProducerFactory
+# see the following link for more details.
+# http://maxwells-daemon.io/producers/#custom-producer
+#custom_producer.factory=
```
osheroff: add an example for passing in settings to the custom producer, too:  `#customer_producer.custom_prop=foo`

osheroff:
good to go after example add 👍

osheroff:
sorry I missed this @tomcollinsproject, but this needs to be formatted with tabs instead of spaces.  sorry for the last go-round, should have caught that earlier.

tomcollinsproject:
Ah, nuts. I had some of my work format settings still in my IDE. Sorry I spaced that. Okay, terrible pun. Sorry for that, too. :(

osheroff:
thanks for all the loops, nice work.

tomcollinsproject:
Thanks for all your help! We're really excited to integrate this in to our system.

Issue #804, opened 2017-11-10
Message Ordering inconformity


AlvinYang-2017:
My topic have 5 partitions, I have been set  producer_partition_by=database, but the same database ddl and dml statement  were assigned to different partition, so caused consume message ordering inconformity. 

ddl message key : 3f083bf2-b5f7-4089-a6ed-0a47a00ddcda
dml message key :  {"database":"p2","table":"test_a","pk.id":1}

The reason may be message key format different cause  producer_partition_by function ineffective.
Who can tell me the real reason~  
thanks a lot.

@liulikun

liulikun:
Did you config a different topic for ddl `ddl_kafka_topic`?

osheroff:
this seems like a bit of a bug; DDL messages totally ignore producer_partition_by IIRC.  Obviously DDL messages would have to ignore producer_partition_by=column and sometimes =table (if the DDL applied to a database and not a table), but we should probably make a best-effort here.

osheroff:
@AlvinYang-2017 getting back to this, are you using kafka or kinesis?

osheroff:
oh, crud.  I think I see the issue.

osheroff:
ddl statements now properly honor partitioning by database

PR #837, opened 2017-12-30
Fix-up DDL partitioning

osheroff:
DDL now will support at least being partitioned by database or table.
Addresses issue #804.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/partitioners/HashStringTable.java
+++ b/src/main/java/com/zendesk/maxwell/producer/partitioners/HashStringTable.java
@@ -9,6 +9,12 @@ import java.util.List;
  */
 public class HashStringTable implements HashStringProvider {
 	public String getHashString(RowMap r, List<String> partitionColumns, String partitionKeyFallback) {
-		return r.getTable();
+		String t = r.getTable();
+
+		// support table-to-database fallback for DDL that has no table associated
+		if ( t == null && partitionKeyFallback.equals("database") )
+			return r.getDatabase();
+		else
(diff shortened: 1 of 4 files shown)


osheroff:
@timbertson @liulikun and whoever else is reviewing PRs @ zendesk these days

src/main/java/com/zendesk/maxwell/producer/partitioners/HashStringTable.java
```diff
@@ -9,6 +9,12 @@
  */
 public class HashStringTable implements HashStringProvider {
 	public String getHashString(RowMap r, List<String> partitionColumns, String partitionKeyFallback) {
-		return r.getTable();
+		String t = r.getTable();
+
+		// support table-to-database fallback for DDL that has no table associated
+		if ( t == null && partitionKeyFallback.equals("database") )
```
liulikun: maybe `"database".equals(partitionKeyFallback)` in case `partitionKeyFallback` is `null`

src/main/java/com/zendesk/maxwell/producer/partitioners/HashStringTable.java
```diff
@@ -9,6 +9,12 @@
...
+			return r.getDatabase();
+		else
+			return r.getTable();
```
liulikun: can just return `t`

liulikun:
I just got back from holidays. Thanks for the help during this holiday season. This PR looks fine apart from a couple of nit picks.

osheroff:
ok updated

PR #841, opened 2018-01-09
Fix to the position commit issue in Google Pub/Sub

finnplay:
Hello,

When streaming the events to Google Pub/Sub, we found that it gets stuck after some time. It seems Maxwell stops sending the event after sent around 1000 messages to Google Pub/Sub, then we restart the Maxwell, found it started to replay from the beginning. Finally we located the issue that in MaxwellPubsubProducerWorker.sendAsync(), the CallbackCompleter is not invoked in order to commit the position, so we update the sendAsync() method to use PubsubCallback to commit the position.

BR,
Yang


osheroff:
👍 wow this was broken before eh.

PR #845, opened 2018-01-10
Ben/rewrite partition choice

osheroff:
this removes the overworked "HashStringProvider" bits and centralizes the logic of choosing a partition key into a `switch` statement.  Also centralizes the fallback logic.

Note that this includes PR #837, probably best to review after merging that.


src/main/java/com/zendesk/maxwell/producer/partitioners/AbstractMaxwellPartitioner.java
```diff
@@ -5,30 +5,36 @@
...
 			case "database":
+				return PartitionBy.DATABASE;
+			case "primary_key":
+				return PartitionBy.PRIMARY_KEY;
+			case "column":
+				return PartitionBy.COLUMN;
 			default:
-				this.provider = new HashStringDatabase();
-				break;
+				return null;
```
liulikun: The default was by database before?

src/main/java/com/zendesk/maxwell/producer/partitioners/AbstractMaxwellPartitioner.java
```diff
@@ -43,7 +49,32 @@ static protected String getPrimaryKey(RowMap r) {
 		return r.pkAsConcatString();
 	}
 
+	public String getHashString(RowMap r, PartitionBy by) {
+		if ( by == null )
+			return UUID.randomUUID().toString();
```
liulikun: Is this new behaviour? Should we use `database` as default?

src/main/java/com/zendesk/maxwell/producer/partitioners/AbstractMaxwellPartitioner.java
osheroff: the default for invalid strings was database, yeah.  I think if you pass an unknown string this should throw a runtime error, but if you pass `null` you get `PartitionBy.DATABASE`.  Since this is also used to determine the fallback partition-by I think that'll cover both comments.

PR #849, opened 2018-01-17
bump kpl to 0.12.8 (ATS cert compat)

chrislewis:
AWS is requiring kinesis clients to upgrade to KPL version >= `0.12.6` to support ATS tickets. This pull updates the KPL to the latest version of `0.12.8`, which is in the same major/minor family as the current dependency of `0.12.3`. We've been running this since last night in [our fork](https://github.com/meetup/maxwell) with no problems so far.


osheroff:
great, thanks 👍

PR #852, opened 2018-01-18
added SSL config options for shyiko binlog replicator

### LINKED ISSUE #738: BinlogConnectorReplicator does not support SSL
`BinlogConnectorReplicator` does not support SSL, but the underlying `BinaryLogClient` from `mysql-binlog-connector-java` implemented SSL support in [version 0.4.0](https://github.com/shyiko/mysql-binlog-connector-java#using-ssl).

cadams5:
We are using Maxwell to connect to a production MySQL DB and need the communication to be over TLS. This is to add a config entry for Maxwell to be able to customize the [SSLMode](https://github.com/shyiko/mysql-binlog-connector-java/blob/eead0ec2338f7229ba74a7a04188f4967f13d4b7/src/main/java/com/github/shyiko/mysql/binlog/network/SSLMode.java).

Shyiko MySQL Binary Log connector documentation is [here](https://github.com/shyiko/mysql-binlog-connector-java#using-ssl).

Resolves #738


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -303,6 +306,17 @@ public RowMap getRow() throws Exception {
...
 
+	private SSLMode getBinlogSSLMode(String sslMode) {
+		if (sslMode != null) {
+			for (SSLMode mode : SSLMode.values()) {
+				if (mode.toString().equals(sslMode)) {
+					return mode;
+				}
+			}
+		}
+		return SSLMode.DISABLED;
```
osheroff: should add a warning string here that the user passed an invalid mode string.

osheroff:
thanks for the PR!  any way you could unify this code with the other ssl-enabling code?  I'd like to see a single flag be able to turn on ssl for all connections, or each of the individual connections, so there'd be 3 flags:

```
--ssl=MODE (turns on ssl for the maxwell-store connection, other connections inherit this setting unless specifed)
--replication_ssl=mode (for binlog-connector)
--schema_ssl=MODE (for the schema-capture connection, if used)
```

Also, what's the behavior of this (and other ssl-behaviors) if `javax.net.ssl.trustStore=<truststore>` is not set?  I'd prefer a graceful exit-with-instructions-on-how-to-fix if possible.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
cadams5: done!

cadams5:
to unify them, what do you think about changing the SSLMode of the binlog connector depending on the JDBC option `requireSSL=true` and `verifyServerCertificate=true`. something like:

```
if (mysqlConfig.jdbcOptions.contains("verifyServerCertificate=true")) {
	this.client.setSSLMode(SSLMode.VERIFY_CA);
}
else if (mysqlConfig.jdbcOptions.contains("requireSSL=true")) {
	this.client.setSSLMode(SSLMode.REQUIRED);
}
```

that way if the connection to store schema/binlog is over SSL, the binlog connector will also use it.

cadams5:
or the opposite, setting those JDBC options based upon the flags set in the config

osheroff:
i think the opposite is more slick.  feels like "real" support

cadams5:
agreed. I'll get that done tomorrow!

cadams5:
I updated with those changes - a few notes:

- I added 3 flags for SSL, with two latter being individual overrides for the first
- modified `config.properties` docs to only specify trustStore (keystore should not be needed)
- changed `MaxwellMysqlConfig` to use a Map instead of List, so duplicate keys cannot be added to the JDBC options
- changed `MaxwellMysqlConfig` to use `URIBuilder` for connection URI to increase robustness
- added tests

if a truststore is not specified, and SSL certs on the MySQL side are not signed by a CA in the provided Java truststore, then it will fail with an SSL exception (I can run some tests locally to verify the exact exception). so maybe the documentation should be updated to be more specific about when -Djavax.net.ssl.trustStore is needed? whatever you guys think.

cadams5:
another thing to note, that would only be applicable if using `VERIFY_CA` or `VERIFY_IDENTITY`

osheroff:
think you missed the boat a little bit: `MaxwellMysqlConfig` is designed to represent a *single* connection's options, so there should just be one `sslMode` value in there, as there's 3 instances of `MaxwellMysqlConfig` (one for main, replica, schema-capture)

cadams5:
ahh I see that now, that makes more sense. okay I will update to set on those 3 instances.

cadams5:
@osheroff thanks for the feedback, this is now ready for review

src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
```diff
@@ -24,46 +32,112 @@ public MaxwellMysqlConfig() {
...
+		if (sslMode != null && sslMode != SSLMode.DISABLED) {
+			this.useSSL(true); // for all SSL modes other than DISABLED, use SSL
+
+			this.verifyServerCertificate(false); // default to not verify server cert
+			this.requireSSL(false); // default to not require SSL
+
+			if (sslMode == SSLMode.REQUIRED || sslMode == SSLMode.VERIFY_CA
+					|| sslMode == SSLMode.VERIFY_IDENTITY) {
+				this.requireSSL(true);
+			}
```
osheroff: could boil this down to:
```
   this.requireSSL(sslMode == SSLMode.REQUIRED || sslMode == SSLMode.VERIFY_CA || sslMode == SSLMode.VERIFY_IDENTITY)
```

src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
```diff
@@ -24,46 +32,112 @@ public MaxwellMysqlConfig() {
...
+
+			if (sslMode == SSLMode.VERIFY_IDENTITY) {
+				this.verifyServerCertificate(true);
```
osheroff: same here, `this.verifyServerCertificate(sslMode == SSLMode.VERIFY_IDENTITY)`

minor, though, as written is fine too

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -61,6 +62,10 @@ public BinlogConnectorReplicator(
 		transactionRowCount = metrics.getRegistry().histogram(metrics.metricName("transaction", "row_count"));
 
 		this.client = new BinaryLogClient(mysqlConfig.host, mysqlConfig.port, mysqlConfig.user, mysqlConfig.password);
+
+		SSLMode binlogSslMode = mysqlConfig.sslMode;
+		this.client.setSSLMode(binlogSslMode);
```
osheroff: no need for the local variable here, yeah?

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -366,6 +371,10 @@ private void setup(OptionSet options, Properties properties) {
 		this.redisDatabase		= Integer.parseInt(fetchOption("redis_database", options, properties, "0"));
 		this.redisPubChannel	= fetchOption("redis_pub_channel", options, properties, "maxwell");
 
+		if (this.maxwellMysql.sslMode == null) {
+			this.maxwellMysql.sslMode = SSLMode.DISABLED;
+		}
+
```
osheroff: maybe just set this as the default in the `MawellMysqlConfig` initializer?

cadams5: the reason I was defaulting to null was so we could differentiate between not set and disabled, since replication and schema SSLMode's need to be overridden only if they're not set

osheroff: I see.  Ok, then that's ok for now.  I may clean this up a bit later by maybe unifying some of this stuff under something like `MaxwellMysqlConfig#clone(replicationHost, replicationPassword, ...)` which can be smart about inheriting stuff.

cadams5: that sounds great. yeah I don't like the null checks, there's definitely a better way of doing it

cadams5:
any other changes you'd like to see for this PR?

config.properties.example
```diff
@@ -39,6 +39,17 @@ password=maxwell
 # name of the mysql database where maxwell keeps its own state
 #schema_database=maxwell
 
+# SSL/TLS options
+# To use any modes other than DISABLED, you must set the trust store:
+#   set JAVA_OPTS="-Djavax.net.ssl.trustStore=<truststore> -Djavax.net.ssl.trustStorePassword=<password>"
+# MODE must be one of DISABLED, PREFERRED, REQUIRED, VERIFY_CA, or VERIFY_IDENTITY
```
osheroff: so I am testing the code locally, it seems like it works fine with `--ssl=REQUIRED` without bothing to set any of the `trustStore` options -- maybe those are only required for `VERIFY_CA` and `VERIFY_IDENTITY`?

cadams5: I'll locally test as well - that makes sense, though. [here's](https://github.com/shyiko/mysql-binlog-connector-java/blob/master/src/main/java/com/github/shyiko/mysql/binlog/BinaryLogClient.java#L669) where the SSLMode is used in the shyiko binlog connector, and [here](https://github.com/shyiko/mysql-binlog-connector-java/blob/master/src/main/java/com/github/shyiko/mysql/binlog/BinaryLogClient.java#L90) are the certificate checking configs.

so for `REQUIRED` and `PREFERRED`, there aren't any checks that happen (those `TrustManager` methods are stubbed out), but for the higher ones, the default check is used.

cadams5: if that's the case, I can update the comments here to include that

cadams5:
just did some testing locally to confirm when `javax.net.ssl.trustStore` is needed:

when anything above `REQUIRED` is specified (`VERIFY_CA` or `VERIFY_IDENTITY`) and trustStore is not set, the full error is:

```
javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

the way to fix this would be to set the Java opts when running the application, or importing the MySQL keystore into the global Java store.

another note, if using `VERIFY_IDENTITY` and the cert is not configured with the correct host, it will fail with something like:

```
com.github.shyiko.mysql.binlog.network.IdentityVerificationException: "localhost" identity was not confirmed
```

hopefully this is helpful for anyone who runs into issues in the future.

I will update comments in the `config.properties.example` file accordingly

cadams5:
made those changes to `config.properties.example`

osheroff:
nice work, merging 👍

PR #853, opened 2018-01-19
Rework #698, fix for two tables named the same

osheroff:
This picks up the work started in #698 and simplifies the fix to a one-liner.

@liulikun


osheroff:
ping here, would like to get some bugfixes in for next release, would love to follow a 👍 CR policy as well... @timbertson you still halfway-maintaing maxwell?

PR #870, opened 2018-02-06
Custom rowmap attributes

jkgeyti:
With the new ability to create custom producer, it would be nice to be able to add extra attributes to RowMaps. This PR adds a Map holding extra attributes, which are simply added to the root json object is `toJSON`.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/errors/ProtectedAttributeNameException.java
+++ b/src/main/java/com/zendesk/maxwell/errors/ProtectedAttributeNameException.java
@@ -1,0 +1,7 @@
+package com.zendesk.maxwell.errors;
+
+public class ProtectedAttributeNameException extends RuntimeException {
+	public ProtectedAttributeNameException(String message) {
+		super(message);
+	}
+}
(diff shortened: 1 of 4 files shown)


liulikun:
The code looks good. Would be nice to have a unit test on this new feature.

jkgeyti:
Added a test `testGetExtraAttributesMaps` that's pretty much a clone of `testGetDataMaps`, only using extra attributes.

One thing I'm a little unsure about, is what to do in case user adds an extra attribute that already exists (like "database"). I guess it's kind of nice to have the ability, but it feels a little dirty.

Would you prefer having it kick up a fuss if adding a protected key?

jkgeyti:
This last commit is a somewhat opinionated attempt at adding protected field names. It's a bit more invasive, so someone might not agree with the approach. 

RowMap throws a new ProtectedAttributeNameException RuntimeException if an extra attribute with a protected name (database, table, etc) is added.

liulikun:
Good thinking on the protection of internal fields.

liulikun:
👍

Issue #866, opened 2018-02-05
Managing kafka's binlog position in custom producer


jkgeyti:
First of, apologies for posting on the issue tracked, but I couldn't seem to find a suitable community to post this in.

I assume `context.SetPosition` stores the offset in maxwell's mysql database, so the producer can continue from that position on restart. Please correct me if I'm wrong.

The [CustomProducer example](https://github.com/zendesk/maxwell/blob/master/src/example/com/zendesk/maxwell/example/producerfactory/CustomProducer.java) calls `context.setPosition(r.getPosition());` only on DDL and heartbeat events. Why is that? Should it not be called only after writing (and hence persisting) messages to stdout? Is `context.setPosition` ever called indirectly by other parts of maxwell?

Being able to choose when to store a binlog position in kafka's mysql database is a crucial part of writing a safe custom producer, and it would probably be worth highlighting how and when to do it safely in the custom producer example. I'll be happy to do a PR, but I wanted to make sure I understand the expected behavior and responsibilities of `AbstractProducer.push` and `MaxwellContext.setPosition` before doing so.

osheroff:
Nice catch.   Yeah, please do a PR.  A producer may go off and write rows in the background of `push`, only calling `setPosition` later (the kafka/kinesis producers do this), but yes, generally a producer should be calling `setPosition` every time it's finished processing a row.

jkgeyti:
https://github.com/jkgeyti/maxwell/pull/1

PR #873, opened 2018-02-08
Change setPosition behaviour in custom producer example

jkgeyti:
https://github.com/zendesk/maxwell/issues/866

### DIFF
--- a/src/example/com/zendesk/maxwell/example/producerfactory/CustomProducer.java
+++ b/src/example/com/zendesk/maxwell/example/producerfactory/CustomProducer.java
@@ -9,6 +9,16 @@ import java.util.Collection;
 
 /**
  * Custom {@link AbstractProducer} example that collects all the rows for a transaction and writes them to standard out.
+ * To ensure atomicity, the producer saves its position in the binlog only after writing a full transaction to stdout.
+ *
+ * When writing a producer, bear in mind the following ordering guarantees provided by maxwell and the binlog (in row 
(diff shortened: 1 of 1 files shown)


src/example/com/zendesk/maxwell/example/producerfactory/CustomProducer.java
```diff
@@ -25,19 +35,28 @@ public void push(RowMap r) throws Exception
 	{
 		// filtering out DDL and heartbeat rows
 		if(!r.shouldOutput(outputConfig)) {
+			// though not strictly necessary (as skipping has no side effects), we store our position,
+			// so maxwell won't have to "re-skip" this position if crashing and restarting.
```
osheroff: this comment is just a touch mis-leading, we'd prefer to not re-process DDL statements, although I *think* they're idempotent to maxwell

jkgeyti: In this example producer we ignore all DDL events (`f(!r.shouldOutput(outputConfig))`), so it can re-processes them as many times it wants in case of restarts. I'm not to familiar with the rest of the codebase, so I don't know if choosing to commit our position or not has any side effects.

Rereading the message though, I think it's just unnecessarily confusing. If the producer can safely ignore the message, the example probably shouldn't commit the position, and we can remove that comment to avoid any confusion. And if committing/not committing our position has side effects, then that should be what's highlighted in the comment instead.

PR #874, opened 2018-02-09
mysql config doc fix

ashashwat:
In [mysqld],  it is server_id and not server-id.

It can be tested from mysql,

```
mysql> SHOW VARIABLES LIKE 'server%id';
+---------------+--------------------------------------+
| Variable_name | Value                                |
+---------------+--------------------------------------+
| server_id     | 1                                    |
| server_uuid   | 9f6ef86a-e632-11e7-b679-45de8b1dee56 |
+---------------+--------------------------------------+
2 rows in set (0.00 sec)
```

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -5,7 +5,7 @@ operate:
 
 ```
 [mysqld]
-server-id=1
+server_id=1
 log-bin=master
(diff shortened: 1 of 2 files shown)


timbertson:
Indeed it is, thanks 👍

PR #877, opened 2018-02-12
Fix jdbc option parsing for cases when value contains =

justinmichaud:
Currently, `jdbc_options=sessionVariables=wait_timeout=12` does not get set properly. This change should fix it.


osheroff:
thx for the PR.

PR #878, opened 2018-02-13
Kafka 1.0.0 support and add KAFKA_BROKERS to support kafka cluster.

MadeInChina:
Usage
```
maxwell:
    image: zendesk/maxwell
    container_name: dev-maxwell
    environment:
      MYSQL_PASSWORD: "@password"
      MYSQL_USERNAME: "maxwell"
      MYSQL_HOST: mysql
      KAFKA_BROKERS: broker1:9092,broker2:9092,broker2:9092
```


Dockerfile
```diff
@@ -16,4 +16,4 @@ RUN apt-get update \
 
 WORKDIR /app
 
-CMD [ "/bin/bash", "-c", "bin/maxwell --user=$MYSQL_USERNAME --password=$MYSQL_PASSWORD --host=$MYSQL_HOST --producer=kafka --kafka.bootstrap.servers=$KAFKA_HOST:$KAFKA_PORT $MAXWELL_OPTIONS" ]
+CMD [ "/bin/bash", "-c", "bin/maxwell --user=$MYSQL_USERNAME --password=$MYSQL_PASSWORD --host=$MYSQL_HOST --producer=kafka --kafka.bootstrap.servers=$KAFKA_BROKERS $MAXWELL_OPTIONS" ]
```
osheroff: is there a way to add support for `KAFKA_BROKERS` without breaking existing boot scripts that rely on `KAFKA_HOST` and `KAFKA_PORT`?

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@
 ### [v1.11.0](https://github.com/zendesk/maxwell/releases/tag/v1.11.0): "the latest, the greatest"
 
 
-     - default kafka client upgrades to 0.11.0.1
+     - default kafka client upgrades to 1.0.0
```
osheroff: minor, undo this change.

Dockerfile
MadeInChina: There is no change to do that.

osheroff: if you can't do it in docker, please add it to `bin/maxwell`

MadeInChina: https://kafka.apache.org/10/documentation.html
Here is the bootstrap.servers definition
A list of host/port pairs to use for establishing the initial connection to the Kafka cluster. The client will make use of all servers irrespective of which servers are specified here for bootstrapping—this list only impacts the initial hosts used to discover the full set of servers. This list should be in the form host1:port1,host2:port2,....

MadeInChina: i see.

xuebingli:
Done in #1008.

PR #883, opened 2018-02-17
make nexus deploy part of release process

osheroff:
@timbertson @ericwush this gets "mvn deploy" (which uploads to sonatype) as part of the deploy process.  want to make sure you all have the right credentials before I merge this tho.

### DIFF
--- a/build/release
+++ b/build/release
@@ -10,8 +10,10 @@ BINDIR=File.dirname(__FILE__)
 
 CURRENT_REV=`#{BINDIR}/current_rev`
 
+SUBCOMMANDS=%w(changelog bump build tag release docs nexus)
+
 if !( ($version = ARGV[0]) && $version =~ /^\d+\.\d+\.\d+(-\w+)?$/)
-  message =  "usage: #{$0} VERSION [bump|build|tag|release|changelog|docs]|\n"
+  message =  "usage: #{$0} VERSION [#{SUBCOMMANDS.join('|')}]\n"
   message += "  (current version is #{CURRENT_REV})"
   abort(message)
 end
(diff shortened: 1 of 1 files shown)


timbertson:
Yep, this is fine. We might ping you once we've got our users set up though because I don't know if we have permission yet.

ericwush:
@osheroff I'm very close to release the artifacts however I'm getting
```
[ERROR] Possible causes of 403 Forbidden:
[ERROR]  * you have no permissions to stage against profile with ID "a4a927da5c94bc"? Get to Nexus admin...
```
What process do I need to follow to get the permission granted? I couldn't find the jira that opened maxwell project.

osheroff:
@ericwush here's what I heard back from sonatype:

> In cases like this, we recommend that you create accounts/usernames in this JIRA for the folks you wish to be able to deploy to your groupId. When that's done, comment here with the usernames, and we can grant them the appropriate permissions and, if necessary, revoke permissions from your account.

if I recall correctly, the procedure is:

1. setup a jira account over at https://issues.sonatype.org - we probably should setup a shared account here.  
2. give them a gpg key
3. All releases then authenticate / share off that key.

My memory is a *bit* foggy tho.

ericwush:
Thanks @osheroff . If I understand it correctly, I should 
- create a shared account with sonatype, share with you the username and encrypted password (or you can just use your existing account?)
- distribute gpg public key to a key server (and share with you the private key? or you can just use your existing gpg key?)

You'll update the JIRA to allow the newly created account to deploy. Does it sound right?

ericwush:
I uploaded the public key. JIRA account details:
```
Username: zendesk_maxwell
email: goanna@zendesk.com
Full Name: Zendesk Maxwell
```

osheroff:
ok I sent a jira request

Issue #875, opened 2018-02-09
Running maxwell as daemon


ashashwat:
Can we run Maxwell as a daemon?  

Say something like Kafka where we can use `"$KAFKA_HOME/bin/kafka-server-start.sh -daemon $KAFKA_HOME/config/server.properties"` to run it like a daemon.

osheroff:
not currently.  PRs welcome.

PR #885, opened 2018-02-18
Running maxwell as daemon.

ashashwat:
With `--daemon `as an optional argument, maxwell now runs as a daemon.

This is with reference to issue [875](https://github.com/zendesk/maxwell/issues/875).

### DIFF
--- a/bin/maxwell
+++ b/bin/maxwell
@@ -45,8 +45,26 @@ do
 		use_kafka "${key#*=}"
 		;;
 
+        --daemon)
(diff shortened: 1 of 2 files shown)


ashashwat:
`mvn -e -Dtest=InflightMessageListTest#testAddMessageWillWaitWhenCapacityIsFull test` works fine locally.  Seems like an intermittent test failure to me.

ashashwat:
@osheroff Thanks for merging this.  Let me add some documentation.

Issue #832, opened 2017-12-22
Add support for parsing binlog from file


leonjoyce:
Currently maxwell seems only parse binlog from mysql, who will only dump "online" binlog files(show master logs). In our experience with Aliyun RDS, online binlog files is periodically purged, so that we have to frequently manually modify maxwell.position table to redirect the position to latest binlog. The consequence is that we skipped some old binlogs and DML/DDL is lost. So there is a need to pickup those lost binlogs and supplement their data to maxwell producer.

I have a PR for adding support for parsing binlog from local/online file. Not sure whether it can be approved.

seanlook:
I'm interested in this feature. Wondering how you parse offlined binlog as the data source. I archive this by moving those binlogs to a new mysql server. But the table schema is the prolem.

osheroff:
I'm interested to see the PR, but as @seanlook points out, the database schema is a problem; you'd have to be able to supply maxwell with the proper schema for the files you need to replay.  

What happens to maxwell when Aliyun purges the binlog files?  Maxwell crashes?  How?

seanlook:
Hi @osheroff ,
That's why I opened issue #833 . I could create database schemas filtered by `include_tables`, not all of the schemas. But binlog contains those tablemap that does not in **fake** mysql binlog server.

leonjoyce:
@seanlook Schema should not be a problem. Let's consider the senario: maxwell positions to binlog file1 and in replica DB we have online master logs of 4,5,6,7. So binlog 2,3 is purged.
In my new code maxwell will start from 1, restoring Schema to that point. then download 2,3 from Aliyun OSS. Parse them. and then we need manually point maxwell position at binlog.(4).

leonjoyce:
@osheroff   when Aliyun purges the binlog files, like mentioned in the senario above, Maxwell will repeated restart, with the following exception:
Nov 25 11:13:53 prod-consul-1 consul[27239]: 11:13:53,919 WARN BinlogConnectorReplicator - replicator stopped at position: 1a65b0bd-49b5-11e7-b8d2-008cfae27ce0:1-53881705,35a3a92d-49b5-11e7-b8d3-7cd30ac4ed32:1-245379,16a6d3a9-61f8-11e7-9708-7cd30ac4272a:1-30407903 -- restarting
[...]

PR #886, opened 2018-02-18
file reader

leonjoyce:
Add support for parsing binlog from file：
https://github.com/zendesk/maxwell/issues/832


osheroff:
this is far too specific to your use-case; the code to download the binlogs from alyiun certainly doesn't belong in maxwell.  The file-reading code is a maaaaaybe addition to maxwell, but it certainly needs to be better than a giant copy/paste of globs of code.

Issue #782, opened 2017-10-13
Ensure new client_id maxwells don't borrow incorrect schemas


Hackeruncle:
```
14:33:20,420 INFO  Server - Started @2886ms
14:33:20,635 INFO  TaskManager - Stopping 4 tasks
14:33:20,635 ERROR TaskManager - cause: 
java.lang.ClassCastException: java.math.BigDecimal cannot be cast to [B
        at com.zendesk.maxwell.schema.columndef.StringColumnDef.asJSON(StringColumnDef.java:78) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.replication.BinlogConnectorEvent.writeData(BinlogConnectorEvent.java:104) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.replication.BinlogConnectorEvent.buildRowMap(BinlogConnectorEvent.java:152) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.replication.BinlogConnectorEvent.jsonMaps(BinlogConnectorEvent.java:179) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getTransactionRows(BinlogConnectorReplicator.java:179) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getRow(BinlogConnectorReplicator.java:276) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.replication.AbstractReplicator.work(AbstractReplicator.java:148) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:27) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.Maxwell.startInner(Maxwell.java:190) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.Maxwell.start(Maxwell.java:140) ~[maxwell-1.10.8.jar:?]
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:211) ~[maxwell-1.10.8.jar:?]
14:33:20,639 INFO  TaskManager - Stopping: com.zendesk.maxwell.schema.PositionStoreThread@4b6b8148
[...]

PR #887, opened 2018-02-18
use positions from other clients when starting up a new client

osheroff:
since we're not re-capturing schemas, it's wrong to assume that we
should pick up from the master's current position: if there is another
(old, stopped) client, we'll end up corresponding the current position
to an older schema, which is no good.

Addresses #782

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
@@ -171,9 +171,9 @@ public class MysqlPositionStore {
 		return new Position(pos, rs.getLong("last_heartbeat_read"));
 	}
 
-	public Position getAny() throws SQLException {
+	public Position getLatestFromAnyClient() throws SQLException {
 		try ( Connection c = connectionPool.getConnection() ) {
-			PreparedStatement s = c.prepareStatement("SELECT * from `positions` where server_id = ?");
+			PreparedStatement s = c.prepareStatement("SELECT * from `positions` where server_id = ? ORDER BY last_heartbeat_read desc limit 1");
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlPositionStore.java
```diff
@@ -156,24 +156,37 @@ public Long getLastHeartbeatSent() {
...
+			gtid,
+			null,
+			rs.getLong("binlog_position"),
+			rs.getString("binlog_file")
+		);
+
+		return new Position(pos, rs.getLong("last_heartbeat_read"));
+	}
+
+	public Position getAny() throws SQLException {
```
timbertson: Would it be worth instead having `getLatest()` (sorted by `last_heartbeat_read`)? That way you don't pick up an offset from an old abandoned client.

Also, should this filter `gtid_set=<am-i-using-gtid>` to get only a compatible position?

osheroff: yup, good point on sorting, will change. 

re: GTID... hmmm.  If we have an old client running non-gtid, and we boot a new client with gtid, we still want to borrow the binlog position of the old client, or else we'll run into the same issue with schema sharing.  But I don't think it'd work properly to run in GTID mode with that old, non-gtid position.  

For that matter, Maxwell has *never* had a decent story around a graceful non-GTID -> GTID upgrade path; I think our answer has always been "drop database maxwell;".  I mean try reading the mysql docs on how to get an initial GTID lock, it's really quite bad.

timbertson: Fair point, I don't know how a transition would actually work.

Issue #889, opened 2018-02-19
Maxwell version issue


ashashwat:
Upon starting Maxwell we see:

``16:00:06,008 INFO  Maxwell - Maxwell v?? is booting (MaxwellKafkaProducer), starting at Position[BinlogPosition[1.000006:9011507], lastHeartbeat=1519034557599]``.

The Maxwell version here is replaced with "`??`".  

Upon inspection, version is picked from `getClass().getPackage().getImplementationVersion()`.  This is always _null_ here.  

When I built the package, the content of META-INF/MANIFEST.MF in maxwell-1.13.0.jar were,  

```
Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Built-By: shashwatanand
Created-By: Apache Maven 3.5.2
Build-Jdk: 1.8.0_152
```
After adding,  

```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-jar-plugin</artifactId>
  <configuration>
    <archive>
      <manifest>
        <mainClass>com.zendesk.maxwell.Maxwell</mainClass>
        <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
        <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
      </manifest>
    </archive>
  </configuration>
</plugin>

```

in pom.xml finally gives the content of META-INF/MANIFEST.MF in maxwell-1.13.0.jar as expected, 

```
Manifest-Version: 1.0
Implementation-Title: maxwell
Implementation-Version: 1.13.0
Archiver-Version: Plexus Archiver
Built-By: shashwatanand
Implementation-Vendor-Id: com.zendesk
Created-By: Apache Maven 3.5.2
Build-Jdk: 1.8.0_152
```   

However, `getClass().getPackage().getImplementationVersion()` is still null.  

The way I see it, parsing the META-INF/MANIFEST.MF at runtime is an option.  Alternatively, this information can be pulled from a property file.

ashashwat:
#891 fixes this.  Closing it.

PR #890, opened 2018-02-19
Version fix

ashashwat:
This PR fixes the issue [#889](https://github.com/zendesk/maxwell/issues/889) .  

To summarize, currently, Maxwell version is always missing.  A property file `src/main/resources/maxwell.property` can be used to pull the version information from pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -333,6 +333,19 @@
         </configuration>
       </plugin>
       <plugin>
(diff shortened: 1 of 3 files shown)


osheroff:
hey, thanks for the PR.  looks like this broke when we switched back to maven.  here's a much simpler fix: https://github.com/zendesk/maxwell/pull/891

PR #892, opened 2018-02-19
Adding missing ``break'' in SQS producer.

ashashwat:
Without this SQS producer will fallback to pubsub producer.  
I suppose this crept in during  [58b910](https://github.com/ashashwat/maxwell/commit/58b910677cae2303a31d9a68343fe5d4f3238c31).


osheroff:
yuck.  nice catch. 👍

PR #895, opened 2018-02-21
Add daemon parameter in config.

ashashwat:
In docs/config.md updating with the extra parameter.  
Keeping `argument` and `default` as empty.  Is there a better option?


osheroff:
empty is fine

PR #900, opened 2018-02-23
Added http_bind_address support in config, server, and docs. Fixes #899

### LINKED ISSUE #899: No config option to control what interfaces http metrics service binds to
I'd like to be able to configure the http metrics to only be available on a single interface (e.g. 127.0.0.1) but there's no option to do that right now, it binds to all interfaces.

macboy012:
Added support for configuring http bind address while remaining backwards compatible.


osheroff:
okey doke 👍

PR #906, opened 2018-02-27
add javascript filtering

osheroff:
This introduces javascript per-row hooks that are called before passing
a row off to the producer.  This lets users munge data, filter out
output based on arbitrary code, and choose their own kafka topics.  More
possibilities are obviously available, but it seems like a good place to
start.

TODO:  some testing, a bit more documentation.


src/main/java/com/zendesk/maxwell/scripting/WrappedRowMap.java
```diff
@@ -0,0 +1,77 @@
...
+	public String getTable() {
+		return row.getTable();
+	}
+
+	public String getDatabase() {
+		return row.getDatabase();
+	}
+
+	public String getPosition() {
+		BinlogPosition p = row.getPosition().getBinlogPosition();
```
ashashwat: NPE if row.getPosition() will ever be NULL.  Not sure if that can ever happen though.

docs/docs/config.md
```diff
@@ -95,6 +95,26 @@ you can use `exclude_columns` to filter out columns by name.
...
+		var username = row.data.get("username");
+		if ( username == "osheroff" )
+			row.suppress();
+
+		row.data.put("username", username.toUpperCase());
+	}
+}
+```
+
+There's a longer example [here]().
```
ashashwat: This link points to nowhere.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -533,6 +539,15 @@ private void setup(OptionSet options, Properties properties) {
 				}
 			}
 		}
+
+		if ( this.javascriptFile != null ) {
```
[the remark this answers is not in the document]
osheroff: nah, is cool.  let 'em name it whatever.

docs/docs/config.md
osheroff: yeah i will fix this up once I get it merged (& get a reliable github link)

osheroff:
@timbertson @liulikun any comments?

osheroff:
pingo pongo

liulikun:
A very interesting approach! I like the idea. Will the performance be an issue?

osheroff:
@liulikun sorry for the delay, I was in Montreal and dropped my laptop out of bed.  I'm going to write a little performance harness for maxwell as it'd be great to get decent benchmarks on stuff like this (as well as other performance work).

osheroff:
ok @liulikun I dusted off this code and started perf-testing it.  The overhead is quite low, something on the order of 2-4% for just entering the interpreter, calling a function, exiting.  Of course more complicated javascript would have different overheads.

osheroff:
@liulikun or @ericwush this is ready with tests and I'd love a look-see.  I wasn't exactly sure what to do with errors from the script engine, so I'm just letting maxwell crash for sure.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -151,6 +155,9 @@ public void work() throws Exception {
 		rowCounter.inc();
 		rowMeter.mark();
 
+		if ( scripting != null )
+			scripting.invoke(row);
+
```
liulikun: Should we skip `processRow` if it's suppressed?

liulikun:
Looks good! We can run some performance tests once it's merged/released.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
osheroff: the thing I'd be worried about is the user shooting themselves in the foot, suppressing everything including maxwell updates, etc.  we have lots of code in the filters to make sure that doesn't happen, but nothing here.

liulikun: 👍

osheroff:
@liulikun can you do another pass when you get a sec?  I made some changes to allow ddl rows and heartbeat rows work better.

liulikun:
Nice one! 👍

PR #907, opened 2018-03-02
Fix CREATE SCHEMA x CHARACTER SET = DEFAULT parsing bug

justinmichaud:
This test case came up while running maxwell. It is valid sql, so this change should allow it to be parsed correctly

Thanks!


osheroff:
great, thx for the bug fix 👍

PR #908, opened 2018-03-03
Add logging for binlog errors to make them easier to skip

justinmichaud:
This is related to the problem I had in the last PR. I am trying to figure out how to skip a bad binlog entry.

With this change, this is the process I followed to skip over a log entry:
- This error message was printed: `Error on bin log position Position[BinlogPosition[mysql-bin.000010:238759], lastHeartbeat=0]`
- `mysqlbinlog mysql-bin.000010 --verbose --base64-output=decode-rows | grep -C 100 "at 238759"` gave:
```
# at 238759
#180303  0:08:57 server id 1  end_log_pos 238900 	Query	thread_id=5	exec_time=0	error_code=0
SET TIMESTAMP=1520035737/*!*/;
SET @@session.sql_mode=0/*!*/;
CREATE SCHEMA IF NOT EXISTS my_sqhema CHARACTER SET = DEFAULT
/*!*/;
# at 238900
#180303  0:08:58 server id 1  end_log_pos 238971 	Query	thread_id=5	exec_time=0	error_code=0
SET TIMESTAMP=1520035738/*!*/;
BEGIN
/*!*/;
# at 238971
# at 239053
#180303  0:08:58 server id 1  end_log_pos 239053 	Table_map: `db1`.`table1` mapped to number 884
#180303  0:08:58 server id 1  end_log_pos 239109 	Write_rows: table id 884 flags: STMT_END_F
### INSERT INTO table1.db1
```
- I ran maxwell with `--init_position=mysql-bin.000010:238900:1520035738`

I am wondering:
- Is using the timestamp as a heartbeat the intended way?
- Since maxwell doesn't shut down properly, what can I do to ensure that every bin log entry before the bad one was processed properly and written to kafka at least once? How does this affect schema changes?

Thanks for being so responsive and helpful!

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSchemaStore.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSchemaStore.java
@@ -89,7 +89,14 @@ public class MysqlSchemaStore extends AbstractSchemaStore implements SchemaStore
 
 
 	public List<ResolvedSchemaChange> processSQL(String sql, String currentDatabase, Position position) throws SchemaStoreException, InvalidSchemaError {
-		List<ResolvedSchemaChange> resolvedSchemaChanges = resolveSQL(getSchema(), sql, currentDatabase);
+		List<ResolvedSchemaChange> resolvedSchemaChanges;
+		try {
+			resolvedSchemaChanges = resolveSQL(getSchema(), sql, currentDatabase);
+		} catch (Exception e) {
+			LOGGER.error("Error on bin log position " + position.toString());
+			e.printStackTrace();
+			throw e;
+		}
 
 		if ( resolvedSchemaChanges.size() > 0 ) {
 			try {


osheroff:
skipping is generally not safe;  any further changes to `my_schema` will cause maxwell to crash if it misses this statement.  As to your questions:

> Is using the timestamp as a heartbeat the intended way?

Maxwell doesn't use server timestamps as heartbeats, it instead generates its own timestamp.  If you're curious look for updates to `maxwell.heartbeats`

> Since maxwell doesn't shut down properly, what can I do to ensure that every bin log entry before the bad one was processed properly and written to kafka at least once? How does this affect schema changes?

you'd just have to check kafka.  I believe the kafka producer makes a best-effort to flush its buffer on shutdown, but when maxwell is crashing there's really no guarantees.

But again, see earlier point: if you try to skip the statement things will go generally poorly.  you could in theory fake a DDL event in `maxwell.schemas` that would allow you to continue, but... eh.

justinmichaud:
Thanks for the response. I would like to submit a PR to update the documentation for this (if that is OK with you), so I really appreciate it.

I am trying to figure out what to do if maxwell encounters a log entry that it cannot understand in production without data loss. In this case, the new schema was blacklisted, so we don't really care if maxwell skips it (we are looking into making a PR to fix that soon as well, to reduce the possibility of crashing).

I have a few questions:

1) What are your thoughts about the current PR? 
2) If something unexpected comes up and we do need to skip over a bad binlog entry, what should we use for the heartbeat to pass to --init_position? If we get the heartbeat wrong, what will happen?
3)
> you'd just have to check kafka. I believe the kafka producer makes a best-effort to flush its buffer on shutdown, but when maxwell is crashing there's really no guarantees.

When maxwell crashes, is it guaranteed that everything up until the last entry in the positions table was sent to kafka? 

If so, would you accept a PR for a skip_binlogs option to allow specifying a list of positions to skip? Also, in that case, would the following procedure work without data loss?
a) Check the crash log from this pr to find the bad binlog entry. 
b) Make sure no more schema changes are happening, and get a list of all of the binlog positions involving the bad schema
c) Save the last heartbeat and binlog position from maxwell. Truncate the maxwell database (to wipe any bad state?)
d) Restart maxwell, with init_position from above and passing the list of bad binlog positions as skip_binlogs

Thanks again for the help. I really appreciate it.

osheroff:
1.  It’s fine, mergable.
2.  Re: heartbeat, I have to check the code.  I’m not *quite* sure why it got required for init_position.  

> When maxwell crashes, is it guaranteed that everything up until the last entry in the positions table was sent to kafka?

Yeah, we don’t write to positions until Kafka acks.

osheroff:
Re: skip_positions:  I’m conflicted.  I’ve thought about trying out an “auto-skip” feature where if maxwell can’t process DDL it’ll immediately attempt to re-capture the database, which would save us from crashing on parser errors, but then my fear is we’d never get the bug reports we need to fix it.

At the very least, we should try to flush all outstanding rows in the Kafka buffer before crashing so that the binlog pointer is precisely at the crash SQL and could be safely skipped in cases like yours.

osheroff:
I'm gonna merge this.  If you wanna continue chatting about this one, can open an issue.

Issue #904, opened 2018-02-24
When filtering, the db or table name is case-sensitive in config.properties file,so the filtering is failed.


Hackeruncle:
Maxwell version: 1.10.5~1.13.1 (tested)
Current Maxwell version: 1.10.8

**1.DB Name is upper in config.properties**
include_dbs=TEST
include_tables=timezonetest

**2.Table Name is upper in config.properties**
include_dbs=test
include_tables=TIMEZONETEST

**3.Test**
 MySQL SQL: insert into timezonetest(date_c,datetime_c,timestamp_c) values('2018-02-25','2018-02-25','2018-02-25 10:00:00');

**4.Run Maxwell producer as stdout
_The console will not print the sql json,is issue._**

**5.Modify MaxwellFilter.java**
**Maxwell-904.patch**
[Maxwell-904.patch.zip](https://github.com/zendesk/maxwell/files/1754225/Maxwell-904.patch.zip)

6.Test is ok.

PR #909, opened 2018-03-04
filter-case

Hackeruncle:
https://github.com/zendesk/maxwell/issues/904

Maxwell version: 1.10.5~1.13.1 (have tested)
Current Maxwell version: master

**MySQL(Linux) :**
default: 
lower_case_table_names = 0, is case sensitive.

but dba will set lower_case_table_names = 1 in product env, isn't case sensitive.
create db "test" and table "timezonetest".

1.DB Name is upper in Maxwell config.properties
include_dbs=TEST
include_tables=timezonetest

2.Table Name is upper in config.properties
include_dbs=test
include_tables=TIMEZONETEST

3.Test
MySQL SQL: insert into timezonetest(date_c,datetime_c,timestamp_c) values('2018-02-25','2018-02-25','2018-02-25 10:00:00');

4.Run Maxwell producer as stdout
The console will not print the sql json,is issue.

5.Modify MaxwellFilter.java
Maxwell-904.patch
https://github.com/zendesk/maxwell/files/1754225/Maxwell-904.patch.zip

6.Test is ok.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellFilter.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellFilter.java
@@ -50,32 +50,38 @@ public class MaxwellFilter {
 	) throws MaxwellInvalidFilterException {
 		if ( includeDatabases != null ) {
 			for ( String s : includeDatabases.split(",") )
-				includeDatabase(s);
+				//includeDatabase(s);
+			    includeDatabase(s.toUpperCase()); // toUpperCase
 		}
 
 		if ( excludeDatabases != null ) {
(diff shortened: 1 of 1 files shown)


osheroff:
this isn't correct for *most* setups - this is really correct only on windows/osx, or when `lower_case_table_names` is explicitly set.  If you want to make this work, you'll have to make it conditional on that system variable.

osheroff:
Hey @Hackeruncle do you have interest in finishing this PR?  As I said, the idea is somewhat correct but it needs to take into account whether the server is running case sensitive or not.

Issue #780, opened 2017-10-13
com.zendesk.maxwell.schema.ddl.InvalidSchemaError: Couldn't find table 'wms_policy_query' in database wms


Hackeruncle:
**Current architecture:**
MySQL(master) --> MySQL(slave) -->maxwell -->kafka

The architecture running of a long time is ok.
But ,when creating a **new table**  of 'wms_policy_query'  in **MySQL(master)**, then maxwell process is stopped.

**These error log:**
```
09:39:54,504 INFO  ContextHandler - Stopped o.e.j.s.ServletContextHandler@22f0ef3f{/,null,UNAVAILABLE}
com.zendesk.maxwell.schema.ddl.InvalidSchemaError: Couldn't find table 'wms_policy_query' in database wmsb35
        at com.zendesk.maxwell.schema.Database.findTableOrThrow(Database.java:55)
        at com.zendesk.maxwell.schema.ddl.TableAlter.resolve(TableAlter.java:37)
        at com.zendesk.maxwell.schema.ddl.TableAlter.resolve(TableAlter.java:12)
        at com.zendesk.maxwell.schema.AbstractSchemaStore.resolveSQL(AbstractSchemaStore.java:58)
        at com.zendesk.maxwell.schema.MysqlSchemaStore.processSQL(MysqlSchemaStore.java:92)
        at com.zendesk.maxwell.replication.AbstractReplicator.processQueryEvent(AbstractReplicator.java:104)
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.processQueryEvent(BinlogConnectorReplicator.java:300)
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getRow(BinlogConnectorReplicator.java:275)
        at com.zendesk.maxwell.replication.AbstractReplicator.work(AbstractReplicator.java:157)
        at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:27)
        at com.zendesk.maxwell.Maxwell.startInner(Maxwell.java:183)
        at com.zendesk.maxwell.Maxwell.start(Maxwell.java:134)
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:204)
09:39:55,852 INFO  TaskManager - Stopped all tasks
```

osheroff:
this is odd;  this is definitely a table alter, not a table create;  any idea of the SQL statements that preceded this crash?

bokelo:
[...]

PR #914, opened 2018-03-09
Fix maxwell process DDL twice 

dongxiaohe:
Related to issue https://github.com/zendesk/maxwell/issues/780

Schema chain building conflicts with message processing on DDL.

In more details: 

What happen is that when Maxwell starts, it will compare the current position with schema binlog position and then build the schema linked list. This behaviour is correct and helps maxwell to reproduce all the db changes at any point of time. 

If the binlog position is DML, it is working fine. However, there is an edge case, if the binlog position is pointing to a DDL statement, maxwell would include this DDL statement into the schema chain. After that, maxwell will be unable to reprocess this statement again in `BinlogConnectorReplicator.processQueryEvent`. i.e. it couldn't create same table twice, it couldn't alter the same table twice and etc.

Solution: Use the parent schema from the latest if `binlog_position` and `target_position` are same.

Here are other solutions I can think of, but I feel like the drawback of it are pretty clear.
- Fix it in the `BinlogConnectorReplicator.processQueryEvent`: I don't think we should leave it at processing stage, it is very late and error prone.
- Point to previous binlog position: If it is DML, the same message would be reprocessing twice.

Any other ideas @zendesk/goanna  ?

@osheroff I am sorry to ask you review it, it is little abrupt, but this is my first code change to maxwell, I hope it won't break anything :)

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -575,7 +575,7 @@ public class MysqlSavedSchema {
 			PreparedStatement s = connection.prepareStatement(
 				"SELECT id from `schemas` "
 				+ "WHERE deleted = 0 "
-				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position <= ?)) AND server_id = ? "
+				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position < ?)) AND server_id = ? "
 				+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1");
 
 			s.setLong(1, targetPosition.getLastHeartbeatRead());
--- a/src/test/java/com/zendesk/maxwell/MysqlSavedSchemaTest.java
+++ b/src/test/java/com/zendesk/maxwell/MysqlSavedSchemaTest.java
@@ -57,9 +57,7 @@ public class MysqlSavedSchemaTest extends MaxwellTestWithIsolatedServer {
 		this.position = MaxwellTestSupport.capture(server.getConnection());
 		this.context = buildContext(position);
 		this.schema = new SchemaCapturer(server.getConnection(), context.getCaseSensitivity()).capture();
-		this.savedSchema = new MysqlSavedSchema(this.context.getServerID(), this.context.getCaseSensitivity(),
-				this.schema, makePosition(position.getBinlogPosition().getOffset() - 1L,
-				position.getBinlogPosition().getFile(), position.getLastHeartbeatRead()));
+		this.savedSchema = SavedSchemaSupport.getSavedSchema(this.context, this.schema, this.position);
 	}
 
 	@Test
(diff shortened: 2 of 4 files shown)


osheroff:
@dongxiaohe, was this something you encountered in production or just a theoretical repro?  I can kind of see happening in some worlds, but... hm... seems a bit unlikely.  For this to happen (and correct me if I'm wrong), it'd have to be:

- maxwell has already processed the DDL at `binlog:456`, but *not* advanced maxwell's position to `binlog:491`.  Now maxwell crashes, leaving the position pointer in place, and yeah, when we restart, we pick up the schema twice and things go wrong. 

To be clear: your fix appears correct, I'm just a bit dubious about the frequency of the edge case.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -573,7 +573,7 @@ private static Long findSchema(Connection connection, Position targetPosition, L
 			// Only consider binlog positions before the target position on the current server.
 			// Within those, sort for the latest binlog file, then the latest binlog position.
 			PreparedStatement s = connection.prepareStatement(
-				"SELECT id from `schemas` "
+				"SELECT id, binlog_position, base_schema_id from `schemas` "
 				+ "WHERE deleted = 0 "
 				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position <= ?)) AND server_id = ? "
```
osheroff: I think this whole change could actually be boiled down to changing `binlog_position <= ?` to `binlog_position < ?`, couldn't it?

dongxiaohe:
@osheroff  Yeah, it happened in the production. After that, I can reproduce the issue locally via zdi.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
dongxiaohe: Nice !. Good thought, previously I thought this change may reproduce the same DML twice. On a second, this logic is only to find the latest schema change. I assume this should fix this edge case. I will amend the pull request and test it later on. 👍

src/test/java/com/zendesk/maxwell/MysqlSavedSchemaTest.java
```diff
@@ -56,7 +57,9 @@ public void setUp() throws Exception {
 		this.position = MaxwellTestSupport.capture(server.getConnection());
 		this.context = buildContext(position);
 		this.schema = new SchemaCapturer(server.getConnection(), context.getCaseSensitivity()).capture();
-		this.savedSchema = new MysqlSavedSchema(this.context, this.schema, position);
+		this.savedSchema = new MysqlSavedSchema(this.context.getServerID(), this.context.getCaseSensitivity(),
```
dongxiaohe: This change is to fix all other tests.

src/test/java/com/zendesk/maxwell/SavedSchemaSupport.java
```diff
@@ -0,0 +1,20 @@
...
+final public class SavedSchemaSupport {
+	private SavedSchemaSupport() { }
+
+	public static MysqlSavedSchema getSavedSchema(MaxwellContext context, Schema schema, Position position) throws Exception {
+		if (context.getConfig().gtidMode) {
+			return new MysqlSavedSchema(context, schema, position);
+		}
+
+		return new MysqlSavedSchema(context.getServerID(), context.getCaseSensitivity(),
+				schema, new Position(new BinlogPosition(position.getBinlogPosition().getOffset() - 1L,
```
osheroff: sorry, I'm kinda confused;  why do we need the `getOffset() - 1` bits?

dongxiaohe: Hi @osheroff , thanks for the question. This is only for `testing data`. The purpose of adding `- 1` is to make sure the seeding saved schema's position is prior to the current position. It fixed lots of failed tests in `MysqlSavedSchemaTest` and `RecoveryTest` (Previously, the saved schema is same position with current position. (lots of null pointer exception in the tests)). :)

osheroff: ok.  it's a little nasty, but ok.

Issue #915, opened 2018-03-09
bug: set type PrimaryKey exception


chen56:
version : v1.13.2

run this test:

```java
	@Test
	public void testSetTypePrimaryKey() throws Exception {
		List<RowMap> list;
		String before[] = { "create table pksen (Id set('android','iphone','ipad'), primary key(ID))" };
		String input[] = {"insert into pksen set id ='android'"};
		String expectedJSON = "{\"database\":\"shard_1\",\"table\":\"pksen\",\"pk.id\":[\"android\"]}";
		list = getRowsForSQL(null, input, before);
		assertThat(list.size(), is(1));
		assertThat(list.get(0).pkToJson(RowMap.KeyFormat.HASH), is(expectedJSON));
	}
```

output:
```bash

    java.lang.IllegalStateException: No ObjectCodec defined for the generator, can only serialize simple wrapper types (type passed java.util.ArrayList)
	at com.fasterxml.jackson.core.JsonGenerator._writeSimpleObject(JsonGenerator.java:1725)
	at com.fasterxml.jackson.core.base.GeneratorBase.writeObject(GeneratorBase.java:327)
	at com.fasterxml.jackson.core.JsonGenerator.writeObjectField(JsonGenerator.java:1415)
	at com.zendesk.maxwell.row.RowMap.pkToJsonHash(RowMap.java:155)
	at com.zendesk.maxwell.row.RowMap.pkToJson(RowMap.java:134)
	at com.zendesk.maxwell.MaxwellIntegrationTest.testSetTypePrimaryKey(MaxwellIntegrationTest.java:117)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #916, opened 2018-03-09
bugfix : bug: set type PrimaryKey exception #915

chen56:
https://github.com/zendesk/maxwell/issues/915

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowMap.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowMap.java
@@ -21,6 +21,18 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Pattern;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -147,7 +159,7 @@ private String pkToJsonHash() throws IOException {
 				if ( data.containsKey(pk) )
 					pkValue = data.get(pk);
 
-				g.writeObjectField("pk." + pk.toLowerCase(), pkValue);
+				writeValueToJSON(g, false, "pk." + pk.toLowerCase(), pkValue);
```
osheroff: I think `includeNullField` should be `true` here, no?

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -170,7 +182,7 @@ private String pkToJsonArray() throws IOException {
 				pkValue = data.get(pk);
 
 			g.writeStartObject();
-			g.writeObjectField(pk.toLowerCase(), pkValue);
+			writeValueToJSON(g, false, pk.toLowerCase(), pkValue);
```
osheroff: here too

osheroff:
thx for the PR, looks generally good and will merge after sorting out `includeNullField`

PR #936, opened 2018-04-05
Add --skip_positions option

justinmichaud:
This PR adds a --skip_positions=file1:123-456,file2:111-111 option. The intent behind this option is to allow maxwell to safely skip over a log entry in the event of a bug.

Question for reviewer: Currently, `testThatSkipPositionsDoesNotParseCrashes` does not work as intended. The intent here is to make sure that maxwell does not try to parse the binlog. The test is not strictly necessary, but if you can think of a way to test this, please let me know. I manually tested this by reverting https://github.com/zendesk/maxwell/pull/907/files, to give you an idea of the case that I am worrying about. If there was some way of injecting an invalid binlog into the stream that maxwell reads, that would be great.

Let me know if you want any changes. Thanks!

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -14,14 +14,12 @@ import joptsimple.BuiltinHelpFormatter;
 import joptsimple.OptionDescriptor;
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
+import org.apache.commons.lang3.Range;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
 import java.util.regex.Pattern;
 
 public class MaxwellConfig extends AbstractConfig {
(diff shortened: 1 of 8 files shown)


justinmichaud:
Hey @osheroff, do you have any ideas for how I can test that this does no parsing?

justinmichaud:
Hey @osheroff, are you still willing to merge this feature?

liulikun:
Feel a bit heavy handed to handle this situation since the config is checked on every single event even it may only happen occasionally. What about manually updating `positions` table on those occasions to skip the position?

osheroff:
sorry for the long delay @justinmichaud I've had this PR in the back of my mind for awhile.  I have a second use for this PR in wall-papering over some race conditions in the `SchemaCapturer`, but I think what I'd like to see is instead of a list of positions to skip, make this instead "ignore-schema-errors-through".  The premise is that you'd pick a target binlog position, and maxwell would ignore any ddl-errors that result until you've processed through that position.  It'd be roughly analogous to the `skip-slave-errors` option in mysql.

This would have the side-benefit of only checking the variable when the ddl throws.

Would that be an ok approach to solving your problem?

justinmichaud:
Hmm, I'm not sure. I was hoping to use this in case of another parsing error, but I suppose those are quite rare so it might be fine.

If we would cache the check that the config option is enabled, would that solve your concerns about performance?

osheroff:
So I have less of a performance concern, I just want to be able to ignore DDL errors between two different positions - in my code I won’t know where the specific error is (if there is any)

Issue #935, opened 2018-04-04
Support environment variable based configurations


ericwush:
/cc @zendesk/goanna 

We'd like to bring this up and seek feedback especially from @osheroff regarding supporting environment variable based configurations.

Currently Maxwell supports configurations to be loaded from command line arguments or properties file. In our case, we've got heaps of Kafka configs which are not currently supported and extendable when command line arguments approach is used. Automating the process to compose the properties file is always an option but it could be simplified if Maxwell supports environment variable based configurations, e.g. with a prefix `MAXWELL_`. This could be useful in containers too.

Any thoughts/concerns?

osheroff:
Is there much precedent for this?  I guess some gnu utils allow *some* env based config, I've never heard of a program that can fully configure itself via env.  In theory I don't mind the idea, definitely fits with the way the world is heading wrt containers and such. 

wondering if there's any security risks?  someone upstream injects variables into your env and you don't realize it?  `sudo` seems very concerned about that sort of thing, I'm not sure if we'd need to be.....

ericwush:
@osheroff 

> definitely fits with the way the world is heading wrt containers and such
Exactly. In our case, some configs come from env vars directly, like _mysql_host_, _kafka_host_. The others might be externalised to env vars for easy configuration, like kafka related settings. Rather than env vars -> config file -> maxwell, we think env vars -> maxwell directly should make life easier.

> someone upstream injects variables into your env and you don't realize it?
That was my initial concern, but the prefix should minimise the chance of it. Also the prefix should be configurable too.

I can probably start with a PR and see how it tastes.

PR #937, opened 2018-04-05
MELINF-373: support environment variable based configuration

ericwush:
/cc @zendesk/goanna @osheroff 

Address #935 .

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -18,11 +18,10 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
+import java.util.function.Supplier;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 public class MaxwellConfig extends AbstractConfig {
 	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellConfig.class);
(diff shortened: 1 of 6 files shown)


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -318,6 +319,12 @@ private void parse(String [] argv) {
 			properties = parseFile(DEFAULT_CONFIG_FILE, false);
 		}
 
+		if (options.has("env_config_prefix")) {
+			String prefix = ((String) options.valueOf("env_config_prefix")).toLowerCase();
+			System.getenv().entrySet().stream().filter(map -> map.getKey().toLowerCase().startsWith(prefix))
+					.forEach(config -> properties.put(config.getKey().toLowerCase().replace(prefix, ""), config.getValue()));
```
xuebingli: nit: use `replaceFirst` instead of `replace`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -318,6 +319,12 @@ private void parse(String [] argv) {
 			properties = parseFile(DEFAULT_CONFIG_FILE, false);
 		}
 
+		if (options.has("env_config_prefix")) {
+			String prefix = ((String) options.valueOf("env_config_prefix")).toLowerCase();
+			System.getenv().entrySet().stream().filter(map -> map.getKey().toLowerCase().startsWith(prefix))
```
xuebingli: What if the user makes a typo and have env like `maxwell_passwod`? Should we defend/filter against this scenario and/or throw a warning?

ericwush: Probably not, because if using config file you'll face the same problem. And the change uses the same mechanism as config file.

osheroff:
I'd say let's go ahead and make `env_config_prefix` available in the configuration file too.  Other than that it looks good to me.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -318,6 +318,19 @@ private void parse(String [] argv) {
...
 		}
 
+		Optional<String> prefix = Stream.<Supplier<Optional<String>>>of(
+				() -> Optional.ofNullable((String) options.valueOf("env_config_prefix")),
+				() -> Optional.ofNullable(properties.getProperty("env_config_prefix")))
+				.map(Supplier::get)
+				.filter(Optional::isPresent)
+				.map(Optional::get)
+				.map(String::toLowerCase)
+				.findFirst();
```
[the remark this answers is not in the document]
ericwush: I appreciate the functional way java provides but yeah, you're right, the boilerplate doesn't improve the readability at all...

ericwush:
Kinda being blocked by this, will merge and address later if there is further feedback.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -329,6 +338,11 @@ private void parse(String [] argv) {
 		}
 	}
 
+	private String getEnvConfigPrefix(OptionSet options, Properties properties) {
+		String prefix = (String) options.valueOf("env_config_prefix");
+		return prefix != null ? prefix : properties.getProperty("env_config_prefix");
+	}
+
```
osheroff: this is fine, but I think you can also just use `fetchOption`, no?

Issue #948, opened 2018-04-22
Can maxwell do bootstrapping with specific maxwell client id?


tanapoln:
I just wonder that if I run multiple instances of Maxwell with a single master and I want to do bootstrapping data, how can I specify which client id to do that so that other Maxwell instances would just do their job and subscribes would not have to process rows.

If it's not possible at this time, is there an alternative way to achieve that behavior?

Thank you.

osheroff:
this is not currently possible, but clearly needed.

osheroff:
this is possible in 1.14.1

PR #951, opened 2018-04-25
client_ids in bootstrapping

osheroff:
addresses #948


src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -6,6 +6,7 @@
 import com.zendesk.maxwell.replication.BinlogPosition;
 import com.zendesk.maxwell.producer.MaxwellOutputConfig;
 import com.zendesk.maxwell.replication.Position;
+import com.zendesk.maxwell.schema.Schema;
```
ericwush: Nit: unused?

osheroff:
ok, removed unused include, docs updated, tests fixed.

PR #959, opened 2018-05-01
Use a smaller docker base image to have a final docker image 1GB smaller

ndeslandes:
The actual zendesk/maxwell image is 1.13GB.
I am proposing to change the maven base image to the openjdk-alpine image, install the build dependencies, build and then cleanup.
The resulting docker image is 154MB

```$ docker images
REPOSITORY                                                    TAG                 IMAGE ID            CREATED                  SIZE
zendesk/maxwell                                               smaller             211419470306        Less than a second ago   154MB
zendesk/maxwell                                               latest              8b3c07c16cca        6 days ago               1.13GB```

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -1,17 +1,16 @@
-FROM maven:3.5-jdk-8
+FROM openjdk:8u151-jdk-alpine
```
liulikun: I heard there were some issues with `snappy` on alpine before. FYI `snappy` can be configured to turn on compression for kafka producer.

ndeslandes: we can keep the maven image and do a cleanup after instead
it saves about 300MB

ndeslandes:
New version is 497MB @liulikun

osheroff:
I'm fine with this, @liulikun you know more and can merge at will

ericwush:
@ndeslandes well, it turns out you can still use alpine but have to install the package. Details below:
- apk add the `java-snappy-native` package
- in `bin/maxwell`'s java command, add `-Dorg.xerial.snappy.use.systemlib=true`

I was referring to https://stackoverflow.com/a/45351085 and have verified working. Personally I like using alpine because the less it contains, the less vulnerable it is. You can test and see how much it saves in terms of the size and decide whether it is worth (my testing image contains more than only maxwell so I don't know the exact size).

liulikun:
Nice work @ndeslandes, it's already an improvement.
Up to you if you want to incorporate @ericwush suggestions.

Dockerfile
```diff
@@ -1,17 +1,16 @@
...
 ENV MAXWELL_VERSION=1.14.3 KAFKA_VERSION=0.11.0.1
 
 COPY . /workspace
 
-RUN apt-get update \
-    && apt-get -y upgrade \
-    && apt-get install -y build-essential \
+RUN apk --no-cache add --virtual .build-dependencies make maven \
+    && apk --no-cache add java-snappy-native \
```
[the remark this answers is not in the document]
ndeslandes: If i do that it will be [art of the virtual .build-dependencies and be removed at the end

PR #960, opened 2018-05-01
always pass maxwell.heartbeats updates through

### LINKED ISSUE #884: Maxwell does not setPosition in some situations, leading to errors
In normal situation, maxwell captures an event from binlog, send it to producer then use context.setPosition to update binlog position in maxwell.position table.
However in the following cases maxwell does not update the position table:
1 For any DDL event.
2 if the filter include some dbs/tables, maxwell will only set position for the filtered tables/dbs.
The consequences of 1&2 are that maxwell's position is sometime behind the latest binlog, and in some cloud services like Aliyun, the binlog file maxwell positioning to are often purged, leading to maxwell reported errors(fail to read binlog file).

So I proposed some fixes:
For 1, when maxwell send DDL event to producer, it should also set positions.
For 2,  maxwell may keep setting positions only for filtered tables. However when maxwell restart, if it find the binlog is purged (it is positioning to a binlog file which is offline) , it should look for the latest binlog file.

I am new to maxwell community. Would like to hear comments on the fixes above. Thanks.

osheroff:
this may fix #884

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/AbstractReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/AbstractReplicator.java
@@ -118,7 +118,7 @@ public abstract class AbstractReplicator extends RunLoopProcess implements Repli
 	 */
 	protected boolean shouldOutputEvent(String database, String table, MaxwellFilter filter) {
 		Boolean isSystemWhitelisted = this.maxwellSchemaDatabaseName.equals(database)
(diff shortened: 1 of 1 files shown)


osheroff:
Yup you’ve got it right.

leonjoyce:
The solution above only works for scenario when maxwell schema DB is the same as replication DB.
If they are different, then the position will not bump.
@osheroff please confirm.

PR #961, opened 2018-05-02
fixes for detecting replication-hosts and maxwell-hosts as the same

osheroff:
pr #958 introduced a regression that disabled bootstrapping due to a bad equivalence check.


osheroff:
gonna merge this w/out +1 as I need to get the fix out ahead of folks downloading 1.4.1

Issue #950, opened 2018-04-25
NullPointerException in SchemaCapturer 


BradyPT:
```
12:26:09,054 DEBUG MysqlSavedSchema - 12/116 Capturing DOC_goals...
java.lang.NullPointerException
    at com.zendesk.maxwell.schema.SchemaCapturer.captureTablesPK(SchemaCapturer.java:214)
    at com.zendesk.maxwell.schema.SchemaCapturer.captureTables(SchemaCapturer.java:196)
    at com.zendesk.maxwell.schema.SchemaCapturer.captureDatabase(SchemaCapturer.java:140)
    at com.zendesk.maxwell.schema.SchemaCapturer.capture(SchemaCapturer.java:111)
    at com.zendesk.maxwell.schema.AbstractSchemaStore.captureSchema(AbstractSchemaStore.java:44)
    at com.zendesk.maxwell.schema.MysqlSchemaStore.restoreOrCaptureSchema(MysqlSchemaStore.java:68)
    at com.zendesk.maxwell.schema.MysqlSchemaStore.getSchema(MysqlSchemaStore.java:58)
    at com.zendesk.maxwell.Maxwell.start(Maxwell.java:169)
    at com.zendesk.maxwell.Maxwell.main(Maxwell.java:203)
```

I think it's possible we're experiencing a race condition between the captureTables query and captureTablesPK query, it ran fine for the longest time before. This is the line that fails:

table.setPKList(l.get(key));

osheroff:
ping, can you reproduce this issue @BradyPT?

BradyPT:
Not reliably, we're not entirely sure what's going on either that's just our best guess, but it's happened multiple times. It's either the table.setPKList or the l.get, but I'm having trouble finding out how that's possible as well.

osheroff:
What version are you running?  Possible mismatch between the stack-trace and the code we're looking at? 

doing `l.get(null)` wouldn't throw an NPE, `l` is clearly not null, and I can't see a case where `tables` would contain a null value either.

BradyPT:
1.8.1, but the schemacapturer hasn't been updated since. I'll look at updating if you think that'll fix it.

osheroff:
your stack traces don't line up to either 1.8.1 or HEAD.  have you thrown debug logging in there?  In v1.8.1 the call to `captureTablesPK` should have been on line 189, but your stack trace has it on 196.  

[...]

PR #963, opened 2018-05-04
Fix one race in SchemaCapturer

osheroff:
In this race, a table is added between when we capture the tables and
when we capture the PKs.  This race, when encountered, will not cause a
future crash as we will simply process the `CREATE TABLE` after the
schema is captured.

Addresses #950.

Note that this is not the end of race conditions inside `SchemaCapturer`; other races will cause crashes after processing the racing DDL, or, in some cases, a corrupt view of the schema.


osheroff:
@ericwush agree that the streams approach is more elegant/compact, but yeah, I'm gonna merge this as-is and we can do a java-8-ify pass at the whole file if we see fit.

PR #969, opened 2018-05-10
Basic benchmark harness

osheroff:
this is a bit of code that boots up a server, generates data, and then
leaves that data in place for you to run performance tests upon.
For now we use a custom producer to know when to shutdown, but in the
long run it might be useful to be able to benchmark other producers,
too. 

As of for now I've left the code in the `test` package, as I'm not sure there's a reason to distribute it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -264,6 +264,11 @@
       <artifactId>aws-java-sdk-sqs</artifactId>
       <version>1.11.68</version>
     </dependency>
+    <dependency>
+      <groupId>me.tongfei</groupId>
+      <artifactId>progressbar</artifactId>
+      <version>0.6.0</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/src/test/onetimeserver
+++ b/src/test/onetimeserver
@@ -107,7 +107,7 @@ then
   echo "exec: " $CACHE_DIR/wrapper $DEBUG $BLOCK $MYSQL_VERSION -ppid $PARENT_PID -type mysql -- $EXTRA_ARGS 1>&2
 fi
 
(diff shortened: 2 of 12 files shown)


pom.xml
```diff
@@ -264,6 +264,11 @@
       <artifactId>aws-java-sdk-sqs</artifactId>
       <version>1.11.68</version>
     </dependency>
+    <dependency>
+      <groupId>me.tongfei</groupId>
+      <artifactId>progressbar</artifactId>
+      <version>0.6.0</version>
+    </dependency>
```
liulikun: `test` scope?

src/test/onetimeserver
```diff
@@ -82,9 +94,24 @@ then
...
+fi
+
 if [ "$DEBUG" == "-debug" ]
 then
-  echo "exec: " $CACHE_DIR/wrapper $DEBUG $MYSQL_VERSION -ppid $PPID -type mysql -- $EXTRA_ARGS 1>&2
+  echo "exec: " $CACHE_DIR/wrapper $DEBUG $BLOCK $MYSQL_VERSION -ppid $PARENT_PID -type mysql -- $EXTRA_ARGS 1>&2
 fi
 
-exec $CACHE_DIR/wrapper $DEBUG $MYSQL_VERSION -ppid $PPID -type mysql -- $EXTRA_ARGS
+GO_ARGS="$DEBUG $DEBUG $BLOCK $MYSQL_VERSION $NO_CLEAN $REUSE -ppid $PARENT_PID -type mysql"
```
liulikun: `$DEBUG` twice?

pom.xml
osheroff: tried that, unfortunately that seemed to dump dependencies only into ~.m2 and it became super hard to run.

PR #970, opened 2018-05-11
improve performance of appendFractionalSeconds

osheroff:
this little change is worth a stunning performance increase in
maxwell-benchmark, from 46k rows/s to 56k rows/second.  I grant you
that this is a benchmark that calls this code about 20mm times, but...
still.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
+++ b/src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
@@ -1,10 +1,5 @@
 package com.zendesk.maxwell.schema.columndef;
 
-import java.sql.Time;
-import java.sql.Timestamp;
-import java.math.BigDecimal;
-import java.math.BigInteger;
-
 public abstract class ColumnDefWithLength extends ColumnDef {
 	protected Long columnLength;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
```diff
@@ -65,13 +52,15 @@ protected static String appendFractionalSeconds(String value, int nanos, Long co
...
 		StringBuilder result = threadLocalBuilder.get();
 		result.append(value);
 		result.append(".");
-		result.append(strFractional);
 
+		String asStr = String.valueOf(fractional);
+		for ( int i = 0 ; i < columnLength - asStr.length(); i++ )
+			result.append("0");
+
+		result.append(asStr);
```
liulikun: Looks good! Though a unit test will help to iron out some boundary issues.

liulikun:
Wow, that's amazing result.

src/main/java/com/zendesk/maxwell/schema/columndef/ColumnDefWithLength.java
osheroff: bunch of unit tests in `ColumnDefTest` already...

liulikun: 👍

PR #971, opened 2018-05-11
support predefined database name to partition mapping for kafka producer

liulikun:
When using the default kafka partitioner, the partition is computed as 
```
Math.abs(hashFunc.hashCode(databaseName) % numPartitions);
```
Which is hard to control which partition for the events from one database to land on. Using a property file we can define the partition for each database explicitly. 

/cc @osheroff @zendesk/goanna

### DIFF
--- a/maxwell-partitions.properties.example
+++ b/maxwell-partitions.properties.example
@@ -1,0 +1,4 @@
+#database name to partition mapping
+db1=1
+db2=2
+db3=3
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -50,6 +50,8 @@ public class MaxwellConfig extends AbstractConfig {
 	public String kafkaPartitionKey;
 	public String kafkaPartitionColumns;
 	public String kafkaPartitionFallback;
+	public String kafkaPartitionConfigFile;
+	public Map<String, Integer> kafkaPartitionMap;
 	public String bootstrapperType;
 	public int bufferedProducerSize;
 
@@ -363,6 +365,7 @@ public class MaxwellConfig extends AbstractConfig {
 		this.kafkaPartitionKey  	= fetchOption("kafka_partition_by", options, properties, null);
 		this.kafkaPartitionColumns  = fetchOption("kafka_partition_columns", options, properties, null);
 		this.kafkaPartitionFallback = fetchOption("kafka_partition_by_fallback", options, properties, null);
+		this.kafkaPartitionConfigFile = fetchOption("kafka_partition_config_file", options, properties, null);
 
 		this.kafkaPartitionHash 	= fetchOption("kafka_partition_hash", options, properties, "default");
 		this.ddlKafkaTopic 		    = fetchOption("ddl_kafka_topic", options, properties, this.kafkaTopic);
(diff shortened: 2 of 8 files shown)


src/main/java/com/zendesk/maxwell/producer/partitioners/AbstractMaxwellKafkaPartitioner.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.zendesk.maxwell.row.RowMap;
+
+public abstract class AbstractMaxwellKafkaPartitioner extends AbstractMaxwellPartitioner {
+
+	public AbstractMaxwellKafkaPartitioner(String partitionKey, String csvPartitionColumns, String partitionKeyFallback) {
+		super(partitionKey, csvPartitionColumns, partitionKeyFallback);
+	}
+
+	public abstract int kafkaPartition(RowMap r, int numPartitions);
```
osheroff: this looks like maybe `AbstractMaxwellKafkaPartitioner` wants to be an interface rather than an abstract class

osheroff: or nothing whatsoever, it'd be fine to have the two kafka partitioners inherit directly from  `AbstractMaxwellPartitioner`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -517,6 +520,15 @@ private void setup(OptionSet options, Properties properties) {
...
 		}
+
+		if (this.kafkaPartitionConfigFile != null) {
+			kafkaPartitionMap = new HashMap<>();
+			Properties partitionProps = readPropertiesFile(this.kafkaPartitionConfigFile, true);
+
+			for (String databaseName: partitionProps.stringPropertyNames()) {
+				kafkaPartitionMap.put(databaseName, Integer.parseInt(partitionProps.getProperty(databaseName)));
+			}
+		}
```
osheroff: this code should be moved to `validate` -- `setup` is not called in embedded mode.

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaDBNamePartitioner.java
```diff
@@ -0,0 +1,20 @@
...
+	private Map<String, Integer> kafkaPartitionMap;
+
+	public MaxwellKafkaDBNamePartitioner(Map<String, Integer> kafkaPartitionMap) {
+		super(null, null, null);
+		this.kafkaPartitionMap = kafkaPartitionMap;
+	}
+
+	@Override
+	public int kafkaPartition(RowMap r, int numPartitions) {
+		return kafkaPartitionMap.get(r.getDatabase());
```
osheroff: please account for the following cases:

- database doesn't exist in `kafkaPartitionMap`
- invalid partition number in `kafkaPartitionMap`

maxwell-partitions.properties.example
```diff
@@ -0,0 +1,4 @@
+#database name to partition mapping
+db1=1
```
osheroff: if we're going down this road, I'd suggest that we go ahead with a table mapping too.  I'd suggest:

```
# table `t1` in any database maps to partition 43
*.t1 = 43  
# all tables in `d1` database map to 34
d1.* = 34
# just d2.t2 
d2.t2 = 99
```

PR #991, opened 2018-05-23
revert alpine linux

osheroff:
This broke snappy at first, and then it broke the kinesis producer.  Two
strikes is plenty.


ericwush:
Frankly speaking, snappy isn't broken, because it is documented: https://github.com/zendesk/maxwell/blob/c5ff05a5b5d76fa8686b7c19050c2adb794299eb/docs/docs/producers.md#snappy-compression
Now you might want to delete that too 🙂

osheroff:
ok, removed snappy note

PR #992, opened 2018-05-23
Better filtering

osheroff:
the old filters were cumbersome and confusing.  I've rewritten them so that filters are now a list of rules, executed in order, like:

```
   exclude: *.*, include: db.tbl1, include: db.tbl2
```


osheroff:
this is now ready for review, maybe @ericwush could have a look.  sorry on the length, the old code just needed to be taken out back & shot.

ericwush:
Sorry @osheroff for the delay. The email notification title is always [WIP] so I didn't check further 😂 

Thanks for the _very_ sophisticated filter upgrade. It took quite a while for me to read through the PR and I'm not quite sure if I understand the the use of `exclude` correctly. From the code it looks like `include` has higher priority than `exclude` (because `FilterResult` only has `include` flag), is that correctly? If yes, does `include: foodb.tbl` have the same effect as `exclude: foodb.*, include: foodb.tbl`?

osheroff:
There's no priority in the new rules - they're simply evaluated in the order given, and if a rule applies to a db/table it either includes or excludes.  They also don't *imply* anything.  so the rule:

`include: foodb.tbl` has no effect by itself (because everything is included by default).

src/main/java/com/zendesk/maxwell/filtering/Filter.java
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		if ( includeDatabases != null ) {
+			for ( String s : includeDatabases.split(",") )
+				filterRules.add("include: " + s + ".*");
+
+		}
+
+		if ( excludeDatabases != null ) {
+			for (String s : includeDatabases.split(","))
```
ericwush: Should be `excludeDatabases`?

src/main/java/com/zendesk/maxwell/filtering/Filter.java
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		if ( includeTables != null ) {
+			for ( String s : includeTables.split(",") )
+				filterRules.add("include: *." + s);
+		}
+
+		if ( excludeTables != null ) {
+			for ( String s : excludeTables.split(",") )
+				filterRules.add("exclude: *." + s);
```
ericwush: Same here, if it already has `exclude: *.*`.

src/main/java/com/zendesk/maxwell/filtering/Filter.java
```diff
@@ -0,0 +1,159 @@
...
+
+		if ( includeDatabases != null ) {
+			for ( String s : includeDatabases.split(",") )
+				filterRules.add("include: " + s + ".*");
+
+		}
+
+		if ( excludeDatabases != null ) {
+			for (String s : includeDatabases.split(","))
+				filterRules.add("exclude: " + s + ".*");
```
ericwush: If it already has `exclude: *.*`, do we still want to add these duplicate rules?

src/main/java/com/zendesk/maxwell/filtering/Filter.java
osheroff: very good catch.  will write a test or two around the upgrade path.

src/main/java/com/zendesk/maxwell/filtering/Filter.java
osheroff: yes, because of how the old filtering worked:  you could do like:

```
include_dbs=foodb
exclude_tables=bartbl
```

which semantically meant: `exclude=*.*; include: foodb.*; exclude: *.bartbl`, ie: "only output everything in `foo` except for `bartable`"

osheroff:
@ericwush was noodling on this a bit, it might be nice to unify all the filtering under one syntax, so for column-level filtering you could say:

```
exclude: db.tbl, include: db.tbl.column = "good_value"
```

or

```
exclude: db.tbl.column = "bad_value"
```

thoughts?

ericwush:
@osheroff I just realised the old filter's config supports `exclude_columns` but the code is actually not doing anything with it? Or am I missing anything?

Back to your suggestion, firstly are we going to support both column name and column value filtering? From the examples you gave, seems the column level config is optional?

I'm just curious how the implementation would look like (sorry I turned on the solution mode...). Currently there is no overlapping checks in `shouldOutputEvent` and `shouldOutputRowMap` so they are checked separately, which means table level filtering and column level filtering are independent. From the new config they look like unified. My only concern is if the check is merged and falls onto `RowMap` level then it would not be as efficient as is, i.e. if we have to check every single column in each `RowMap`.

If I turn off the solution mode - from user/configuration's point of view, it's definitely clearer if all the filtering configs can be unified.

osheroff:
re `exclude_columns` -- that actually means "don't output this columns" and is more rightly part of the output config, and should probably be renamed `output_exclude_columns` or `output_suppress_columns`.  It looks like a mistake that I deprecated the config option.  good catch there.

osheroff:
re: 

> I'm just curious how the implementation would look like 

Yeah, I think I'd probably keep two separate filter lists (one for db/table, one for column values) but present it to the end user as a single interface.

osheroff:
> Back to your suggestion, firstly are we going to support both column name and column value filtering? From the examples you gave, seems the column level config is optional?

just column value filtering

osheroff:
added some tests around upgrade-from-v2 and fixed the two issues (exclude_columns, bug in upgrade)

osheroff:
@ericwush I'm still going to go unify the column-level filtering, but would you mind approving or giving any more comments here?  I'll do the column-level filtering as a chained pull-request.

ericwush:
@osheroff ah, I was about to ask if you'd do the column-level filtering in this PR. All good 👍

PR #1001, opened 2018-06-06
fix schema resolver issue for master failover

liulikun:
When master fails over, we write a record in `schemas` table with the recovered position from the new master and `[]` as `deltas`. We'd use the this schema change as the latest schema during schema resolving. 

/cc @zendesk/goanna @osheroff

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -573,16 +573,22 @@ public class MysqlSavedSchema {
 			// Only consider binlog positions before the target position on the current server.
 			// Within those, sort for the latest binlog file, then the latest binlog position.
 			PreparedStatement s = connection.prepareStatement(
-				"SELECT id from `schemas` "
-				+ "WHERE deleted = 0 "
-				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position < ?)) AND server_id = ? "
-				+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1");
+				"SELECT id from `schemas` " +
+				"WHERE deleted = 0 AND " +
+				"last_heartbeat_read <= ? AND " +
+			    "((binlog_file < ?) OR " +
+				"(binlog_file = ? AND binlog_position < ?) OR " +
+				"(binlog_file = ? AND binlog_position = ? AND deltas = '[]')) AND " + // master failover
+				"server_id = ? " +
+				"ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1");
 
 			s.setLong(1, targetPosition.getLastHeartbeatRead());
 			s.setString(2, targetBinlogPosition.getFile());
 			s.setString(3, targetBinlogPosition.getFile());
 			s.setLong(4, targetBinlogPosition.getOffset());
-			s.setLong(5, serverID);
+			s.setString(5, targetBinlogPosition.getFile());
+			s.setLong(6, targetBinlogPosition.getOffset());
+			s.setLong(7, serverID);
 
 			ResultSet rs = s.executeQuery();
 			if (rs.next()) {
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -576,7 +576,7 @@ public class MysqlSavedSchema {
 				"SELECT id from `schemas` " +
 				"WHERE deleted = 0 AND " +
 				"last_heartbeat_read <= ? AND " +
-			    "((binlog_file < ?) OR " +
+				"((binlog_file < ?) OR " +
 				"(binlog_file = ? AND binlog_position < ?) OR " +
 				"(binlog_file = ? AND binlog_position = ? AND deltas = '[]')) AND " + // master failover
 				"server_id = ? " +
--- a/src/test/java/com/zendesk/maxwell/MysqlSavedSchemaTest.java
+++ b/src/test/java/com/zendesk/maxwell/MysqlSavedSchemaTest.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.sql.SQLException;
 import java.sql.Connection;
+import java.util.Collections;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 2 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -573,16 +573,22 @@ private static Long findSchema(Connection connection, Position targetPosition, L
...
 			PreparedStatement s = connection.prepareStatement(
-				"SELECT id from `schemas` "
-				+ "WHERE deleted = 0 "
-				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position < ?)) AND server_id = ? "
-				+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1");
+				"SELECT id from `schemas` " +
+				"WHERE deleted = 0 AND " +
+				"last_heartbeat_read <= ? AND " +
+			    "((binlog_file < ?) OR " +
+				"(binlog_file = ? AND binlog_position < ?) OR " +
```
dongxiaohe: Understood and nice fix, may I know where is the code to write `deltas` to be `[]` when master is failover ?

dongxiaohe: Also, is there any other cases that deltas is written to `[]` but it is actually not a failover ? 
I could be wrong but I feel like there are several schema deltas values have special meanings (little confusing and little hard to maintain):

- Empty -> Beginning
- [{...}] -> Real delta and schema
- [] -> Master failover

liulikun: It's in `MysqlSchemaStore` clone method where we specify an empty list for `deltas`: https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/schema/MysqlSchemaStore.java#L122-L134

We only write `[]` as deltas for master failover case AFAIK.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -573,16 +573,22 @@ private static Long findSchema(Connection connection, Position targetPosition, L
...
-				"SELECT id from `schemas` "
-				+ "WHERE deleted = 0 "
-				+ "AND last_heartbeat_read <= ? AND ((binlog_file < ?) OR (binlog_file = ? and binlog_position < ?)) AND server_id = ? "
-				+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1");
+				"SELECT id from `schemas` " +
+				"WHERE deleted = 0 AND " +
+				"last_heartbeat_read <= ? AND " +
+				"((binlog_file < ?) OR " +
+				"(binlog_file = ? AND binlog_position < ?) OR " +
+				"(binlog_file = ? AND binlog_position = ? AND deltas = '[]')) AND " + // master failover
```
timbertson: Why can't we just make this `binlog_position <=` instead of one condition for `<` and one for `=` which only matches empty deltas? I'm assuming there's a good reason, but I'd like to remember it ;)

osheroff: @timbertson has an interesting idea here, but I think "equal to our position" doesn't quite make sense.  better would be to keep the last-applied delta in-memory and check that we haven't already applied it.

timbertson: How would you check that, do you have to ask mysql about the changes and see if they match the before or after state of the delta you're talking about? That sounds complex :(

osheroff: nah, we suck all the deltas into memory when we boot: https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java#L387.  All we'd *really* need to do is to stash away the position of the last delta we processed.  If we hit a schema change at a position that matches the stashed position (as it would in the dupe-schema case), we can ignore it.

timbertson: Maybe we could clarify the scenarios here:

 1. Failover and then crash. We'll have a `position` equal to the last schema, which was a "no-op" schema inserted when we discovered the corresponding position from our old master. It doesn't matter whether we apply this delta (it's empty), but we must apply either this schema or its base_schema_id, not an old schema for the current server_id.
2. Schema change then crash. We'll store the DDL position in `positions`, but when we restart we'll begin "at" the DDL and will process it again. We want to pretend the most recent delta didn't happen, so that we can reprocess it without crashing. This could be done either by selecting a strictly-less-than position, or by ignoring the final delta in the chain when doing the restore.
3. Any position which is between schema changes. We always want to apply the full schema chain, since the DDL has definitely been applied.

It seems like maybe we're storing `position` inconsistently. For the case of regular data, we use `RowMap.getPosition()`, which is actually `nextPosition`. So the stored position is "the next message we should consume".
Whereas in the case of DDL, we're apparently processing a DDL message, then setting our position to the _beginning_ of that DDL, a.k.a "the last message we consumed". Shouldn't we be saving a position pointing to the message _after_ that DDL. I don't know where this code path is, do you @osheroff ?

If we are storing positions inconsistently, we should just fix that instead and then this problem goes away 🤞 .

timbertson: To answer my own question: I think it is consistent, but the schema we're storing in the case of a failover is not valid. That schema's position is pointing to "the next message you'll process", but the schema it represents is in the past.

The hacky thing would be to take `1` off the position when we store a recovery schema so that it's strictly in the past, but that's pretty gross because it's not a valid position (does it matter though? it's a fake schema anyway...). A proper solution would return the next `position` when doing a recovery, but would create a schema for the "last" position (i.e. the previous event).

dongxiaohe: To be clear, I guess this pull request won't fix the scenario where binlog position is pointing to DDL and master failover happen at the same time. It would still process DDL twice. 

So can we only insert position table if it is DML ?

osheroff:
> Whereas in the case of DDL, we're apparently processing a DDL message, then setting our position to the beginning of that DDL, a.k.a "the last message we consumed". 

guh, you're right.  Both DDL and DML positions are wrong, in different ways:

DML events store `nextPosition` as their only position.  So the set in `maxwell.positions` is correct, but if you turn on `--output_binlog_position`, it's wrong -- maxwell points you directly *after* the event.

DDL events store `position` as their only position, so their output (and schema-position is correct), but their position in `maxwell.positions` is wrong. 

I haven't gone through every last scenario, but thinking through the common ones, I think the actual fix here is:

1.  Set both `position` and `nextPosition` in `RowMap` and `DDLMap`, and use them properly. 
2.  Revert the code back to the `<=` behavior, now that we won't be pointing to the start of events.

liulikun:
Replaced by #1003

PR #1003, opened 2018-06-07
Separate out binlog-start-position from binlog-next-position

osheroff:
turns out for DDL we were setting maxwell.positions to the *start* of the event, and for DML we were outputting a position that was the *end* of the event.

We went through a round of PRs that didn't address the core issue until @timbertson pointed out that we were doing something really, really dumb.  

This reverts #914 and should eliminate the need for #1001, I think.


timbertson:
Yay, thanks! 🎉 

I'd like to confirm something though. When we store a position in the `positions` DB, it's always the _next_ position? i.e. the first event we haven't yet processed.

If so, I think we want to keep the schema finding logic to a strictly `<` rather than `<=`, because if we pick up a schema equal to the current position it's for an event we haven't seen yet.

I think on top of these changes we also need to alter the recovery process to:
 - create a schema with `position`, but
 - set our post-recovery position to `nextPosition`

Here's a commit to show what I mean, I haven't updated tests yet: https://github.com/zendesk/maxwell/commit/34ce8313875936d74eee02da4ed073ab8b0ec747. Does this make sense?

osheroff:
Do all the tests break when using `<` instead of `<=`?  The thing about it is it's quite useful to say `<=` when capturing a schema on an non-moving mysql server.

osheroff:
hmmmm, ok.  

I think I'd like to say:

`(<= position AND base_schema_id is NULL) OR ( < position and base_schema_id IS NOT NULL )`

this will handle the capture case -- when we capture the initial schema, we don't really have a *next* position to start from, just the resting position.

timbertson:
Oh yep, that sounds good - I hadn't considered the capture case.
(I haven't actually ran the tests yet, if they _do_ pass we might want to add some coverage around this area)

osheroff:
ok cherry-picked your commit here and added the `base_schema_id` check.  this might be just the ticket.

liulikun:
Like this approach! Will do some verification today.

liulikun:
👍 It's working great! Since we're cleaning up positions, one thing I'm not too sure is we're publishing `nextPosition` for `position` field for each event. That seems wrong too?

osheroff:
@liulikun the problem of outputting `nextPosition` (in the JSON should be fixed in this PR too.  At least, I think).

osheroff:
nope @liulikun was right, and I was wrong.  I fixed the `position` vs `nextPosition` bits.

PR #1008, opened 2018-06-08
update to kafka 1.0, muck with docker

osheroff:
we add an intermediate maxwell-docker shell script in order to support both KAFKA_HOST:KAFKA_PORT and KAFKA_BROKERS style args.  

I also mucked with the Dockerfile a bit to support faster rebuilds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,16 @@
 FROM maven:3.5-jdk-8
-ENV MAXWELL_VERSION=1.14.7 KAFKA_VERSION=0.11.0.1
-
-COPY . /workspace
+ENV MAXWELL_VERSION=1.14.7 KAFKA_VERSION=1.0.0
 
 RUN apt-get update \
     && apt-get -y upgrade \
-    && apt-get install -y make \
-    && cd /workspace \
+    && apt-get install -y make
+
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,12 +1,16 @@
 FROM maven:3.5-jdk-8
-ENV MAXWELL_VERSION=1.14.7 KAFKA_VERSION=0.11.0.1
-
-COPY . /workspace
+ENV MAXWELL_VERSION=1.14.7 KAFKA_VERSION=1.0.0
```
xuebingli: Is it worth making `KAFKA_VERSION` an `ARG`, so it may be optionally specified in the newly added ` bin/maxwell-docker` script?

xuebingli:
This PR overshadows https://github.com/zendesk/maxwell/pull/878.

Dockerfile
osheroff: so this version just specifies the default kafka version that the maven package gets built with.  The runtime version is specified elsewhere (with the command line flag `--kafka_version`)

Now, it'd be quite nice to have a whole host of common env-variables for the docker image, but I think it's outside the scope of this PR

PR #1010, opened 2018-06-11
Hoist column-level filters into `--filter` system

osheroff:
The only thing tricky here is that for efficiency's sake we do a two-pass filter:  first we check if the table for an event is included/excluded (based on database/table).  If we find that the table is excluded, we must make sure there's not a rule later that could bring a *row* out of the table in.

@ericwush

### DIFF
--- a/src/main/java/com/zendesk/maxwell/filtering/FilterColumnPattern.java
+++ b/src/main/java/com/zendesk/maxwell/filtering/FilterColumnPattern.java
@@ -1,9 +1,15 @@
 package com.zendesk.maxwell.filtering;
 
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 public class FilterColumnPattern extends FilterPattern {
+	private static final Logger LOGGER = LoggerFactory.getLogger(FilterColumnPattern.class);
 	private final String columnName;
 	private final Pattern columnPattern;
 	private final boolean columnPatternIsNull;
@@ -40,6 +46,13 @@ public class FilterColumnPattern extends FilterPattern {
 				}
 			}
 		}
+	}
+
+	@Override
+	public boolean couldIncludeColumn(String database, String table, Set<String> columns) {
+		return type == FilterPatternType.INCLUDE
+			&& appliesTo(database, table)
+			&& columns.contains(columnName);
 	}
 
 	@Override
--- a/src/main/java/com/zendesk/maxwell/filtering/FilterColumnPattern.java
+++ b/src/main/java/com/zendesk/maxwell/filtering/FilterColumnPattern.java
@@ -1,0 +1,50 @@
+package com.zendesk.maxwell.filtering;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public class FilterColumnPattern extends FilterPattern {
+	private final String columnName;
+	private final Pattern columnPattern;
+	private final boolean columnPatternIsNull;
+
+	public FilterColumnPattern(
+		FilterPatternType type,
+		Pattern dbPattern,
+		Pattern tablePattern,
+		String columnName,
+		Pattern columnPattern
+	) {
+		super(type, dbPattern, tablePattern);
+		this.columnName = columnName;
+		this.columnPattern = columnPattern;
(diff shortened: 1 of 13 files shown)


src/main/java/com/zendesk/maxwell/filtering/FilterColumnPattern.java
```diff
@@ -0,0 +1,71 @@
...
+		Pattern columnPattern
+	) {
+		super(type, dbPattern, tablePattern);
+		this.columnName = columnName;
+		this.columnPattern = columnPattern;
+		this.columnPatternIsNull = "^null$".equals(columnPattern.toString().toLowerCase());
+	}
+
+	@Override
+	public void match(String database, String table, FilterResult match) { }
```
ericwush: This line and [this](https://github.com/zendesk/maxwell/pull/1010/files#diff-0a3ff09dd0063fec1059f7538d686c75R28) make me think whether `FilterColumnPattern` _is_a_ `FilterPattern`. Strictly speak, whether it _is_a_ `FilterDbTablePattern`. But I don't have a better way to address it. Maybe it's because what you said that we do a two-pass filter and we are checking db and table twice (in both filter patterns).

ericwush: I'm thinking out loud so bear with me if it doesn't make sense.

Is it possible to consolidate table and column patterns so eventually the table pattern _has_a_ column pattern (can only be done after all the rules have been parsed). Because we already have `couldIncludeFromColumnFilters` method so I assume the table pattern could be aware if it has an associated column pattern. 

It might break the encapsulation of `shouldOutputEvent` b/c it will return the table pattern(s) that are evaluated as true (by extending `FilterResult.patterns`), instead of only the boolean (`FilterResult.include`) at the moment. Then `shouldOutputRowMap` will use the associated column patterns (if there is) to filter. It will make the filtering kinda stateful compared to the current completely stateless impl but it won't check db/table the second time on each `RowMap`. 

Maybe it's somehow similar to your 1st alternative? Would it work at all?

osheroff: I don't think a table pattern can strictly has-a column pattern, since rules could look like:

```
   exclude: db.*
   include: db.included_table
   exclude: db.included_table.col = reject_value_1
   exclude: db.included_table.col = reject_value_2
```

unless I'm not quite getting your idea.

ericwush: Right, I wasn't clear enough. I meant a table column _has_a_ optional column pattern(s). Anyways, I tried it out and don't think it's practical in terms of associating the optional column pattern with the table patten right after the parsing. If we can, it will make it much easier later on.

I put together some hacky code 
[7655c5f](https://github.com/zendesk/maxwell/commit/7655c5fe922dba3bfa4fbe4b9eb2e9cfeb8e695f) to demonstrate the second part of my idea. To summarise:

- separate the two patterns completely (they might both extend `AbstractPattern` to reuse some code).
- Use `FilterResult` to hold the state, i.e. tablePatterns and columnPatterns that match the db and table in `shouldOutputEvent` so in `shouldOutputRowMap` it doesn't have to check db/table any more.

The downside is it breaks the encapsulation of the filter. The upside is to remove the duplicate db/table check.  

BTW, I wonder why we need those static methods in `Filter`, e.g. `includes`s

osheroff:
I see.  Yeah, not really sure we can avoid the duplicate checks: when you write it this way, you break the ordering implied.  If someone writes something goofy like `include: foo.*, exclude: foo.bar.col=val, include: foo.bar`, it really should work the way they expect.  I'm not saying this is a great usage, but, eh.

So if this was really cleaner code, fine, but eh, without actually measuring the cost here (and I'm guessing running even like 5-10 db/table regexp matches per row is negligible) it feels very premature optimization to me.

Something that might be cleaner than the weird `couldIncludeFromColumn` check is to just say `hasColumnFiltersFor(ColumnList)`, and if that's true we have to run all filters no matter what.

ericwush:
I don't quite get
> when you write it this way, you break the ordering implied

but I'm not happy with my change either anyway. Just wanted to have a feel on how it might look like, sometimes it's hard to just imagine the code in my head.

PR #1011, opened 2018-06-11
Gessnerfl/1009 modular maxwell

gessnerfl:
This is a first draft of how it can be implemented. It uses the spring framework for dependency injection as it allows inversion of control and with this a simple modular design.

the standalone application represents the original maxwell. 
There are spring auto configurations for spring boot. The standalone application is not using spring boot to avoid unnecessary dependencies.


osheroff:
oh, my...

osheroff:
this PR, is... ahem.  It's way too big.  This would break mergability for any other outstanding PR, and it's got so many changes I can't really make heads or tails of what's important and what's not.

osheroff:
more concretely:

- no, not everything needs to be an interface.  I'll take a large pass on that particular java trope.  If something really lends itself to being swapped with a similar component that shares nothing, so be it, but the thought that someone is gonna wholesale re-implement `MaxwellContext` or `MaxwellConfig`?  nah. 

- you've just sorta started going in and tearing up stuff as you see fit.  Adding those constants in `MaxwellConfig`?  Might take a PR around that, if it weren't intermingled with a crap-ton of other changes.  But any large-scale stuff like this requires waaay more discipline in small bits of sensible changes.

gessnerfl:
@osheroff I did not expected that you just say hey cool lets merge it. Basically the idea was to make a propose and check if this in general goes into a direction which makes sense also for you.

- I agree that not everything needs to be an interface and I think that I could easily get rid of them now. In general I think the whole API project could be removed but it helped me while moving the parts of the code into the different modules and thereby understand the code much better.
Additionally for the configuration the idea was to limit who can mutate the state of this entities but this can be easily achieved by changing the setters to package local visibility when needed.
- The constants I would remove again. I did too much work there. I spent some time to see if the idea works in general but sure if you think that it makes sense in general we can discuss how we can achieve the same without too many changes at once (e.g. fist split maxwell core and the standalone application. Afterward each producer and each metric reporter individually).

Just let me know if you are interested in general to migrate maxwell into such a direction.

gessnerfl:
Basically the interfaces helped me to untangle the dependencies between the different modules without the need of a big bang change but doing the changes step by step. With a better understanding of the application I think that the project structure can be simplified:

```
 --------------------                      --------------
| Maxwell Standalone |------------------->| Maxwell Core | 
 --------------------                      --------------
          |                                       ^           ----------------
          |      -----------------------          |----------| Producer Kafka | 
          `---->| Metrics Reporter Core |         |           ----------------
                 -----------------------          |
                                ^                 |           ------------------
                                |                 |----------| Producer Kinesis | 
 --------------------------     |                 |           ------------------
| Metrics Reporter DataDog |----|                 |
 --------------------------     |                 |           -----------------
                                |                 |----------| Producer Pubsub | 
     -----------------------    |                 |           -----------------
    | Metircs Reporter HTTP |---´                 |
     -----------------------                      |           -------------------
                                                  |----------| Producer RabbitMQ | 
                                                  |           -------------------
                                                  |
                                                  |           ----------------
                                                  |----------| Producer Redis | 
                                                  |           ----------------
                                                  |
                                                  |           --------------
                                                  `----------| Producer SQS | 
                                                              --------------
```
With this we can avoid all the interfaces. And can simplify the project structure significantly.
It might be needed to add some API projects if clear separation should be achieved. 
Basically the idea would be the following:

* Standalone is the main application and responsible for parsing command line arguments and configuration
* Core is responsible for the MySQL processing and defines the interfaces for the producers
* Metrics Reporter projects implement the different metric reporter. They should not have a dependency to the core project and only depend on the metrics and health check registry
* Core should not have a dependency to the metric reporter but only provide data through the metric/health check registry

However it will still need a significant change on the project. I think it can be achieved iteratively. For me the question is whether this is something worth to invest or if you think that it is hard to achieve e.g. because of time limitations, breaking changes, ....

osheroff:
Sorry if I came off brusque there.  I have been known to.  

I guess I'd ask a broad question:  what are we trying to accomplish?  What do we win by breaking everything up?  It's a pretty ascii graph, but it's really also more-or-less how Maxwell works, except we cross package boundaries instead of module boundaries. 

You stated the original goal of reducing the package size of an embedded maxwell deploy, ok, I guess that's something, but what more?  Do we improve readbility?  Do we enable features that would be previously tricky to implement?  Do we make life easier for those running embedded maxwell?

gessnerfl:
From my point of view I would limit the scope to breaking up the application into modules so that it can be integrated into applications easily. Maybe as a background what I would target for:

Basically I would like to integrate 'Maxwell Core' and 'Maxwell Producer RabbitMQ' into a spring boot app and extend this with rest interface so that I can control the service through a remote api. Maxwell would be used in this case to synchronize MySQL with Elasticsearch. In case full re-synchronization is needed (e.g. because of data migration) I would be able to recreate this in a blue-/green fashion but therefore I need control over Maxwell instance(s) and therefor I would like to integrate it into a service with a remote API.

As I basically only need the RabbitMQ producer it would help us to get rid of all the dependencies. 
I know it is a very specific use case and maybe you might think of the Maxwell is the wrong decision in this case. I just thought that because a similar discussion was already started with #634 it's worth to trigger a discussion about such a major change here.

osheroff:
your use case sounds perfectly sane.  

Regarding splitting out producers:  I could see it.  Maybe someone wants to maintain a separate public producer, and it'd be nice to have a method for each producer to be able register its own config options and keywords, and to have those kept closer to the actual producer code.

Regarding splitting out api, metrics:  maaaaybe. 

Regarding using a lot of dependency-injection stuff:  use sparingly.  

There's patch-points that could make a lot of sense to allow injection for (filters, message routing, data munging points), but just busting everything apart and allowing people to cobble together pieces doesn't equate to re-usable code.  `MaxwellContext`, for example, is a horrible, nasty dumping ground of code I couldn't find a better place for.  Some of it belongs there, but most of it is just there because it seemed like a good idea at a time.  You can shove it into a interface, but that really just makes everything worse: now you've got bad interfaces, and codified in a way that you can't change without breaking upstream consumers.

The cost of dependency-injection and interfaces is that at any level of code you're always looking at some AbstractFoo or IFoo object and left to scratch your head about where the *real* code lives.  I've seen far too many libraries and projects out there where everything is an Interface, and there's exactly one implementation of that Interface.  And then you go do a grep or other search to ferret out the XFooBarImpl class that does the real work, and then that class references some other abstract interface and then you throw up your hands.  If asked, someone will invariably mutter something about "implementation detail hiding" or some stuff.   Anyway.  No.  Interfaces are fine and nice for things people *will* want to rip out and roll their own of.

PR #1019, opened 2018-06-14
fix bootstrapping for JSON columns

### LINKED ISSUE #1017: JSON Data Type Parser Throwing Exception
The Maxwell Kafka Producer is breaking upon read of what appears to be larger size JSON data types from MySQL. It is throwing the following error:

`java.lang.ClassCastException: java.lang.String cannot be cast to [B at com.zendesk.maxwell.schema.columndef.JsonColumnDef.asJSON(JsonColumnDef.java:18) at com.zendesk.maxwell.bootstrap.SynchronousBootstrapper.setRowValues(SynchronousBootstrapper.java:241) at com.zendesk.maxwell.bootstrap.SynchronousBootstrapper.startBootstrap(SynchronousBootstrapper.java:66) at com.zendesk.maxwell.bootstrap.AsynchronousBootstrapper$1.run(AsynchronousBootstrapper.java:74) at java.lang.Thread.run(Thread.java:748)`

This seems to occur exclusively on json data type columns from MySQL, but I haven't been able to see any obvious gremlin characters in the JSON. Is it possible this is due to encoding issues? Looks like the JSON encoder is also reliant upon this: https://github.com/shyiko/mysql-binlog-connector-java/blob/master/src/main/java/com/github/shyiko/mysql/binlog/event/deserialization/json/JsonBinary.java#L181 could it be an issue with the binlog deserialization? I can't paste the exact JSON object, but the only characters that look even remotely suspect would just be ()[]/=: none of which are very extraordinary for JSON objects. Is there any limitation on the size of the JSON object?

osheroff:
bootstrap values come back as strings from JDBC vs byte[] from binlog-connector.

fixes #1017

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/columndef/JsonColumnDef.java
+++ b/src/main/java/com/zendesk/maxwell/schema/columndef/JsonColumnDef.java
@@ -14,13 +14,19 @@ public class JsonColumnDef extends ColumnDef {
 
 	@Override
 	public Object asJSON(Object value) {
-		try {
-			byte[] bytes = (byte[]) value;
-			String jsonString = bytes.length > 0 ? JsonBinary.parseAsString(bytes) : "null";
-			return new RawJSONString(jsonString);
-		} catch (IOException e) {
-			throw new RuntimeException(e);
+		String jsonString;
+
+		if ( value instanceof String ) {
+			jsonString = (String) value;
+		} else {
+			try {
+				byte[] bytes = (byte[]) value;
+				jsonString = bytes.length > 0 ? JsonBinary.parseAsString(bytes) : "null";
+			} catch (IOException e) {
+				throw new RuntimeException(e);
+			}
 		}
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/schema/columndef/JsonColumnDef.java
```diff
@@ -14,13 +14,19 @@ public JsonColumnDef(String name, String type, int pos) {
...
-		try {
-			byte[] bytes = (byte[]) value;
-			String jsonString = bytes.length > 0 ? JsonBinary.parseAsString(bytes) : "null";
-			return new RawJSONString(jsonString);
-		} catch (IOException e) {
-			throw new RuntimeException(e);
+		String jsonString;
+
+		if ( value instanceof String ) {
+			jsonString = (String) value;
```
yosefrat: Oh, I see thanks for the quick action!

osheroff:
ping

PR #1031, opened 2018-06-26
--  add producer_partition_by mode : priority_column

Gechmind:
we often face this case, we want data belong to a root key send  to the same partition.there may be many subsidiary table. so we may first try the foreign  key and then other keys, because uuid generator is very common,so we can handle the uniform distribution case。


osheroff:
I see.  This makes decent sense.  Can you name this `first_column` instead and document it a bit?

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java
```diff
@@ -20,6 +20,8 @@ public MaxwellKafkaPartitioner(String hashFunction, String partitionKey, String
 	}
 
 	public int kafkaPartition(RowMap r, int numPartitions) {
-		return Math.abs(hashFunc.hashCode(this.getHashString(r)) % numPartitions);
+		int partition =  Math.abs(hashFunc.hashCode(this.getHashString(r)) % numPartitions);
+		System.out.println("partition:" + partition);
+		return partition;
```
osheroff: obviously get rid of this

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -226,6 +213,26 @@ public String buildPartitionKey(List<String> partitionColumns) {
...
+		for(String pc:partitionColums){
+			if (data.containsKey(pc)){
+				pcValue = data.get(pc);
+			}
+
+			if (pcValue != null){
+				break;
+			}else{
+				continue;
+			}
```
osheroff: no need for the `else` here

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -226,6 +213,26 @@ public String buildPartitionKey(List<String> partitionColumns) {
...
+		}
+		if(pcValue != null){
+			return pcValue.toString();
+		}else{
+			return "";
+		}
+	}
+
```
osheroff: please reformat your `if` and `else` statements with space.

src/main/java/com/zendesk/maxwell/producer/partitioners/MaxwellKafkaPartitioner.java
```diff
@@ -20,6 +20,6 @@ public MaxwellKafkaPartitioner(String hashFunction, String partitionKey, String
 	}
 
 	public int kafkaPartition(RowMap r, int numPartitions) {
-		return Math.abs(hashFunc.hashCode(this.getHashString(r)) % numPartitions);
+		return  Math.abs(hashFunc.hashCode(this.getHashString(r)) % numPartitions);
```
osheroff: please undo this whitespace change

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -197,6 +197,26 @@ public void testBuildPartitionKey() {
...
+//		rowMap.putData("sharding_value", "foo");
+//		rowMap.putData("id", "buzz");
+		rowMap.putData("last_name", "bar");
+		rowMap.putData("salary", "4000");
+		rowMap.putData("department", "science");
+
+		List<String> partitionColumns = Arrays.asList("subject_id",  "sharding_value","id");
+		String partitionKey = rowMap.buildPartionPriorityKey(partitionColumns);
+
+		Assert.assertEquals("9001", partitionKey);
```
osheroff: how is this test actually passing when the `subject_id` putData is commented out?

osheroff: i see you merged master.  I fixed a problem with the test suite scripts that was passing failing tests.

PR #1057, opened 2018-08-03
Add schema_id as an optional field in output

macboy012:
**Motivation:**
I want to have an id that can be used by consumers to know whether or not they're ok to consume a message. If they know they're ok to consume all schema_id 5 and they see schema_id 6, they can stop consuming and check against another service what schema is ok to consume, not continuing until they get the ok to consume schema_id 6.

**Uncertainties:**

- There's some differing conventions  in the code regarding capitilization of id, SchemaStore users `ID`, other places like RowMap, it's `Id`. So, BinlogConnectorReplicator has the boundary `Long schemaId = this.schemaStore.getSchemaID();`
- Wasn't really sure what schema_id to return from RecoverySchemaStore, I put 0, I didn't ponder it too long.
- DDL output currently has the new schema_id, should be the very first instance of it, it could instead have the previous schema_id, the very last instance of it. I thought this way made more sense.

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -80,7 +80,7 @@ password=maxwell
 #output_gtid_position=true
 
 # records include fields with null values (default true).  If this is false,
-# fiels where the value is null will be omitted entirely from output.
+# fields where the value is null will be omitted entirely from output.
 #output_nulls=true
 
 # records include server_id (default false)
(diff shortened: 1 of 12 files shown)


osheroff:
not sure I fully understand your use-case, but I'm generally ok with the code here.  Throw in a test around it and we can be good to go.

PR #1081, opened 2018-08-28
Add message.publish.age metric

timbertson:
This tracks the age of the event _after_ it's been published (according to DB timestamp), i.e. end-to-end lag.

Since MySQL timestamp is only accurate to seconds, this is not accurate for subsecond lag. Negative values are clipped to 0.

/cc @zendesk/goanna

### DIFF
--- a/docs/docs/monitoring.md
+++ b/docs/docs/monitoring.md
@@ -23,6 +23,7 @@ metric                         | description
 `inflightmessages.count`       | the number of messages that are currently in-flight (awaiting acknowledgement from the destination, or ahead of messages which are)
 **Timers**
 `message.publish.time`         | the time it took to send a given record to Kafka, in milliseconds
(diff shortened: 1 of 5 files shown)


osheroff:
do we have a place where we document the metrics we expose?  If so, can you add it to there, if not, can you hack one out real quick?  code looks fine.

osheroff:
note that truncating to zero is not the right method to get a good *average* latency -- it's always going to be high by 500ms.   If you're only looking for an average latency the proper way to do it is to do `publishTime - dbTime - 0.5`.  But you do end up with some occasionally goofy negative latencies.

timbertson:
True. I've taken off 500ms now, though I'm still truncating to 0 to filter out whacky negatives. For the most part we don't care about the accuracy of subsecond lag measurements, as the lag grows then it gets both more accurate and more worth paying attention to ;)

We do have docs, updated now 👍 .

PR #1082, opened 2018-08-29
Kinesis: fail records > 1MB instead of crashing

reist:
The error caused Maxwell to crash until the end of the retention period,
at which point the binlog file with the change disappeared and Maxwell
failed with "Could not find first log file name in binary log index file"

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
@@ -120,9 +120,9 @@ public class MaxwellKinesisProducer extends AbstractAsyncProducer {
 	public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) throws Exception {
 		String key = this.partitioner.getKinesisKey(r);
 		String value = r.toJSON(outputConfig);
+		int vsize = value.length();
 
 		ByteBuffer encodedValue = ByteBuffer.wrap(value.getBytes("UTF-8"));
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellKinesisProducer.java
```diff
@@ -132,6 +132,12 @@ public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) thro
 		FutureCallback<UserRecordResult> callback = new KinesisCallback(cc, r.getNextPosition(), key, value,
 				this.succeededMessageCount, this.failedMessageCount, this.succeededMessageMeter, this.failedMessageMeter, this.context);
 
-		Futures.addCallback(future, callback);
+		try {
+			ListenableFuture<UserRecordResult> future = kinesisProducer.addUserRecord(kinesisStream, key, encodedValue);
+			Futures.addCallback(future, callback);
+		} catch(IllegalArgumentException t) {
+			callback.onFailure(t);
+			logger.error("Database:" + r.getDatabase() + ", Table:" + r.getTable() + ", PK:" + r.pkAsConcatString() + ", Size:" + Integer.toString(vsize));
```
osheroff: should we log the exception too?  Or is that going to be logged in `callback.onFailure`?

reist: `callback.onFailure` will log it. I was wondering whether to also log the beginning of the query (first 100 chars? 1000?) in case it wasn't an insert or update, but I'm not sure about that.

Issue #1091, opened 2018-09-07
bug: ddl containing "\n drop index" will be ignored


chen56:
```java
package com.zendesk.maxwell.schema.ddl;

public class DDLParserTest {
	@Test
	public void bug_ddl_contains_newline_drop_index_will_be_ignored() {
		// Statements containing "\n drop index" will be ignored.
		assertThat(parse("ALTER TABLE t\n" +
				"  DROP `tatle`,\n" +
				"  ADD `title` varchar(50),  \n  DROP INDEX index_foo\n"), is(nullValue()));
		assertThat(parse("ALTER TABLE t\n" +
				"  DROP `tatle`,\n" +
				"  ADD `title` varchar(50),   DROP INDEX index_foo\n"), is(not(nullValue())));

		//i find the reason:  Pattern.MULTILINE
		Pattern SQL_BLACKLIST_NOW=Pattern.compile("^\\s*(ALTER|CREATE|DROP)\\s+((ONLINE|OFFLINE|UNIQUE|FULLTEXT|SPATIAL)\\s+)*(INDEX)", Pattern.CASE_INSENSITIVE | Pattern.MULTILINE);
		assertThat( SQL_BLACKLIST_NOW.matcher("alter table t \n drop index `foo`").find(),is(true));
		assertThat( SQL_BLACKLIST_NOW.matcher("alter table t drop index `foo`").find(),is(false));
	}

}
```

osheroff:
this will be fixed when #1148 merges

PR #1092, opened 2018-09-07
fix bug: ddl containing "\n drop index" will be ignored #1091

chen56:
fix bug: ddl containing "\n drop index" will be ignored #1091 
add more info on json error


chen56:
@osheroff ok, waiting you solve this problem.

osheroff:
fixed more precisely in #1148

Issue #974, opened 2018-05-14
Monitoring Maxwell using Prometheus


dula24:
Hi, 
Is there a way to expose metrics in prometheus supported format?

ericwush:
Not familiar with prometheus but seems you can use
https://mvnrepository.com/artifact/io.prometheus/simpleclient_dropwizard, since `dropwizard` is used for metrics in Maxwell. Feel free to write a PR.

dmarmugi:
should this be closed? it looks like it was fixed by #1096

PR #1096, opened 2018-09-08
Add support for Prometheus metrics

vasekboch:
I was trying to deploy Maxwell to production, but I noticed that its missing metrics in Prometheus (Prometheus.io) format. I made a implementation based out suggestions on https://github.com/zendesk/maxwell/issues/974

Could you guys check it. I was struggling with the configuration options, so I could use an second opinion.
1) Use following solution. Add another metric_type ```prometheus``` and enable the endpoint when needed.
2) Enable Prometheus automatically when ```http``` is provided.
3) Starting different http server on separate port (too much resources in my opinion)

I'm looking forward to your ideas. Basically its my first PR :D

### DIFF
--- a/docs/docs/monitoring.md
+++ b/docs/docs/monitoring.md
@@ -33,6 +33,7 @@ When the HTTP server is enabled the following endpoints are exposed:
 | endpoint       | description                                                                    |
 |:---------------|:-------------------------------------------------------------------------------|
 | `/metrics`     | return all metrics as JSON                                                     |
+| `/prometheus`  | return all metrics as Prometheus format                                        |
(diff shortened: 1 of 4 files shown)


osheroff:
yeah I peeked at it, that test was flaky and relied on timers.  I patched over it in the current master.

osheroff:
re: the code, I think it's ok to go and and enable the new `/prometheus` endpoint if http is enabled.  It doesn't really cost much of anything to enable it, and we save having to specify another config flag.

vasekboch:
Cool, thx for the review. I've redone the code, according your suggestion. I think that you are right. Better to have less configuration.

osheroff:
you want this merged or are ya still working on it?

vasekboch:
Ah, sorry. Yes, please. Merge it. I think that is ready.

Issue #1075, opened 2018-08-20
Non-english column name support?


d8888:
When I used Maxwell Daemon to stream data from table containing Chinese column name (such as "你好"), I noticed that the key corresponding to column name became question marks, that is:

I expect message should be:

{..., "data": {"你好":"123"}}

however in reality the message will looks like:

{..., "data": {"????":"123"}}

I also noticed that the "columns" table in "maxwell" database used latin_swedish_ci encoding for "name" column (which corresponds to column name), the "name" column also incorrectly stored column name as question marks (e.g stored "????" instead of "你好"). I changed encoding for "name" column into UTF-8 and repeated bootstraping process however the problem persisted.

Does Maxwell Daemon support non-English column name? Or is there anywhere to circumvent this problem? Thanks in advance!

osheroff:
Interesting bug, I will have to look.  No clue why I would have picked latin-1 for column names, seems overly optimistic :)

d8888:
I have proposed pull request for this issue. In my own manual testing, Chinese letters now display correctly in JSON message in stdout.
https://github.com/zendesk/maxwell/pull/1099

osheroff:
fixed in v1.18.0

PR #1099, opened 2018-09-10
Non-english characters support for db, table, column names

d8888:
An attempted fix for this issue: https://github.com/zendesk/maxwell/issues/1075
Problem: When using Maxwell Daemon to stream data from table containing Chinese column name (such as "你好"), key in JSON message corresponding to column name became question marks, that is:
expected outcome: {..., "data": {"你好":"123"}}
actual outcome: {..., "data": {"????":"123"}} 

This PR changed Maxwell to use UTF8 in JDBC connection. Corresponding test is also changed. 
Charset of db, table, column names are changed to UTF-8 in .sql files

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellMysqlConfig.java
@@ -107,6 +107,9 @@ public class MaxwellMysqlConfig {
 			uriBuilder.addParameter(jdbcOption.getKey(), jdbcOption.getValue());
 		}
 
+		// added by d8888 2018/09/10, force JDBC to use UTF-8 to support using non-english db, table & column names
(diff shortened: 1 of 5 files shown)


osheroff:
ok, can you add some tests around this?  `DDLIntegrationTest` is probably a good place to write one...

d8888:
I've added test into DDLIntegrationTest however the continuous integration failed
I got multiple "java.lang.NullPointerException" from multiple tests which I haven't modified (according to commit logs)
Retriggering the build process didn't help.
No error occured in "make test" of my own machine.

Is there any clues to solve these problems? thanks

osheroff:
sorry yeah I broke the builds with the changes to `onetimeserver`.  just pushed a fix

d8888:
Thanks for your kind help, the build succeeded.

PR #1103, opened 2018-09-13
Bump mysql-connector-java version for mysql 8 support

macboy012:
Current version 5.1.39 doesn't work with mysql 8.
If it worked, I'd bump it to 8.0.12 (latest) but I got a compilation error when I tried it.


macboy012:
😬worked on my local, not sure what's different here

osheroff:
looks like I broke the build with some unrelated change to the test-suite tooling.  restarted.

Issue #1111, opened 2018-09-20
[Feature Request] Offical Nats Connector


jainnidhi703:
Creating official [Nats](https://nats.io/documentation/streaming/nats-streaming-intro/) connector from Maxwell. Nats is an upcoming streaming server and is growing fast. It would be good to have an official Nats maxwell connector.

osheroff:
released in 1.31.0

PR #1115, opened 2018-09-25
Nats Streaming Producer

jainnidhi703:
**What**
Nats Streaming Async Producer (#1111)

**How**
Using the [Nats Java Connector](https://github.com/nats-io/java-nats-streaming) publishing message to Nats

**Usage**
Set `producer=nats` and other config related to nats: 
```
nats_url=<nats_url>
nats_cluster_id=<cluster_id>
nats_client_id=<client_id>
nats_subject=<subject>
```

Issue

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -134,6 +134,7 @@ public class MaxwellConfig extends AbstractConfig {
 	public Long natsPubAckTimeout;
 	public Long natsReconnectWait;
 	public int natsMaxReconnects;
+	public int natsMaxAttempts;
 
 
 	public MaxwellConfig() { // argv is only null in tests
@@ -424,11 +425,12 @@ public class MaxwellConfig extends AbstractConfig {
 
 		this.natsUrl = fetchOption("nats_url", options, properties, "nats://localhost:4222");
 		this.natsClusterId = fetchOption("nats_cluster_id", options, properties, "test-cluster");
-		this.natsClientId = fetchOption("nats_client_id", options, properties, "test-client");
-		this.natsSubject = fetchOption("nats_subject", options, properties, "test-subject");
+		this.natsClientId = fetchOption("nats_client_id", options, properties, "maxwell");
+		this.natsSubject = fetchOption("nats_subject", options, properties, "maxwell");
 		this.natsPubAckTimeout = fetchLongOption("nats_pub_ack_timeout", options, properties, null);
-		this.natsMaxReconnects = Integer.parseInt(fetchOption("nats_max_reconnects", options, properties, "0"));
+		this.natsMaxReconnects = fetchLongOption("nats_max_reconnects", options, properties, 0L).intValue();
 		this.natsReconnectWait = fetchLongOption("nats_reconnect_wait", options, properties, null);
+		this.natsMaxAttempts = fetchLongOption("nats_max_attempts", options, properties,5L).intValue();
 
 
 		String kafkaBootstrapServers = fetchOption("kafka.bootstrap.servers", options, properties, null);
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -132,6 +132,8 @@ public class MaxwellConfig extends AbstractConfig {
 	public String natsClientId;
 	public String natsSubject;
 	public Long natsPubAckTimeout;
+	public Long natsReconnectWait;
+	public int natsMaxReconnects;
 
 
 	public MaxwellConfig() { // argv is only null in tests
@@ -425,6 +427,9 @@ public class MaxwellConfig extends AbstractConfig {
 		this.natsClientId = fetchOption("nats_client_id", options, properties, "test-client");
 		this.natsSubject = fetchOption("nats_subject", options, properties, "test-subject");
 		this.natsPubAckTimeout = fetchLongOption("nats_pub_ack_timeout", options, properties, null);
+		this.natsMaxReconnects = Integer.parseInt(fetchOption("nats_max_reconnects", options, properties, null));
+		this.natsReconnectWait = fetchLongOption("nats_reconnect_wait", options, properties, null);
+
 
 		String kafkaBootstrapServers = fetchOption("kafka.bootstrap.servers", options, properties, null);
 		if ( kafkaBootstrapServers != null )
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -127,6 +127,12 @@ public class MaxwellConfig extends AbstractConfig {
 	public String javascriptFile;
 	public Scripting scripting;
 
+	public String natsUrl;
+	public String natsClusterId;
+	public String natsClientId;
+	public String natsSubject;
+
+
 	public MaxwellConfig() { // argv is only null in tests
 		this.customProducerProperties = new Properties();
 		this.kafkaProperties = new Properties();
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,88 @@
...
+
+	public MaxwellNatsProducer(MaxwellContext context, String natsSubject) {
+		super(context);
+		this.natsSubject = natsSubject;
+		StreamingConnectionFactory cf = new StreamingConnectionFactory(context.getConfig().natsClusterId, context.getConfig().natsClientId);
+		cf.setNatsUrl(context.getConfig().natsUrl);
+		try {
+			this.connection = cf.createConnection();
+		} catch (IOException | InterruptedException e) {
+			logger.error("Exception during Nats Connection", e);
```
osheroff: we'd probably want to throw a `RuntimeException` here.  There's no sense in carrying on with a broken connection.

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+
+	@Override
+	public void sendAsync(RowMap r, CallbackCompleter cc) throws Exception {
+		String msg = r.toJSON();
```
osheroff: please use the call of `toJSON` that takes `outputConfig`

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,88 @@
...
+		NatsAckHandler handler = new NatsAckHandler(cc, r.getNextPosition(), msg, context);
```
osheroff: you probably mean `getPosition` not `getNextPosition`

jainnidhi703: When I check the usage of getNextPosition in Producers such as MaxwellKinesisProducer, MaxwellKafkaProducer (classes extending AbstractAsyncProducer) all of them use getNextPosition. Redis producer is also using the same.

osheroff: yes, they use `getNextPosition` to advance the binlog, but if you're just logging the position of the event, that position is `getPosition`....

jainnidhi703:
@osheroff I didn't understand the use case of Stoppable task, although I implemented it-so when Nats disconnects taskState.stopped(), but when it reconnects the I wasn't able to change taskState to running again. Changed it to a boolean flag isNatsConnected instead.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -405,6 +422,15 @@ private void setup(OptionSet options, Properties properties) {
 		this.redisListKey		= fetchOption("redis_list_key", options, properties, "maxwell");
 		this.redisType			= fetchOption("redis_type", options, properties, "pubsub");
 
+		this.natsUrl = fetchOption("nats_url", options, properties, "nats://localhost:4222");
+		this.natsClusterId = fetchOption("nats_cluster_id", options, properties, "test-cluster");
+		this.natsClientId = fetchOption("nats_client_id", options, properties, "test-client");
+		this.natsSubject = fetchOption("nats_subject", options, properties, "test-subject");
```
osheroff: sorry, just got back to this PR.  Putting a default here of `test-subject` seems a touch optimistic, I think probably better to have it required?  I guess a default of `maxwell` wouldn't be terrible either.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -405,6 +422,15 @@ private void setup(OptionSet options, Properties properties) {
...
+		this.natsPubAckTimeout = fetchLongOption("nats_pub_ack_timeout", options, properties, null);
+		this.natsMaxReconnects = Integer.parseInt(fetchOption("nats_max_reconnects", options, properties, "0"));
```
osheroff: could just use `(Integer) fetchLongOption`...

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,229 @@
...
+		while ( true ) {
+			try {
+				//Irrespective of Nats is working or not, heartbeat should be pushed
+				if (queue.peek() instanceof HeartbeatRowMap) {
+					RowMap row = queue.take();
+					this.push(row);
+				} else if (connectionListener.isNatsRunning()){
+					RowMap row = queue.take();
+					this.push(row);
+				}
```
osheroff: this will be a very tight loop if nats is in a not-yet-connected state, we probably want to sleep for a few milliseconds if we're waiting for the connection to come online, no?

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,229 @@
...
+
+	@Override
+	public void connectionEvent(Connection connection, Events events) {
+		switch (events) {
+			case DISCONNECTED:
+				isNatsRunning = false;
+				LOGGER.error("Nats connection disconnected");
+				if (!context.getConfig().ignoreProducerError) {
+					this.context.terminate(new RuntimeException("Nats Disconnected"));
+					return;
```
osheroff: should this be something that crashes maxwell?  Or can we wait for a reconnect?

jainnidhi703: if ignoreProducerError is false, then we should terminate right?

osheroff: I'm just not sure what the `DISCONNECTED` message signifies: does it mean the client disconnected and will try to re-open the connection, or does it mean that it's dead and has given up?

jainnidhi703: DISCONNECTED callback comes when the nats connection was down.

jainnidhi703:
@osheroff Please review the changes.

Thanks,
Nidhi

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	@Override
+	public void run() {
+		//Waiting for Nats connection
+		int attempts = 0;
+		try {
+			Thread.sleep(1000L);
+			while ( true ) {
+				// Irrespective of Nats is working or not, heartbeat should be pushed
```
osheroff: comment in the wrong place now

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
```diff
@@ -0,0 +1,242 @@
...
+					} else {
+						this.context.terminate(new RuntimeException("Nats Disconnected"));
+						return;
+					}
+				}
+
+			}
+		} catch ( Exception e ) {
+			LOGGER.info("Error: ", e);
+		}
```
osheroff: can you look at `MaxwellKafkaProducer` for an example on how to handle errors, as well as how to exit gracefully (implement StoppableTask)

osheroff:
are you running this in production anywhere?  I generally love code to be run in production before merging.

jainnidhi703:
Not running it in production

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
jainnidhi703: While using the StoppableTask, when Nats sent Reconnected event the task wasn't able to restart and push to Nats.

osheroff:
running it in development, then?  anywhere?

jainnidhi703:
Was hoping to get it merged and then deploy maxwell daemon with Nats feature

osheroff:
Unless it's trivial, I really really love to have code run somewhere before merged.  I can mark the producer as beta I guess, but don't you *also* want to know your code works?

src/main/java/com/zendesk/maxwell/producer/MaxwellNatsProducer.java
osheroff: that doesn't make much sense to me.  StoppableTask really is just an interface that lets the rest of maxwell shutdown a producer/other task when it's time to exit.

osheroff:
nats done got implemented elsewhere

PR #1116, opened 2018-09-26
Fixed column misarrangement bug

pankaj-mt:


osheroff:
thanks for the PR, this showed up in mysql 8 and I already had this fix in #1113

Issue #1041, opened 2018-07-03
Questions about bootstrap per client ID


leonjoyce:
I reviewed the latest code, but have some questions:
**com.zendesk.maxwell.bootstrap.SynchronousBootstrapper.resume(AbstractProducer, Replicator)**

@osheroff
Should we add client ID to the function above? So that we only resume bootstrap of the specified maxwell client?

If not, I am afraid that the following code will throw NULL pointer exception.
public boolean isBootstrapRow(RowMap row) {
return row.getDatabase().equals(this.context.getConfig().databaseName) &&
row.getTable().equals("bootstrap") &&
**row.getData("client_id")**.equals(this.context.getConfig().clientID);
}

osheroff:
@leonjoyce sure, nice catch.  Otherwise if one maxwell client was bounced it'd clobber the in-progress bootstrap from another.  can you go ahead and make that PR?

osheroff:
fixed now

PR #1120, opened 2018-10-08
clientID added to resume

leonjoyce:
According to the following discussion, added client ID to bootstrap.resume
https://github.com/zendesk/maxwell/issues/1041

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -524,4 +524,5 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 		return this.schemaStore.getSchema();
 	}
 	
+
 }


src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -150,7 +152,8 @@ public void resume(AbstractProducer producer, Replicator replicator) throws Exce
 			// This update resets all rows of incomplete bootstraps to their original state.
 			// These updates are treated as fresh bootstrap requests and trigger a restart
 			// of the bootstrap process from the beginning.
-			String sql = "update `bootstrap` set started_at = NULL where is_complete = 0 and started_at is not NULL";
+			String clientid = ((BinlogConnectorReplicator)replicator).getClientID();
```
osheroff: better to extract `clientID` from `context`

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
```diff
@@ -150,7 +152,8 @@ public void resume(AbstractProducer producer, Replicator replicator) throws Exce
...
+			String sql = "update `bootstrap` set started_at = NULL where is_complete = 0 and started_at is not NULL and client_id = '" + clientid + "'";
```
osheroff: this needs to use regular sql quoting via `client_id = ?` and `PreparedStatement#setString()`

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
leonjoyce: Agree

src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
leonjoyce: I fix the code according to review comments. But not sure why the build fails. Can You help?

osheroff:
this change got merged in another PR

Issue #1100, opened 2018-09-10
memory issue when bootstrapping


jingchima-appf:
We are having memory issue when do bootstrapping. The stream is kinesis.

There are 226 databases on the slice, and roughly 50 tables in each database. We care about 44 databases and 5 tables in each of the 44 databases. In total there are over 1 million records we need. 

When Maxwell is started, it consume about 1Gb memory. But after we started bootstrapping all the tables in all database, the memory usage constantly increases and reached 13GB. After 5 minutes after bootstrapping it still remains 10 GB and won't drop down. 

We've tried to reduce the buffer size to reduce memory usage. Specifically, we set RecordTTL = 1000 milliseconds and recordMaxBufferedTime  = 100. But we didn't see any decrease in memory usage. 

Any way to reduce the memory usage? Thanks in advance.

osheroff:
I'd be interested to know where the memory usage is coming from; either the records are getting buffered somewhere inside the Kinesis producer, or the bootstrap code is holding onto records.  I was pretty sure we had the bootstrap code tuned to not retain records in memory.  

Can you grab a heap histogram with `jmap -histo:live`?

jingchima-appf:
Thanks for the reply! I am also wondering what is using memory. I'll run `jmap -histo:live` and update the results tomorrow.

BTW, we also tried limiting the max memory JVM can use. When we limit it to 5GB, the memory usage is 4.9 GB, and we didn't get `outOfMemoryException`, and the bootstrap slows down somehow. I'm guessing that has something to do with java's garbage collections.

gautamvij:
Facing same issue. Works fine with real time data at 500mb. 
On bootstrap it spikes to 2.3 gb and won't drop down.

Xmx set to 1024mb.

```
 num     #instances         #bytes  class name
----------------------------------------------
   1:        927319       44511312  com.zendesk.maxwell.producer.KinesisCallback
   2:        927319       37092760  com.google.common.util.concurrent.AbstractFuture$Sync
   3:        932588       29842816  java.util.concurrent.ConcurrentHashMap$Node
   4:        927319       29674208  com.zendesk.maxwell.producer.AbstractAsyncProducer$CallbackCompleter
   5:        929437       22306488  java.lang.Long
   6:        927319       22255656  com.google.common.util.concurrent.ExecutionList
   7:        927319       22255656  com.google.common.util.concurrent.ExecutionList$RunnableExecutorPair
[...]

Issue #816, opened 2017-11-24
Maxwell Running Out of Memory


hongd4356:
We are trying to use maxwell as a way to stream data from a mysql server to kinesis and eventually elastic search. We have a large amount of data in mysql that we need to bootstrap, and it appears that we are running out of memory when we attempt too many bootstraps or bootstraps with massive amounts of data. Based on the CPU usage, it looks like the KPL refuses to release memory after the bootstrap process is finished which eventually leads to a memory crash.  To be clear, we are bootstrapping by inserting into the bootstrap table and not using the bootstrap utility, and we are using the maxwell running process. 

We have a few questions:
Have you seen this issue before?
Is this related to maxwell processes or is this an issue with the KCL?
Is there a setting in the Kinesis properties that can mitigate this issue?

osheroff:
@thomasdziedzic any input here?

@hongd4356 when you say "after the bootstrap process is finished", do you mean that all the data has successfully hit kinesis?  Or is data still streaming to kinesis?

thomasdziedzic:
sorry, i don't know

gautamvij:
@osheroff 
Facing same issue, In my case the data has successfully hit kinesis and successfully processed at other end.

ede-n:
@osheroff Yes after successfully processing. We had a where clause for different ranges of ids and corresponding records in `bootstrap` table. The first row's `bootstrap.is_complete` flag is set to `1` and the process died. No errors on the Kinesis end were recorded.

gautamvij:
@osheroff 
Since, KPL stores records in memory as FutureCallback. In case of large records it pushes all records in memory leading to memory spike. 

We can throttle ingestion in kinesis stream by waiting to let backpressure be processed as mentioned in this aws [blog](https://aws.amazon.com/blogs/big-data/implementing-efficient-and-reliable-producers-with-the-amazon-kinesis-producer-library/).

@osheroff Let me know if this is acceptable, will give the PR.

osheroff:
backpressure mechanism merged and released.

PR #1141, opened 2018-10-24
Back pressure

gautamvij:
#### What
Solves memory issue in case of high DB updates and during bootstrap.
Since, KPL stores records in memory as FutureCallback. In case of large records it pushes all records in memory leading to memory spike.

For Throttling we can configure Kinesis Ingestion with fields
```
kinesis_throttle_limit
kinesis_throttle_sleep
```

#### Fixes
- https://github.com/zendesk/maxwell/issues/816
- https://github.com/zendesk/maxwell/issues/1100

#### References
Aws [blog](https://aws.amazon.com/blogs/big-data/implementing-efficient-and-reliable-producers-with-the-amazon-kinesis-producer-library/)

##### Metrics
For High ingestion you can set throttle to high number like 10000000 and sleep of 1

Tested with 
```
kinesis_throttle_limit=50000
kinesis_throttle_sleep=2000
```
Stable at 900mb ram. We can change the numbers for high throughput.

cc: @osheroff @thomasdziedzic

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -162,6 +162,8 @@ kafka.acks=1
 #            *** kinesis ***
 
 #kinesis_stream=maxwell
+#kinesis_throttle_limit=10000
+#kinesis_throttle_sleep=1000
 
 # AWS places a 256 unicode character limit on the max key length of a record
 # http://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.zendesk</groupId>
   <artifactId>maxwell</artifactId>
-  <version>1.17.1</version>
+  <version>1.19.0</version>
   <packaging>jar</packaging>
 
   <name>maxwell</name>
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -63,6 +63,8 @@ public class MaxwellConfig extends AbstractConfig {
 
 	public String kinesisStream;
 	public boolean kinesisMd5Keys;
+	public Long kinesisThrottleLimit;
+	public Long kinesisThrottleSleep;
 
 	public String sqsQueueUri;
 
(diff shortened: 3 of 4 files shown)


ede-n:
@gautamvij Could you please add more details about the test performed? That would help compare
* Shard count on Kinesis
* Instance size and type used, if on AWS

Is there a heuristic that you recommend when choosing `kinesis_throttle_limit` and `kinesis_throttle_sleep`? 

Thanks.

gautamvij:
@ede-n 
- Tested with 1 shard on a container in c4large instance with memory cap of 1.5GB. Used only 1Gb ram
- Kinesis throttles ingestion on reaching data ingestion limit as well. In order to utilise 100% of kinesis data ingestion offered per shard. Keep RateLimit as 100 instead of default 150.
- For a 2kb record, Keep kinesis_throttle_limit (Records queued in memory) to 50000, with sleep of 2 seconds. Higher shard count doesn't necessarily mean you will have less chances of memory spike.
- For more analysis you can monitor UserRecordsPending metric on AWS to monitor in memory records pending. Higher UserRecordsPending Will lead to more memory consumption.

osheroff:
hmmm..  since this the memory usage while bootstrapping seems to be a problem with a lot of slower producers (pubsub & kinesis.  I think kafka provides its own backpressure), I'd like to find a more general solution to this, maybe putting a bounded blocking-queue somewhere in the producer chain.

osheroff:
merged a competing PR

agillz:
Hi @osheroff , first of all, thanks for developing and maintaining this very useful application. I am working on a streaming architecture in AWS and planning on using maxwell to ingest data from our mysql databases. We have a exponentially growing list of schemas in our mysql instances. Currently hitting 500 and targeting 1500 in a year. Each tenant gets its own schema. In my poc, I was encountering the throttling issues from Kinesis and came across this issue. I am not exactly sure about the status on the issue. What does it mean by 'merged competing PR' ? Is the fix merged in master already? Please help.  Thanks again Aman Gill

osheroff:
We merged a PR that implemented back pressure in a different way, in version 1.19.3

PR #1159, opened 2018-11-14
add backpressure on a per-row basis

osheroff:
The first implementation of InflightMessageList's backpressure mechanism only dealt with rows that were actually tracked inside the list, meaning rows where txCommit is true.  Since no bootstrap rows satisfied this condition, there was no backpressure for producers in bootstrap.

Here we generalize backpressure to use a simple counter.  10k outstanding rows is a complete guess.

@ericwush I'm preserving the blocked-head check, would love your eyes to confirm I haven't messed stuff up here.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
+++ b/src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
@@ -8,12 +8,12 @@ package com.zendesk.maxwell.producer;
 
 import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.replication.Position;
-import com.zendesk.maxwell.row.RowMap;
 
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 
 public class InflightMessageList {
+
 
 	class InflightMessage {
 		public final Position position;
@@ -33,18 +33,18 @@ public class InflightMessageList {
 		}
 	}
 
-	private static final long INIT_CAPACITY = 1000;
+	private static final long DEFAULT_CAPACITY = 10000;
 	private static final double COMPLETE_PERCENTAGE_THRESHOLD = 0.9;
 
 	private final LinkedHashMap<Position, InflightMessage> linkedMap;
 	private final MaxwellContext context;
 	private final long capacity;
+	private long outstanding;
 	private final long producerAckTimeoutMS;
 	private final double completePercentageThreshold;
-	private volatile boolean isFull;
 
 	public InflightMessageList(MaxwellContext context) {
-		this(context, INIT_CAPACITY, COMPLETE_PERCENTAGE_THRESHOLD);
+		this(context, DEFAULT_CAPACITY, COMPLETE_PERCENTAGE_THRESHOLD);
 	}
 
 	public InflightMessageList(MaxwellContext context, long capacity, double completePercentageThreshold) {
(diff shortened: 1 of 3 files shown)


osheroff:
hmmm, I'm having trouble resolving the inflight-message-list-full code with the more general backpressure throttle.

ericwush:
I think the original mechanism kinda falls apart since `linkedMap` is no longer a critical resource, which is fine as long as we define new conditions for when we should crash maxwell. 

The original criteria for termination:
- head is stuck
- list is full (configurable)
- majority has completed (configurable)

Now 1) we don't define the capacity of the list so can no longer say it's full 2) `linkedMap` is not synchronised so the `completePercentage` is not accurate. So maybe we can get rid of `completePercentage` as well.

Instead of using counter and synchronizing the methods on the whole object (less efficient, I think), would you consider using `Semaphore` which I think makes sense here? Just imagine you have 10k or whatever permits to give away. Once all given away, the next guy wants to acquire a permit has to wait for at least one permit used by others to be released.

osheroff:
> Now 1) we don't define the capacity of the list so can no longer say it's full 2) linkedMap is not synchronised so the completePercentage is not accurate. So maybe we can get rid of completePercentage as well.

Yeah.  Hmmmm.   Maybe the heuristic here could be just "> X rows have completed while this row has remained stuck for X seconds".  

Re: semaphore vs counter;  I'm pretty sure I just implemented a poor man's semaphore.  It's fine to convert to that, though I do doubt it's more efficient than locking on the object (I think).

ericwush:
> "> X rows have completed while this row has remained stuck for X seconds"
Yeah I think it makes sense. Assume you meant "this row" is the header so we just need to track how old is the header. If we set X row = 1000 then it should have very similar effect to the current check.

osheroff:
@ericwush ok I've updated the code to basically just count messages processed since the head of the queue started.  Stuck head detection is now defined as "> 3000 messages have been completed since the head of the queue entered".  

I'm not sure this is quite enough to be safe.  I'm actually not sure there *is* a safe way to pull off this kind of check.

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -8,108 +8,110 @@
...
 
-			if (linkedMap.size() >= capacity) {
-				isFull = true;
+	public void freeSlot(long messaageID) {
+		// If the head is stuck for the length of time (configurable) and majority of the messages have completed,
+		// we assume the head will unlikely get acknowledged, hence terminate Maxwell.
+		// This gatekeeper is the last resort since if anything goes wrong,
+		// producer should have raised exceptions earlier than this point when all below conditions are met.
+		if (producerAckTimeoutMS > 0) {
+			Iterator<InflightMessage> it = iterator();
```
timbertson-zd: I don't think you're allowed to do this without synchronizing on something:

https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html
>  If multiple threads access a linked hash map concurrently, and at least one of the threads modifies the map structurally, it must be synchronized externally.

It may be that the inconsistencies we'd see by not synchronizing are benign (e.g. outdated results), but it may also be that we'd hit NullPointerExceptions, we can't know for sure.

It seems like we only need the head of the inflight list for this code. If we don't want to synchronize here, we could instead store a `volatile` field referencing the head message whenever we mutate the underlying list, then this code would only need to read that field.

src/main/java/com/zendesk/maxwell/producer/InflightMessageList.java
```diff
@@ -8,108 +8,110 @@
...
 	public InflightMessage completeMessage(Position p) {
-		synchronized (this.linkedMap) {
-			InflightMessage m = this.linkedMap.get(p);
-			assert(m != null);
-
-			m.isComplete = true;
-
-			InflightMessage completeUntil = null;
-			Iterator<InflightMessage> iterator = iterator();
+		InflightMessage m = this.linkedMap.get(p);
```
timbertson-zd: See my other comment, I don't think this is safe without `synchronized`.

timbertson-zd:
I like the approach, I'm just scared (as usual) about thread safety ;)

osheroff:
ping @timbertson @ericwush

PR #1160, opened 2018-11-14
[BUG] Schema id missing in bootstrapped rows

sas5580:
**Problem**
Schema id not appearing in rows with type `'bootstrap_insert'` even when `outputConfig.includesSchemaId` is set to true.

**Solution**
Grabbed the schema id from the replicator at the start of the bootstrap and added it every bootstrapped row. This assumes that the schema is constant throughout the bootstrap, which I believe is reasonable.

**Note**
Purposely did not add/modify any tests because output configs like binlog position are also not tested in the bootstrapped rows, even though they do appear where configured to do so. Also, considering the current bootstrap tests, it seems that it is not possible to modify the output config fo a single test, so every test would have to be modified (since they use the `runJSON` function).


osheroff:
ok, thanks for the fix!

PR #1167, opened 2018-11-28
Ignore AWS RDS internal events

ericwush:
/cc @zendesk/goanna 

Ignore more AWS RDS internal events.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -9,9 +9,7 @@ import com.github.shyiko.mysql.binlog.event.QueryEventData;
 import com.github.shyiko.mysql.binlog.event.RowsQueryEventData;
 import com.github.shyiko.mysql.binlog.event.TableMapEventData;
 import com.github.shyiko.mysql.binlog.event.deserialization.EventDeserializer;
-import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.MaxwellMysqlConfig;
-import com.zendesk.maxwell.scripting.Scripting;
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -399,20 +397,26 @@ private RowMapBuffer getTransactionRows(BinlogConnectorEvent beginEvent) throws
...
+					if ( upperCaseSql.startsWith(BinlogConnectorEvent.SAVEPOINT)) {
 						LOGGER.debug("Ignoring SAVEPOINT in transaction: " + qe);
 					} else if ( createTablePattern.matcher(sql).find() ) {
 						// CREATE TABLE `foo` SELECT * FROM `bar` will put a CREATE TABLE
 						// inside a transaction.  Note that this could, in rare cases, lead
 						// to us starting on a WRITE_ROWS event -- we sync the schema position somewhere
 						// kinda unsafe.
 						processQueryEvent(event);
-					} else if (sql.toUpperCase().startsWith("INSERT INTO MYSQL.RDS_HEARTBEAT")) {
+					} else if (upperCaseSql.contains("MYSQL.RDS_")) {
```
liulikun: Feels a bit risky here, can we be more restrictive like `upperCaseSql.startsWith("INSERT INTO MYSQL.RDS_") || upperCaseSql.startsWith("DELETE FROM MYSQL.RDS_")` ?

liulikun: Also might be a good idea to move this check to be the last case since it's rare.

ericwush: 👍 however I feel it happens more frequently than `DROP TEMPORARY TABLE` so I will leave the order unchanged.

Issue #1122, opened 2018-10-10
Truncate row instead of discarding it if RecordTooLarge exception has occurred


dongxiaohe:
**Context:**
This is a feature request, if table is too large or Kafka broker side has some certain sized limited configuration, maxwell would continue to run by throwing RecordTooLarge exception. However, the message would be discarded.

**Suggested Solution:**

- It is better we have configured option to allow message can be truncated rather than dropped.
- Also, a flag to indicate the message is _truncated_ would be ideal as well.
- What does truncated db change json looks like, what field should be truncated, it should be still a valid Json, otherwise it would break consumer side.

osheroff:
mmmmmm, sure, I can see this, but how do we go about choosing which field to truncate?  I'm also pretty sure that unless we knew the max-size of the kafka/kinesis/etc message we'd end up with out-of-order issues.  Or is there a way to pull it from the producer?

osheroff:
> With Kafka, I don't think it's doable to pull this information on producer's side. Our best bet is probably the admin client API (i.e. adminClient.describeTopics(topicNames)) and yet I don't see the required info there.

Hmm, maybe we force the user to provide the limit with something like:
`--truncate_at=1000000`.  Note that this won't really account for compression, but hey, whatever.

xuebingli:
That's a good idea. What do you think, @dongxiaohe?

dongxiaohe:
so late for the discussion :( . 

@fsat @timbertson  what do you guys think about `--truncate_at=1000000`. :)

timbertson-zd:
I was actually thinking the message would just be the primary key(s). It's less than might be useful, but it should be a simple distinction since maxwell already serializes the PK in the kafka message key.

Also with regard to ordering, we're thinking about putting it in a separate "dead letter" topic, which you'd pass as an option in order to enable this functionality. It's a little less elegant than having it inline, but the benefit is that it's not a breaking format change. It also means we can publish the information in response to a rejected error, so we don't have to decide when to reject messages ourselves.

Unfortunately this would make each producer responsible for implementing the dead letter logic. We could share the "generate a skeleton message" code, but there'd still be some producer-specific code in each case.

osheroff:
we don't truncate, but we do have fallback topics now

PR #1171, opened 2018-12-04
allow over sized kafka messages publish to a fallback topic with only primary keys

liulikun:
When `dead_letter_topic` is configured, maxwell will publish messages to this topic upon receiving RecordTooLarge exception on the original topic.

The example for the message key for the fallback topic:

```
{"database":"db1","table":"table1","pk.pk1":1,"pk.pk2":2}
```

The example for the message value for the fallback topic:

```
{"database":"db1","table":"table1","reason":"message too large","data":{"pk1":1,"pk2":2}}
```

The partition strategy for the fallback topic is the default Kafka round robin.

Solving issue #1122 

/cc @zendesk/goanna @osheroff


osheroff:
well... hm.  I'm not sure I really love the approach more or less than the truncation -- each comes with its own sets of issues (to truncate is to truncate without confidence, but to put in a separate topic is to lose ordering). 

What's your guys' plan on what you will do with these dead-letter messages?

liulikun:
I think the order is lost even truncating and republishing to the original topic. Truncating doesn't help in our use case since the full content is needed. With primary keys published to a new topic, the full content can be retrieved.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
```diff
@@ -0,0 +1,53 @@
...
+		private String recordKey;
+
+		SimpleKafkaCallback(String recordKey) {
+			this.recordKey = recordKey;
+		}
+
+		@Override
+		public void onCompletion(RecordMetadata recordMetadata, Exception e) {
+			if (e != null) {
+				LOGGER.error("ERROR for record (" + recordKey + "): " + e.getLocalizedMessage());
```
timbertson-zd: Should this be fatal?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -61,6 +63,10 @@ public void onCompletion(RecordMetadata md, Exception e) {
 			LOGGER.error(e.getLocalizedMessage());
 			if ( e instanceof RecordTooLargeException ) {
 				LOGGER.error("Considering raising max.request.size broker-side.");
+				String kafkaFallbackTopic = this.context.getConfig().kafkaFallbackTopic;
+				if (kafkaFallbackTopic != null) {
+					publishFallbackTopic(kafkaFallbackTopic);
```
timbertson-zd: We're doing `cc.markCompleted` below, which happens before the fallback message has been published, which may drop in-flight messages. What if we returned early here, and passed `cc` into the callback for the fallback publish? Or possibly just reuse this callback, as long as we don't get into a republish loop.

src/main/java/com/zendesk/maxwell/row/PrimaryKeyRowMap.java
```diff
@@ -0,0 +1,53 @@
...
+		if (reason != null) {
+			g.writeStringField("reason", reason);
+		}
+		g.writeObjectFieldStart(FieldNames.DATA);
+
+		if (primaryKeyColumns != null) {
+			for (String pk : primaryKeyColumns.keySet()) {
+				serializer.writeValueToJSON(g, true, pk.toLowerCase(), primaryKeyColumns.get(pk));
+			}
+		}
```
timbertson-zd: Does this change the format for the message keys we push to kafka? IIRC the existing format would end up with "pk.id" at the top level, looks like this changes that format to be a nested `"data":{"id": ... }`?

I like the nested `data` for the contents of the fallback message, but I don't think we want to change the kafka key format.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
```diff
@@ -0,0 +1,53 @@
...
+		if (fallBackProducer == null) {
+			synchronized (MaxwellKafkaFallBackProducer.class) {
+				if (fallBackProducer == null) {
+					fallBackProducer = new MaxwellKafkaFallBackProducer(kafkaProperties);
+				}
+			}
+		}
+
+		return fallBackProducer;
+	}
```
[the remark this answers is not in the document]
liulikun: Singleton pattern is used to make sure we only need one producer for all the oversized messages. Otherwise, creating a producer for each message seems costly.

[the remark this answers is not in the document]
timbertson-zd: We only have one instance of `MaxwellKafkaProducer` per process, we could instantiate the fallback producer at the same time (either when we build the regular producer, or in the producer's constructor)

docs/docs/config.md
```diff
@@ -54,6 +54,7 @@ javascript                     | STRING                              | file cont
 **"kafka" producer options **
 kafka.bootstrap.servers        | STRING                              | kafka brokers, given as `HOST:PORT[,HOST:PORT]`     |
 kafka_topic                    | STRING                              | kafka topic to write to.                            | maxwell
+kafka_fallback_topic           | STRING                              | kafka topic to write to when RecordTooLargeException is received on writing to the initial topic. |
```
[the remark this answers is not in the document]
liulikun: `dead_letter_topic` 👍

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Properties;
+
+public class MaxwellKafkaFallBackProducer {
```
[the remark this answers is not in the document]
liulikun: 👍

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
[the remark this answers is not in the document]
liulikun: That'll still require to pass the fallback producer around to every callback? Which is not nice.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
liulikun: maybe, depends on the settings `ignoreProducerError`?

src/main/java/com/zendesk/maxwell/row/PrimaryKeyRowMap.java
liulikun: 👍

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
liulikun: same `cc` can be used, it'll delay clearing the inflight message list a bit.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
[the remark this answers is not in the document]
timbertson-zd: It's not beautiful, but it's simple 🤷‍♂️ - we already pass around a bunch of other things, adding one more argument is less complexity than this code.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: That's what we want (I think), but is there something we can do to prevent an infinite loop if somehow the truncated message is still too large?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaFallBackProducer.java
timbertson-zd: Yes, good point

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -29,17 +31,17 @@
...
 	private Counter failedMessageCount;
 	private Meter succeededMessageMeter;
 	private Meter failedMessageMeter;
 
-	public KafkaCallback(AbstractAsyncProducer.CallbackCompleter cc, Position position, String key, String json,
-	                     Counter producedMessageCount, Counter failedMessageCount, Meter producedMessageMeter,
-	                     Meter failedMessageMeter, MaxwellContext context) {
+	public KafkaCallback(AbstractAsyncProducer.CallbackCompleter cc, Position position, PrimaryKeyRowMap key, String json,
+						 Counter producedMessageCount, Counter failedMessageCount, Meter producedMessageMeter,
+						 Meter failedMessageMeter, MaxwellContext context) {
```
osheroff: i *think* this is how I intended the whitespace.  unsure.

osheroff:
is there a reason we're doing this as a separate producer and not just a feature of the normal kafka producer class?  I feel like the functionality here is mostly the same as `ddl_topic`, and could be treated the same way (another `KafkaProducer` instance in the same class).  Unless there's something that I'm missing here?

timbertson-zd:
I agree, it doesn't seem like it needs to be separate. I've picked up this branch as Likun is on holiday (have fun!), and it's now reusing the same producer + callback classes. I've added a `ProducerContext` which encapsulates the laundry-list of references we pass to every callback (context, metrics, etc) and also the fallback topic. To prevent infinite recursion (seems unlikely, but would be bad), the fallback code path uses a derived context with the fallback topic nulled out.

I also moved all the PK-related serialization into RowIdentity, RowMap just delegates to that where necessary.

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -0,0 +1,138 @@
...
+public class RowIdentity implements Serializable {
+	private final List<AbstractMap.SimpleImmutableEntry<String, Object>> primaryKeyColumns;
+
+	private String database;
+	private String table;
+	private JsonSerializer serializer = new JsonSerializer();
+
+	// slightly less verbose pair construction
+	public static AbstractMap.SimpleImmutableEntry<String, Object> pair(String key, Object value) {
+		return new AbstractMap.SimpleImmutableEntry<>(key, value);
```
xuebingli: If we need immutable key value pairs, `org.apache.commons.lang3.tuple` package has `ImmutablePair`, which may help reduce verbosity.

```java
ImmutablePair<Integer, String> pair = new ImmutablePair<>(2, "Two");
Integer key = pair.getKey();
String value = pair.getValue();
```

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -0,0 +1,138 @@
+package com.zendesk.maxwell.row;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.*;
+
+public class RowIdentity implements Serializable {
+	private final List<AbstractMap.SimpleImmutableEntry<String, Object>> primaryKeyColumns;
```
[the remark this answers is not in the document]
timbertson-zd: > it's the first time I've seen this data structure

You and me both, java is odd sometimes 🤷‍♂️. I found it via
https://stackoverflow.com/questions/521171/a-java-collection-of-value-pairs-tuples

I'll change it to accept the interface though, there's no reason to care about specifics 🤦‍♂️.

> clone the input map to primaryKeyColumns

The full data map gets dropped after we send to kafka, to ease emory pressure. So we'd need to clone a subset of the map (just the primary key entries). But the bigger issue is ordering - maxwell serializes JSON keys with a consistent order, which a plain `map` would lose.

timbertson-zd:
> what if the dead letter queue delivery fails

It uses the same callback, but with no fallback topic. So if you configure maxwell to ignore producer errors, it'll ignore errors publishing to the fallback topic. On second thoughts, this is a bit silly - if you tell it to ignore producer errors _and_ you have a fallback topic configured, it should publish all failures there, not just "too large" errors. That's an unlikely configuration though, I'll add docs to call it out.

But if you have it configured to terminate on producer error, it'll terminate if the fallback publish fails. And if the fallback message fails because it's still somehow too big (maybe you have a stupidly large primary key column?), that just gets logged because there's nothing else we can do.

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
timbertson-zd: There's only one instance of the awkward `AbstractMap.SimpleImmutableEntry` now that I'm using the interface in most places, so this is less of an issue.

timbertson-zd:
Fair point. I've made a less invasive version of the JSON refactor, and rebased this work on top of it in three parts:

 - #1174 : (simpler) JSON refactoring
 - #1175 : pull out RowIdentity
 - #1176 : implement fallback publishing

Looks like I can't change the branch for this PR, and it would probably be confusing anyway. So I'm closing this in favour of #1176.

PR #1174, opened 2018-12-17
introduce MaxwellJson and to decouple JSON-generation logic from RowMap

timbertson-zd:
/cc @osheroff 

This is a refactoring of the JSON generation parts of #1171 

* move json-related objects into MaxwellJson, so they can be reused in #1176 and encapsulated in a single thread-local

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/MaxwellJson.java
+++ b/src/main/java/com/zendesk/maxwell/row/MaxwellJson.java
@@ -1,0 +1,96 @@
+package com.zendesk.maxwell.row;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.zendesk.maxwell.scripting.Scripting;
+import jdk.nashorn.api.scripting.ScriptObjectMirror;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.script.ScriptException;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+public class MaxwellJson {
+	private static final Logger LOGGER = LoggerFactory.getLogger(MaxwellJson.class);
+	private static final JsonFactory jsonFactory = new JsonFactory();
+
+	private final ByteArrayOutputStream buffer;
+	private final JsonGenerator jsonGenerator;
+	private final PlaintextJsonGenerator plaintextGenerator;
+	private final EncryptingJsonGenerator encryptingGenerator;
+
+	private static final ThreadLocal<MaxwellJson> instance =
+			new ThreadLocal<MaxwellJson>(){
+				@Override
+				protected MaxwellJson initialValue() {
+					try {
+						return new MaxwellJson();
+					} catch (IOException e) {
+						LOGGER.error("error initializing MaxwellJson", e);
+						return null;
+					}
+				}
+			};
+
+	private MaxwellJson() throws IOException {
+		buffer = new ByteArrayOutputStream();
+		jsonGenerator = jsonFactory.createGenerator(buffer);
+		jsonGenerator.setRootValueSeparator(null);
+		plaintextGenerator = new PlaintextJsonGenerator(jsonGenerator);
+		encryptingGenerator = new EncryptingJsonGenerator(jsonGenerator, jsonFactory);
+	}
+
+	public static MaxwellJson getInstance() {
+		return instance.get();
+	}
+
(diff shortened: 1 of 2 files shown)


timbertson-zd:
I'd be fine with that - it'd need to expose access to the underlying buffer as well though, since we need to construct _two_ json generators pointing to the same buffer (a plaintext one and an encrypting one).

That's pretty similar to Likun's initial refactor in #1171, can I ask what you didn't like about that code? I assumed you found it too drastic, so I was sticking to the minimum of what needed to change for this PR.

osheroff:
I may have just been moody at seeing a huge churn PR without a whole lot of value (it really only addresses issues of making a class smaller, nothing is really aimed towards re-usability/features).  I don't know.  I'm capricious.  Go ahead and make what changes make sense, I'll be a lot better disposed to ones that are in their own PR. 

I guess the `MaxwellJson` instance could also in theory encapsulate the business of encryption, that could be somewhat useful maybe.

timbertson-zd:
Heh, fair enough. It _was_ a big diff.

> nothing is really aimed towards re-usability/features

It came from the need to share key serialization between RowMap and RowIdentity, which we couldn't just keep in RowMap due to its lifetime (it gets dropped before the publish callback fires).

I've implemented MaxwellJson as you suggested, it's a bit nicer now that it rolls up all 3 generators into a single instance, now there's just one ThreadLocal instead of three.

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -350,8 +266,8 @@ public String toJSON(MaxwellOutputConfig outputConfig) throws Exception {
 		}
 
 		DataJsonGenerator dataWriter = outputConfig.encryptionMode == EncryptionMode.ENCRYPT_DATA
-			? encryptingJsonGeneratorThreadLocal.get()
-			: plaintextDataGeneratorThreadLocal.get();
+			? json.getEncryptingGenerator()
+			: json.getPlaintextGenerator();
```
osheroff: i see.  well, in the future this would be better if `MaxwellJson` just returned some kind of generator that had a unified interface, but this is good enough for now.

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -364,18 +280,11 @@ public String toJSON(MaxwellOutputConfig outputConfig) throws Exception {
 		g.flush();
 
 		if(outputConfig.encryptionMode == EncryptionMode.ENCRYPT_ALL){
-			String plaintext = jsonFromStream();
-			encryptingJsonGeneratorThreadLocal.get().writeEncryptedObject(plaintext, encryptionContext);
-			g.flush();
+			String plaintext = json.consume();
+			json.getEncryptingGenerator().writeEncryptedObject(plaintext, encryptionContext);
```
osheroff: what about the call to `g.flush` here?  was it superfluous?

src/main/java/com/zendesk/maxwell/row/MaxwellJson.java
```diff
@@ -0,0 +1,96 @@
...
+				@Override
+				protected MaxwellJson initialValue() {
+					try {
+						return new MaxwellJson();
+					} catch (IOException e) {
+						LOGGER.error("error initializing MaxwellJson", e);
+						return null;
+					}
+				}
+			};
```
osheroff: check out `ThreadLocal.withInitial(? -> ())` for a cleaner lambda implementation.

osheroff:
👍 except for minor comments.

src/main/java/com/zendesk/maxwell/row/RowMap.java
timbertson-zd: I considered subclassing JsonGenerator, but it's a pretty big interface, so keeping them separate was less work and makes it clear what functionality MaxwellJson actually adds. I also kind of like that you can only access the JsonGenerator by calling `reset()`, which
discourages fly-by writing to an effectively global buffer. But maybe that's me justifying my laziness ;)

src/main/java/com/zendesk/maxwell/row/RowMap.java
timbertson-zd: It's part of `MaxwellJson.consume()`. I found a couple other places where we were still calling it unnecessarily, so I've removed those too.

src/main/java/com/zendesk/maxwell/row/MaxwellJson.java
timbertson-zd: Fancy! 😉

PR #1175, opened 2018-12-17
move key-related serialization from RowMap info new RowIdentity class

timbertson-zd:
* extract primary-key related serialization from RowMap into RowIdentity

This is required work for implementing publishing of fallback records in #1176

/cc @osheroff

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowIdentity.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowIdentity.java
@@ -1,0 +1,107 @@
+package com.zendesk.maxwell.row;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.*;
+
+public class RowIdentity implements Serializable {
+	private String database;
+	private String table;
+	private final List<Map.Entry<String, Object>> primaryKeyColumns;
+
+	// slightly less verbose pair construction
+	public static Map.Entry<String, Object> pair(String key, Object value) {
+		return new AbstractMap.SimpleImmutableEntry<>(key, value);
+	}
+
+	public RowIdentity(String database, String table, List<Map.Entry<String, Object>> primaryKeyColumns) {
+		this.database = database;
+		this.table = table;
+		this.primaryKeyColumns = primaryKeyColumns;
+	}
+
+	public String getDatabase() {
+		return database;
+	}
+
+	public String getTable() {
+		return table;
+	}
+
+	public String toKeyJson(RowMap.KeyFormat keyFormat) throws IOException {
+		MaxwellJson json = MaxwellJson.getInstance();
+		JsonGenerator g = json.reset();
+		if ( keyFormat == RowMap.KeyFormat.HASH )
+			writeKeyJsonHash(g);
+		else
+			writeKeyJsonArray(g);
+		return json.consume();
+	}
+
+	private void writeKeyJsonHash(JsonGenerator g) throws IOException {
+		writeStartCommon(g);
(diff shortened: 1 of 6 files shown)


timbertson-zd:
Updated to point at `master`, /cc @zendesk/goanna

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class RowIdentity implements Serializable {
+	private String database;
+	private String table;
+	private final List<Map.Entry<String, Object>> primaryKeyColumns;
+
+	// slightly less verbose pair construction
+	public static Map.Entry<String, Object> pair(String key, Object value) {
+		return new AbstractMap.SimpleImmutableEntry<>(key, value);
+	}
```
osheroff: elsewhere we use `org.apache.commons.lang3.tuple.Pair`, which is probably cleaner than all of this...

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -0,0 +1,107 @@
...
+		JsonGenerator g = json.reset();
+		if ( keyFormat == RowMap.KeyFormat.HASH )
+			writeKeyJsonHash(g);
+		else
+			writeKeyJsonArray(g);
+		return json.consume();
+	}
+
+	private void writeKeyJsonHash(JsonGenerator g) throws IOException {
+		writeStartCommon(g);
```
osheroff: no need for this to be a function call, you're not calling it more than once.

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -0,0 +1,107 @@
...
+		for (Map.Entry<String,Object> pk : primaryKeyColumns) {
+			g.writeStartObject();
+			MaxwellJson.writeValueToJSON(g, true, pk.getKey().toLowerCase(), pk.getValue());
+			g.writeEndObject();
+		}
+		g.writeEndArray();
+		g.writeEndArray();
+	}
+
+	public String toConcatString() {
```
osheroff: while you're in here let's add some minimal documentation to these functions, just explain the format each is outputting and where it's used...

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -0,0 +1,107 @@
...
+		return json.consume();
+	}
+
+	private void writeKeyJsonHash(JsonGenerator g) throws IOException {
+		writeStartCommon(g);
+		if (primaryKeyColumns.isEmpty()) {
+			g.writeStringField(FieldNames.UUID, UUID.randomUUID().toString());
+		} else {
+			for (Map.Entry<String,Object> pk : primaryKeyColumns) {
+				writePrimaryKey(g, "pk." + pk.getKey().toLowerCase(), pk.getValue());
```
osheroff: same note here, if you've got a one-line function that's called exactly once you're doing something off...

timbertson-zd: This was because I pulled it out of #1176, so it will eventually be used multiple times. But I've inlined it for now.

src/main/java/com/zendesk/maxwell/row/RowIdentity.java
timbertson-zd: Sure, done in e63680bcc76ffdc17f70664316d28dcc9536259b

timbertson-zd:
@osheroff 💇‍♂️

PR #1176, opened 2018-12-17
add support for publishing failed messages to a fallback topic when configured

timbertson-zd:
This is a refactored attempt at #1171 , building on top of #1175 .

/cc @osheroff

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -96,11 +96,7 @@ class KafkaCallback implements Callback {
 		KafkaCallback cb = new KafkaCallback(cc, position, key, json,
 			succeededMessageCount, failedMessageCount, succeededMessageMeter,
 			failedMessageMeter, null, context, producer);
-		try {
-			producer.sendFallbackAsync(fallbackTopic, key, cb, e);
-		} catch (Exception fallbackEx) {
-			cb.onCompletion(md, fallbackEx);
-		}
+		producer.sendFallbackAsync(fallbackTopic, key, cb, md, e);
 	}
 
 	String getFallbackTopic() {
@@ -240,9 +236,13 @@ class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnab
 		sendAsync(record, callback);
 	}
 
-	public void sendFallbackAsync(String topic, RowIdentity fallbackRecord, KafkaCallback callback, Exception reason) throws Exception {
-		ProducerRecord<String, String> record = makeFallbackRecord(topic, fallbackRecord, reason);
-		sendAsync(record, callback);
+	public void sendFallbackAsync(String topic, RowIdentity fallbackRecord, KafkaCallback callback, RecordMetadata md, Exception reason) {
+		try {
+			ProducerRecord<String, String> record = makeFallbackRecord(topic, fallbackRecord, reason);
+			sendAsync(record, callback);
+		} catch (Exception fallbackEx) {
+			callback.onCompletion(md, fallbackEx);
+		}
 	}
 
 	void sendAsync(ProducerRecord<String, String> record, Callback callback) {
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -5,6 +5,7 @@ import com.codahale.metrics.Meter;
 import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.producer.partitioners.MaxwellKafkaPartitioner;
 import com.zendesk.maxwell.replication.Position;
+import com.zendesk.maxwell.row.RowIdentity;
 import com.zendesk.maxwell.row.RowMap;
 import com.zendesk.maxwell.row.RowMap.KeyFormat;
 import com.zendesk.maxwell.schema.ddl.DDLMap;
@@ -29,7 +30,9 @@ class KafkaCallback implements Callback {
 	private final AbstractAsyncProducer.CallbackCompleter cc;
 	private final Position position;
 	private final String json;
-	private final String key;
+	private final RowIdentity key;
+	private final String fallbackTopic;
+	private final MaxwellKafkaProducerWorker producer;
 	private final MaxwellContext context;
 
 	private Counter succeededMessageCount;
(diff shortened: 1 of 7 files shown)


timbertson-zd:
Okay, this last one (I promise) is ready for review now - /cc @osheroff @zendesk/goanna

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -25,60 +26,107 @@
...
-				this.context.terminate(e);
-				return;
+
+			boolean nonFatal = e instanceof RecordTooLargeException || context.maxwell.getConfig().ignoreProducerError;
+
+			if (nonFatal) {
+				if (context.fallbackTopic == null) {
+					cc.markCompleted();
+				} else {
+					publishFallback(md, e);
```
liulikun: We only need to publish for `RecordTooLargeException`.

timbertson-zd: If you set `ignore_producer_errors` to `true`, and you set a fallback topic, this will publish every failed message to the fallback topic. This is easy to describe ("we either publish a successful message or a fallback one"), whereas only publishing a fallback record for _some_ errors and quietly dropping others seems confusing.

liulikun: I thought the purpose of this fallback topic is only for the messages which are too big for the original topic. Otherwise the messages destined for the original topic are ended up in the wrong topic (for example in the case of `TimeoutException`).

liulikun: 👍

osheroff:
again, refactors as part of feature development PRs are great for one man shows.  But me, my head spins.  Can I see what this functionality looks like without the `ProducerContext` refactor?

timbertson-zd:
`ProducerContext` is just pulling out a structure to hold "the stuff we pass to callbacks", because I needed to give `KafkaCallback` another two constructor arguments (`fallbackTopic` and `producer`). The alternative was to end up with 11 constructor arguments, which got a little out of hand.

I've wound this back and split up into two commits, so you can review them independently - the first commit adds the functionality, the second adds ProducerContext.

The added tests are only in the refactored commit though, because making them work without an intermediate object was way too tedious.

osheroff:
I'm kinda meh in general on context objects; I mean is this really better? 

```
this.producerContext = new KafkaCallback.ProducerContext(this, context, context.getConfig().deadLetterTopic,
				succeededMessageCount, failedMessageCount, succeededMessageMeter, failedMessageMeter);
```

In general I'm kinda hating `MaxwellContext`, and trying to avoid it where I can; it makes all the tools less composable and does far too many things at once.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
osheroff: I wonder if it wouldn't be better to pass the fallback message back through the worker queue?  Or are we too far removed from that?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -203,17 +228,24 @@ public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) thro
 		/* if debug logging isn't enabled, release the reference to `value`, which can ease memory pressure somewhat */
 		String value = KafkaCallback.LOGGER.isDebugEnabled() ? record.value() : null;
 
-		KafkaCallback callback = new KafkaCallback(cc, r.getNextPosition(), record.key(), value,
-				this.succeededMessageCount, this.failedMessageCount, this.succeededMessageMeter, this.failedMessageMeter, this.context);
+		KafkaCallback callback = new KafkaCallback(cc, r.getNextPosition(), r.getRowIdentity(), value,
+				this.succeededMessageCount, this.failedMessageCount, this.succeededMessageMeter, this.failedMessageMeter,
+				this.context.getConfig().deadLetterTopic, this.context, this);
```
osheroff: maybe we fish `deadLetterTopic` out of config sooner than this line....

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: Good question. The current code _is_ threadsafe (since the producer is threadsafe, and we don't rely on other mutable state), though it lays a bit of a trap for introducing race conditions in the future.

It would mean you'd delay progress of in flight messages, which I don't like. The fallback item would go to the back of the queue, coming after up to 100 items which were after this one in the input partition. The queue has a fixed capacity, but I don't think it would ever get stuck since the worker doesn't wait for the callback completion.

It's also may just be too much of a pain, since that queue contains `RowMap`s, and this isn't much of a `RowMap`. You could introduce a new class encapsulating "rowmap or fallback record", but I'm not convinced it'd be worth the trouble.

timbertson-zd:
@osheroff any more feedback? I'd quite like to get this merged soon if you're OK with it.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -75,9 +86,25 @@ public void onCompletion(RecordMetadata md, Exception e) {
...
+	}
+
+	private void publishFallback(RecordMetadata md, Exception e) {
+		// When publishing a fallback record, make a callback
+		// with no fallback topic to avoid infinite loops
+		KafkaCallback cb = new KafkaCallback(cc, position, key, json,
+			succeededMessageCount, failedMessageCount, succeededMessageMeter,
+			failedMessageMeter, null, context, producer);
+		try {
+			producer.sendFallbackAsync(fallbackTopic, key, cb, e);
```
osheroff: i think maybe the logic is too fractured between this call point and `sendFallbackAsync` -- it seems like they're splitting the duties a bit too much.  Maybe here we call `producer.fallback(this, e)` and then it does the rest of the work of creating a new callback etc?  

sorry to go another rev, I'm just having trouble following all the tendrils of logic here.

timbertson-zd: So it's easy enough to move the try/catch out of here, you just need to pass the `RecordMetadata` on to the publisher (done in 9cb2a91ceba0992e058c3ed72e7e36382053cb9e). But if you want to move the creation of the callback into the publisher, you'd need to make the ~9 private fields used for construction visible to the publisher (or wrap them with a content object 😛 ).

You'd also need to move the assertions in https://github.com/zendesk/maxwell/blob/3ecbebd0e48b8e8706f7e2a7d71e759117f33d72/src/test/java/com/zendesk/maxwell/producer/KafkaCallbackTest.java#L116-L118 (which check that the fallback topic is `"dead_letters"` on the first error, then `null` on the second) into a new test in MaxwellKafkaProducerWorkerTest. Which is not hard, just tedious due to all the mocking.

So I can do that, but I'm not convinced it would end up particularly nice. To be honest, I really don't mind how it's structured so long as it works and is tested. Would you prefer to have a go at refactoring this branch however you'd like it to be structured?

osheroff:
fuck it 👍

osheroff:
apologies for the delay.  merge at will and I'll release

timbertson-zd:
😆 sorry this ended up needing so much back & forth, always happy to revisit if you have ideas on improving it 👍

Issue #1129, opened 2018-10-15
Unhandled QueryEvent inside transaction


jkrcma:
We have quite a big Percona 5.6 setup running, GTID, 4T database, over 1k transactions per second, Maxwell is connected to a slave node and replicates to RabbitMQ cluster. Repeatedly I observe this behavior to happen:

```
06:53:48,767 WARN  BinlogConnectorReplicator - Unhandled QueryEvent inside transaction: QueryEventData{threadId=1593427976, executionTime=17, errorCode=0, database='', sql='BEGIN'}
```

Once this is logged the whe whole binlog which is being replicated at the moment is ignored, until the next binlog file is opened at replication node. A big GTID gap counting tens of thousand transactions is created, but Maxwell doesn't try to catch it up at all. After there is a critical number of gaps present in the GTID set we are unable to get rid of them as the replication node already purged its binlogs.

I tried to get more deeply into the binlogs in order to find out what exactly causes the *unhandled QueryEvent* but I haven't found anything particularly useful.

osheroff:
is there any logging information before the `unhandled QueryEvent`?

osheroff:
wondering if there's a way we could locate where in the binlog maxwell is failing; unhandled query event seems to indicate that maxwell missed the `COMMIT` of the previous transaction somehow, but I'm... really not sure quite how.  getting a `SHOW BINLOG EVENTS` around the relevant section of the binlog could help.

jkrcma:
There is no information before the event.

I located the particular transaction using `mysqlbinlog` tool but didn't see anything eye-catching besides the `BEGIN` statement itself, I didn't even find any similarity in the transaction contents. Should I look backwards to check the previous statements?

osheroff:
yeah, I'm more interested in the statement directly *before* the BEGIN that triggered `unhandled query event`

jkrcma:
Assuming that the transaction *before* is the one at the edge of GTID interval, then we are talking about a *huuuuge* transaction in today's case. There was definitely a big batch of large transactions going on the past few minutes before the unhandled event happened. At this time there was also a binary backup running at the same node, causing a bit more of replication delay.

I've isolated the transaction into a text file, but it looks complete, including the `COMMIT` statement. Rough stats are:

* 114 inserts
* 284 updates
* 497 deletes

The transaction *before* the GTID edge is a small single-statement UPDATE query.

osheroff:
[...]

PR #1186, opened 2019-01-10
gtid-reconnects

osheroff:
This is a PR to address the situation brought up in #1129.  The summary is that on unstable networks (kubernetes, in this case), maxwell was being disconnected from mysql in a great deal of locations, including in the middle of a large GTID transaction.  When this would happen we'd end up accidentally advancing the binlog position *past* the end of the transaction, hilarity ensues.

To fix this, first we bring in a PR for binlog connector: https://github.com/shyiko/mysql-binlog-connector-java/pull/250 that delays updating the GTID set until we hit the COMMIT or XID event.  This allows us to restart the binary log client in the correct spot.  Then, we modify `getTransactionEvents` to bail out when we detect a restart.

Any/all comments appreciated @ericwush @timbertson @liulikun.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -115,6 +115,7 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 
 		/* setup binlog */
 		this.binlogLifecycleListener = new BinlogConnectorLifecycleListener();
+
 		this.client = new BinaryLogClient(mysqlConfig.host, mysqlConfig.port, mysqlConfig.user, mysqlConfig.password);
 
 		this.client.setSSLMode(mysqlConfig.sslMode);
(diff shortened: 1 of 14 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -492,10 +522,12 @@ public RowMap getRow() throws Exception {
...
-						rowBuffer = getTransactionRows(event);
-						rowBuffer.setServerId(event.getEvent().getHeader().getServerId());
-						rowBuffer.setThreadId(qe.getThreadId());
-						rowBuffer.setSchemaId(getSchemaId());
+						try {
+							rowBuffer = getTransactionRows(event);
+							rowBuffer.setServerId(event.getEvent().getHeader().getServerId());
+							rowBuffer.setThreadId(qe.getThreadId());
+							rowBuffer.setSchemaId(getSchemaId());
+						} catch ( ClientReconnectedException e ) {}
```
timbertson-zd: When this throws, presumably it's from `getTransactionRows()`? In which case `rowBuffer` remains whatever was in there last time around and we carry on processing, is that a problem? Should we `null` it in the catch block or something?

timbertson-zd:
Belated 👍, just had one question

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
osheroff: if you look at line 481 or so, it's more or less guaranteed that `rowBuffer` is null or empty by the time we reach this part of the code (part of the goofy code structure that inverts pushing into polling).  I don't mind and explicit `rowBuffer = null` but it's not strictly necessary afaict.

timbertson-zd: Ah right, `rowBuffer` never sticks around because we drain it at the start of each loop. I believe you, although if a later refactoring made it possible for the following lines (`rowBuffer.set*`) to throw, it would become a problem. I've opened #1191 for the sake of paranoia.

Issue #1194, opened 2019-01-21
Use CP-1252 to decode latin1 instead of ISO-8859-1


borleaandrei:
I have a problem with maxwell's decoding. The database I am working with uses latin1 encoded with CP-1252 which is an extention of (ISO-8859-1) but when maxwell reads data from the database it  decodes it with ISO-8859-1, so all the extended list of characters from CP-1252 are **LOST**.
Wouldn't it be better if maxwell would decode latin1 with CP-1252 ? Having in mind that CP-1252 is the same as ISO-8859-1 only that it has a few more characters.

osheroff:
yeah, after a trip through this issue it looks like you're correct.   Feel free to open a PR (with a test, please), otherwise I'll fix it when time permits.

borleaandrei:
Ok. I opened a pr #1197

osheroff:
fixed in v1.19.7

PR #1197, opened 2019-01-22
Change latin1 & ascii decoding charset to Windows-1252

borleaandrei:
This the solution for Issue #1194


osheroff:
i generally just release at will or when there's enough to warrant it.  will get one out today.

PR #1200, opened 2019-01-25
Support partitioning by transaction ID

hexene:
This allows all changes in a transaction to be processed by the same consumer.


osheroff:
thx for the PR.  looks generally good, will merge when I kick the build over and it goes green.

Issue #1207, opened 2019-02-06
Schema deltas compaction


jakubkulhan:
We have Maxwell running against our staging database. Every night we drop all tables in the database, create again from current schema, and fill with sample data. This causes Maxwell to record quite a lot schema deltas. I've looked into the `maxwell` database and everything refers to `schema_id` `1` and then there's a lot of records in `schemas` table. As a consequence when Maxwell starts it has to replay all the deltas. This causes a lot of overhead and high memory usage.

I've seen that some optimization was done in #1080, so we've upgraded from 1.17.1 to latest version 1.19.7. However, this was to no effect. Memory usage and startup time stayed about the same.

Is there a way to compact schema deltas? If not, would you consider adding such feature (e.g. to commit schema version older than X days and remove older deltas)?

osheroff:
yeah #1080 just deals with the held-in-memory expanded representation of the schema, not with deltas.

I would definitely at least take a patch to compact schemas.  are you offering to do dev work?  unclear from your message :)

jakubkulhan:
@osheroff I'll try to take a look at it. Could you give me pointers how to approach the implementation and where to look in the source?

osheroff:
start here:

https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java#L147

in `saveDerivedSchema`.  

You've seen that Maxwell represents an evolving schema as a linked list, with the full capture at the head of the linked list and a series of deltas chaining off that.  The list is built by the schemas pointing backwards at a previous `baseSchemaID`.

The trick here would be that after the chain grows to N elements (or is N days old, whatever), we break the chain in the middle and nominate a new HEAD of the list.  

If I were implementing this I'd probably do something like this:

- somewhere in `saveDerivedSchema` check the length of the chain (or whatever measurement we're using the break the chain).  If we decide we want to compact the schema, we fire off a new thread.
- this thread restores a duplicate copy of the schema (hmmm, could be memory issues) and starts walking the chain.  When it detects it wants to compact at that point, it:

1.  does the equivalent of `saveSchema` - saves a full copy of the schema as it stands into the various tables. 
2.  removes the back pointer to the previous `base_schema_id`.
3.  lazily starts deleting the old schema.

Things to note:

- this schema will have both a full copy of the schema and a `delta`, will have to make sure that we process that correctly. 

if you decide to take it on, feel free to open PRs early and ask for whatever you need.

[...]

PR #1208, opened 2019-02-11
Schema deltas compaction

jakubkulhan:
PR for #1207


osheroff:
Please turn off the feature of your editor that is reformatting everything....

jakubkulhan:
@osheroff I've run reformat only on files with a lot of changes and `MysqlSchemaStoreTest` had broken indentation. I haven't found any code style rules in the repo and I haven't been able to deduct what the rules are from the sources (every class seems to be formatted differently). I'll try to keep formatting changes to the minimum.

osheroff:
I just can't review a PR where like 300 lines of code change in a file, so yeah, ok, the formatting wasn't all that great, but I'm basically forced to reject a PR that reformats things.  I'm looking over the PR and a lot of these whitespace changes I *don't* like.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -70,26 +60,28 @@ public MysqlSavedSchema(MaxwellContext context, Schema schema, Position position
...
+		}
 	}
 
 	public Long getSchemaID() {
 		return schemaID;
 	}
 
 	private static Long executeInsert(PreparedStatement preparedStatement,
-			Object... values) throws SQLException {
+									  Object... values) throws SQLException {
```
osheroff: like this, this is a terrible whitespace change...

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -289,15 +282,16 @@ public static MysqlSavedSchema restore(MaxwellContext context, Position targetPo
...
 
 	public static MysqlSavedSchema restore(
-		ConnectionPool pool,
-		Long serverID,
-		CaseSensitivity caseSensitivity,
-		Position targetPosition
+			ConnectionPool pool,
+			Long serverID,
+			CaseSensitivity caseSensitivity,
+			Position targetPosition
```
osheroff: as is this...

jakubkulhan:
I understand, I'll revert unnecessary changes. Thank you for your time so far!

Regarding the feature itself:

1. In #1207 you've written that compaction might be handled in `MysqlSavedSchema.saveDerivedSchema()`. However, after I've walked through the code, I think more appropriate would be to have a separate method that gets called in `MysqlSchemaStore.saveSchema()`. Do you agree?
2. Do you think it's essential to do cleanup work in a separate thread? I think latency caused by cleanups won't be large enough, if properly configured, to justify added complexity of the implementation.

osheroff:
> 1. However, after I've walked through the code, I think more appropriate would be to have a separate method that gets called in `MysqlSchemaStore.saveSchema()`. Do you agree?

I... do not.  I think that the business of maintaining the chain (and knowing when we want to break the chain) should stay inside `MysqlSavedSchema`.

> 2. Do you think it's essential to do cleanup work in a separate thread?

I think it'd be OK to commit the new, full schema while blocking maxwell, but I think it's probably for the best to do DELETEs on the old orphan schemas on a separate thread, unless it's super-unreasonably-hard.

jakubkulhan:
Hi. Finally I've had some time to work on the compaction more:

- I reverted back unnecessary whitespace changes.
- Compaction logic is contained in `MysqlSavedSchema.compact()` method. `MysqlSchemaStore.saveSchema()` triggers it after the `save()`.
- Clean-up is run in the same transaction that saves new full schema to DB. It's 4 `DELETE`s, all on columns with indexes - should be really fast. This way it cannot happen that there would remain old schemas in case of a server crash.

I believe the PR is ready for review.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -245,6 +247,8 @@ protected OptionParser buildOptionParser() {
 		parser.accepts( "gtid_mode", "(experimental) enable gtid mode" ).withOptionalArg();
 		parser.accepts( "ignore_producer_error", "Maxwell will be terminated on kafka/kinesis errors when false. Otherwise, those producer errors are only logged. Default to true" ).withOptionalArg();
 		parser.accepts( "recapture_schema", "recapture the latest schema" ).withOptionalArg();
+		parser.accepts( "schema_chain_length", "length of schema chain after compaction" ).withRequiredArg();
```
osheroff: let's just make this 1.

src/main/java/com/zendesk/maxwell/schema/AbstractSchemaStore.java
```diff
@@ -1,72 +0,0 @@
...
-				}
-			} else {
-				LOGGER.debug("ignoring blacklisted schema change");
-			}
-		}
-		return resolvedSchemaChanges;
-	}
-}
-
-
```
osheroff: please put back `AbstractSchemaStore` -- I know it doesn't really do much right now, but it's a sort of nice someday dream that I'll write a `KafkaSchemaStore`

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -761,4 +754,199 @@ public static String getSchemaPositionSHA(Long serverID, Position position) {
 		);
 		return DigestUtils.shaHex(shaString);
 	}
+
+	public int getSchemaChainLength() {
+		return schemaChain.length();
+	}
+
+	public int compact(Connection connection, int chainMaxLength, int chainLength) throws SQLException {
```
osheroff: general thoughts on compaction:  

- let's simplify the code by always truncating the schema all the way down.  I don't think there's a use case for keeping chains around.

- If possible (and I'm not sure 100% sure this is right), I think it's better to insert a new base schema with no back-reference than to re-use the old schemas.  This would let us someday insert the new schema synchronously and then lazily delete the old ones.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -761,4 +754,199 @@ public static String getSchemaPositionSHA(Long serverID, Position position) {
...
+
+		schemaChain = newSchemaChain;
+		if (schemaChain.length() == 0) {
+			deltas = null;
+		}
+
+		return removeIds.length;
+	}
+
+	private static class SchemaChain implements Iterable<Long> {
```
osheroff: confused as to why we're implementing our own linked list.  shouldn't this just be `extends LinkedList<Long>`?  

Actually I'm utterly confused by the value of this, other than maybe the `id()` method....

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaStore.java
```diff
@@ -6,54 +6,57 @@
...
-			context.getServerID(),
-			initialPosition,
-			context.getCaseSensitivity(),
-			context.getFilter(),
-			context.getReplayMode()
+				context,
+				context.getServerID(),
+				initialPosition,
+				context.getFilter(),
+				context.getReplayMode()
```
osheroff: no need to re-tab.  one tab was correct.

src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -76,12 +76,9 @@ private Position attemptMasterRecovery() throws Exception {
 				// load up the schema from the recovery position and chain it into the
 				// new server_id
 				MysqlSchemaStore oldServerSchemaStore = new MysqlSchemaStore(
-					context.getMaxwellConnectionPool(),
-					context.getReplicationConnectionPool(),
-					context.getSchemaConnectionPool(),
+					context,
```
osheroff: Please keep long argument lists rather than passing around `context` -- it makes the components far more reusable when using maxwell as a library.  I know it's a pain, but that's the reasoning.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: (or whatever means, compact all the way down to 1 schema.)

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -761,4 +754,199 @@ public static String getSchemaPositionSHA(Long serverID, Position position) {
...
+			try (PreparedStatement stmt = connection.prepareStatement(
+				"DELETE FROM `schemas` WHERE `id` IN (" + inSQL + ")"
+			)) {
+				for (int i = 0; i < removeIds.length; i++) {
+					stmt.setLong(i + 1, removeIds[i]);
+				}
+				stmt.executeUpdate();
+			}
+
+			connection.commit();
```
osheroff: this method is really really long, let's at least break it up into two sub-methods, compaction and cleanup.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
jakubkulhan: Ok, I've removed this option and the whole chain always gets removed.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
jakubkulhan: - OK, whole chain is always removed now.
- Inserting new schema would violate unique key on `position_sha`.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
jakubkulhan: LinkedList's nodes are mutable, therefore creating copy of LinkedList always re-creates the whole chain. This is an optimization with immutable nodes - creating new chain reuses existing nodes, only allocates new head. Might be useless optimization. If so, I think it would be better to use array of IDs / ArrayList as they have better data locality.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
jakubkulhan: After removing code for removing only part of the chain, I believe this is not necessary anymore?

osheroff: no, I'd still like you to break it in two.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
osheroff: After reading it and trying to think through it a bunch, I think that `SchemaChain` as written is in a weird middle-ground; it's not doing enough that's specific to the chain, and not general enough to be just called "ImmutableLinkedList" or whatever.

I'd say, either: 
- throw it out and use generic data structures; an array is fine, *or*:
- go ahead and add *more* functionality into `SchemaChain` so that it becomes responsible for more stuff, I could easily see all the logic in `restoreDerivedSchema` living inside SchemaChain.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -761,4 +754,152 @@ public static String getSchemaPositionSHA(Long serverID, Position position) {
...
+			}
+
+			// clean up
+			StringBuilder inSQL = new StringBuilder();
+			for (int i = 0; i < schemaChain.length(); i++) {
+				if (inSQL.length() > 0) {
+					inSQL.append(", ");
+				}
+				inSQL.append("?");
+			}
```
osheroff: no need to go to all this trouble to use prepared statements.  Just build up the IN list of schema_ids once and then inject it directly into all the DELETE queries.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
osheroff: gotcha, right `position_sha`.  ok.

mpeyper:
I'm very interested in having this feature (we have a use case for automated e2e tests that create and drop entire databases on a very frequent basis and it doesn't take long before Maxwell crashes on start up.  I've got the changes up to date on master now running, and I'm happy to push them up into a (new?) PR, but I wanted to get a feel for where you're thinking is at with this at the moment.  

Is this still the direction you want to take it?

Is there another way to tackle the issue?

Are there any outstanding changes that would need to be made before this can be merged?

> general question: how will this code behave with two concurrent clients? Note that they share the same database of the schema. will they both try to compact the schema?

We only run a single instance, but for the sake of the PR and my own curiosity I set up a local example with multiple instances to see what would happen.

When the schema chain length exceeds the max, one instance successfully compacts the change and continues running as normal, the other instance crashes with the following error:

```txt
08:52:43,655 ERROR TaskManager - cause: 
com.zendesk.maxwell.schema.SchemaStoreException: com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException: Deadlock found when trying to get lock; try restarting transaction
	at com.zendesk.maxwell.schema.MysqlSchemaStore.processSQL(MysqlSchemaStore.java:118)
	at com.zendesk.maxwell.replication.BinlogConnectorReplicator.processQueryEvent(BinlogConnectorReplicator.java:316)
	at com.zendesk.maxwell.replication.BinlogConnectorReplicator.processQueryEvent(BinlogConnectorReplicator.java:338)
	at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getRow(BinlogConnectorReplicator.java:610)
	at com.zendesk.maxwell.replication.BinlogConnectorReplicator.work(BinlogConnectorReplicator.java:180)
	at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:34)
	at com.zendesk.maxwell.Maxwell.startInner(Maxwell.java:224)
	at com.zendesk.maxwell.Maxwell.start(Maxwell.java:157)
	at com.zendesk.maxwell.Maxwell.main(Maxwell.java:245)
Caused by: com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException: Deadlock found when trying to get lock; try restarting transaction
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:123)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:953)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeUpdateInternal(ClientPreparedStatement.java:1092)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeUpdateInternal(ClientPreparedStatement.java:1040)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeLargeUpdate(ClientPreparedStatement.java:1340)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeUpdate(ClientPreparedStatement.java:1025)
	at snaq.db.CachedPreparedStatement.executeUpdate(CachedPreparedStatement.java:151)
	at com.zendesk.maxwell.schema.MysqlSavedSchema.compact(MysqlSavedSchema.java:801)
	at com.zendesk.maxwell.schema.MysqlSchemaStore.saveSchema(MysqlSchemaStore.java:131)
	at com.zendesk.maxwell.schema.MysqlSchemaStore.processSQL(MysqlSchemaStore.java:115)
	... 8 more
```

Interestingly, this error occurs even if the two instances have significantly different `schema_chain_max_length` lengths (e.g. 10 and 100).  The compacting seems to mess with the other one trying to update the table with the next schema (which makes some sense to me).

Any idea how we might be able to work around this issue?  I'm very new to the codebase, but happy to take a stab at it with some direction.

osheroff:
@mpeyper thanks for the interest, and for looking into the concurrent client issue.  Given what you saw (seems like we're locking the database for too long while we do compaction), and what I've reasoned through (which is that this code is not safe for mulitple clients), I think we'll need to take a different tack here.

Here's what I'm thinking: 

- compaction must move to a background thread
- only one client may be allowed to compact at one time; there should be a "lock" in the database in the form of a single row table, `maxwell.compactions`. 
- compaction may only happen on schemas that all clients have already processed, so, let's say:

- client A is caught up and is processing data for schema 28
- client B is lagged several days and processing data for schema 14

in this case we can compact up to schema 14, and use that as the new "base" or "full" schema, but we can go no further, otherwise client B will end up re-inserting schemas 15-28.  

Let me know if you'd like to give writing the code a shot.  I'd also be happy to put in the time myself, given the general thorniness of the issue.

mpeyper:
> Let me know if you'd like to give writing the code a shot. I'd also be happy to put in the time myself, given the general thorniness of the issue.

I think, given the new complexities, that you might be better positioned to make these changes.  I'm happy to help, either by pushing my branch to give you an up to date starting point if you want to base them of these changes, or by testing the changes you're making (I've got a sample environment running already).

PR #1212, opened 2019-02-14
Minor addition and correction to quickstart guide

nickpad:
I noticed that MySQL 8 is not on the supported versions list even though it's supported according to the changelog.

Also, as far as I can see the DDL statements for setting up the `maxwell` user in the quickstart guide aren't quite right - a `CREATE USER` statement is needed - so I've updated those.

### DIFF
--- a/docs/docs/compat.md
+++ b/docs/docs/compat.md
@@ -1,7 +1,7 @@
 ### Requirements:
 ***
 - JRE 7 or above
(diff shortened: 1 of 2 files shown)


nickpad:
BTW this is an awesome project, thanks for creating it!

osheroff:
ah, yeah, in mysql 8 you don't get to create users implicitly with GRANT.  Thanks for the PR!

PR #1226, opened 2019-03-20
getInitialPosition(): do not take other client's position if the gtidMode is on

flaneur2020:
Hello,

Recently I tried to enable `gtid_mode = true` in production, but have got this error:

```
17:57:52,656 INFO  BinlogConnectorLifecycleListener - Binlog disconnected.
17:57:52,713 WARN  BinlogConnectorReplicator - replicator stopped at position: ac3d712f-51eb-11e8-a672-e04f43ca7a68:1-960618007,be3510e0-fdda-11e8-9ae8-e04f43ca9e26:1-854502567,f9fe5e51-3b48-11e9-b6db-6c0b84d5cfe8:1-235523016 -- restarting
17:57:52,757 INFO  BinaryLogClient - Connected to mysql-xx-master.intra.prod:27010 at ac3d712f-51eb-11e8-a672-e04f43ca7a68:1-960618007,be3510e0-fdda-11e8-9ae8-e04f43ca9e26:1-854502567,f9fe5e51-3b48-11e9-b6db-6c0b84d5cfe8:1-235523016 (sid:6379, cid:10738774)
17:57:52,757 INFO  BinlogConnectorLifecycleListener - Binlog connected.
17:57:52,800 WARN  BinlogConnectorLifecycleListener - Communication failure.
com.github.shyiko.mysql.binlog.network.ServerException: The slave is connecting using CHANGE MASTER TO MASTER_AUTO_POSITION = 1, but the master has purged binary logs containing GTIDs that the slave requires.
    at com.github.shyiko.mysql.binlog.BinaryLogClient.listenForEventPackets(BinaryLogClient.java:882) [mysql-binlog-connector-java-0.16.1.jar:0.16.1]
    at com.github.shyiko.mysql.binlog.BinaryLogClient.connect(BinaryLogClient.java:559) [mysql-binlog-connector-java-0.16.1.jar:0.16.1]
    at com.github.shyiko.mysql.binlog.BinaryLogClient$7.run(BinaryLogClient.java:793) [mysql-binlog-connector-java-0.16.1.jar:0.16.1]
    at java.lang.Thread.run(Thread.java:745) [?:1.8.0_45]
17:57:52,800 INFO  BinlogConnectorLifecycleListener - Binlog disconnected.
17:57:52,857 WARN  BinlogConnectorReplicator - replicator stopped at position: ac3d712f-51eb-11e8-a672-e04f43ca7a68:1-960618007,be3510e0-fdda-11e8-9ae8-e04f43ca9e26:1-854502567,f9fe5e51-3b48-11e9-b6db-6c0b84d5cfe8:1-235523016 -- restarting
17:57:52,905 INFO  BinaryLogClient - Connected to mysql-xx-master.intra.prod:27010 at ac3d712f-51eb-11e8-a672-e04f43ca7a68:1-960618007,be3510e0-fdda-11e8-9ae8-e04f43ca9e26:1-854502567,f9fe5e51-3b48-11e9-b6db-6c0b84d5cfe8:1-235523016 (sid:6379, cid:10738775)
```

After investigation, I found the root cause is Maxwell took another database's GTID position after restart. The mysql master don't have the GTID set, so the error is reported.

Screen Shot 2019-03-20 at 6 02 45 PM

`getLatestFromAnyClient()` takes the last position which `server_id = ?`, but in the GTID mode, all the positions' server_id are `DEFAULT_GTID_SERVER_ID`, which is 0. We deployed many Maxwell instances in production, and they share the same metadata database. After the GTID mode got enabled, the newly created Maxwell instance will always getLatestFromAnyClient() by `server_id = 0` to get wrong GTID position, we have to insert into the `positions` table manually with the correct GTID every time.

Maybe it'd be better to not `getLatestFromAnyClient()` when the GTID mode is on?

Thanks.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -109,7 +109,7 @@ public class Maxwell implements Runnable {
 			   if so we have to start at that position or else
 			   we could miss schema changes, see https://github.com/zendesk/maxwell/issues/782 */
 
-			if ( initial == null ) {
+			if ( initial == null && !config.gtidMode) {
 				initial = this.context.getOtherClientPosition();
 				if ( initial != null ) {
 					LOGGER.info("Found previous client position: " + initial);


osheroff:
>  We deployed many Maxwell instances in production, and they share the same metadata database.

Just to confirm; these are many instances of maxwell (with different client_ids) pointing at the same cluster of servers, yes? 

> After investigation, I found the root cause is Maxwell took another database's GTID position after restart. 

Can you clarify what you mean when you say "another database"?  Is this a totally unrelated set of servers? 

The reason for this code, btw, is this: when you boot a new Maxwell client on a server cluster, it will share the maxwell schema with all other clients.   But we don't know if changes have been made to the schema since the last client stopped replicating, so the only safe things to do would be either (1) the clients all have to keep independent schemas, or (2) set the position of the new client to where old client left off.

Fleurer:
hi osheroff, thank you for your quick response!

> Just to confirm; these are many instances of maxwell (with different client_ids) pointing at the same cluster of servers, yes?

we deploy one maxwell instance for each database in the production, they are not the same mysql server. they'd have different client_id and different gtid set. these are the sample positions:

Screen Shot 2019-03-21 at 11 38 00 AM

all the `server_id`s are same as 0 since maxwell always take `DEFAULT_GTID_SERVER_ID` if the gtidMode is true, but they are different servers:

Screen Shot 2019-03-21 at 11 51 21 AM

If the gtidMode is true, the `server_id` field would not works as before: it could not identify the correct mysql server by the unique `server_id`. the query`SELECT * FROM positions WHERE server_id = ? ORDER BY last_heartbeat_read DESC limit 1` would results an irrelevant GTID position by `server_id = 0`.

osheroff:
oh, my.  

Yeah, what you're doing -- keeping maxwell metadata for unrelated servers in a single maxwell database --  is not an intended configuration.  It does work before you turn on GTID, because (as you noted) everyone is constrained with a unique SERVER_ID.  This configuration that you've got going is simply not going to work with GTID, maxwell will pick up the wrong schema for the server, all sorts of stuff will go horribly and comically wrong. 

if you want to do what you're doing with GTID mode (and all the niceness of being able to effortlessly transition to a new master (and thus new server_id)), we'd need to introduce some kind of abstraction that uniquely identifies a *cluster* of servers.  `client_id` is *almost* that, but it's not exactly, as it's assumed that each maxwell-client is still operating on the same cluster of servers, leading to programming decisions like this one (and like not keeping a client_id in schemas).

The easy fix here is to give each of your clusters its own maxwell database (on the same host is fine, just separate databases).   If we could figure out a more general fix I'd probably accept a PR around making this configuration work...

PR #1231, opened 2019-03-27
Fix fallback publishing for broker-side rejections

timbertson-zd:
The initial code was tested (and worked fine) for client-side rejections, like a message whose size exceeds `max.request.size`.

However, when the broker rejects a message, the exception comes back via a different path. This is harder to manually test, since all the easy ways of making large messages in MySQL compress well, and therefore won't trigger any size issues if you're using compression 🤦‍♂️ .

Specifically, the `KafkaCallback` will be invoked on `kafka-producer-network-thread`. It turns out that calling `send()` from kafka's own network thread is a Bad Plan, and will result in a `TimeoutException` "Failed to update metadata after <...> ms". I'm assuming it's stuck trying to reacquire some lock which it already holds.

This PR pushes the publish onto a background thread (a single threaded executor which is only used for this purpose, and only created the first time it's needed). This isn't particularly elegant, but I couldn't think of a better approach (@osheroff chime in if you have ideas!).

I considered pushing a message onto the worker's `queue`, but:
 - we'd need to combine RowMap with some other type representing a fallback message
 - I think that too can deadlock, since the queue is bounded and in some cases (client rejection) we'd block the thread which reads the queue with a push to the same queue.

We could push the executor up to MaxwellContext for general "do thing in background" support, but if nothing else uses it there's probably not much point yet.

/cc @zendesk/goanna

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
@@ -11,10 +11,7 @@ import com.zendesk.maxwell.row.RowMap.KeyFormat;
 import com.zendesk.maxwell.schema.ddl.DDLMap;
 import com.zendesk.maxwell.util.StoppableTask;
 import com.zendesk.maxwell.util.StoppableTaskState;
-import org.apache.kafka.clients.producer.Callback;
-import org.apache.kafka.clients.producer.KafkaProducer;
-import org.apache.kafka.clients.producer.ProducerRecord;
-import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.clients.producer.*;
 import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.errors.RecordTooLargeException;
 import org.apache.kafka.common.serialization.StringSerializer;
@@ -22,8 +19,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Properties;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
 
 class KafkaCallback implements Callback {
 	public static final Logger LOGGER = LoggerFactory.getLogger(MaxwellKafkaProducer.class);
@@ -138,7 +134,7 @@ public class MaxwellKafkaProducer extends AbstractProducer {
 class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnable, StoppableTask {
 	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellKafkaProducer.class);
 
-	private final KafkaProducer<String, String> kafka;
+	private final Producer<String, String> kafka;
 	private final String topic;
 	private final String ddlTopic;
 	private final MaxwellKafkaPartitioner partitioner;
@@ -149,6 +145,7 @@ class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnab
 	private Thread thread;
 	private StoppableTaskState taskState;
 	private String deadLetterTopic;
+	private ExecutorService backgroundExecutor;
 
 	public static MaxwellKafkaPartitioner makeDDLPartitioner(String partitionHashFunc, String partitionKey) {
 		if ( partitionKey.equals("table") ) {
@@ -158,7 +155,9 @@ class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnab
 		}
 	}
 
-	public MaxwellKafkaProducerWorker(MaxwellContext context, Properties kafkaProperties, String kafkaTopic, ArrayBlockingQueue<RowMap> queue) {
+	public MaxwellKafkaProducerWorker(MaxwellContext context, String kafkaTopic, ArrayBlockingQueue<RowMap> queue,
+		Producer<String,String> producer)
+	{
 		super(context);
 
 		if ( kafkaTopic == null ) {
@@ -168,7 +167,7 @@ class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnab
 		}
 
 		this.interpolateTopic = this.topic.contains("%{");
-		this.kafka = new KafkaProducer<>(kafkaProperties, new StringSerializer(), new StringSerializer());
+		this.kafka = producer;
 
 		String hash = context.getConfig().kafkaPartitionHash;
 		String partitionKey = context.getConfig().producerPartitionKey;
@@ -187,6 +186,13 @@ class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnab
 
 		this.queue = queue;
 		this.taskState = new StoppableTaskState("MaxwellKafkaProducerWorker");
+	}
+
+	public MaxwellKafkaProducerWorker(MaxwellContext context, Properties kafkaProperties, String kafkaTopic,
+		ArrayBlockingQueue<RowMap> queue)
+	{
+		this(context, kafkaTopic, queue,
+			new KafkaProducer<String,String>(kafkaProperties, new StringSerializer(), new StringSerializer()));
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -111,6 +109,7 @@ String getFallbackTopic() {
 	public MaxwellKafkaProducer(MaxwellContext context, Properties kafkaProperties, String kafkaTopic) {
 		super(context);
 		this.queue = new ArrayBlockingQueue<>(100);
+		BlockingQueue<Runnable> runQueue = new ArrayBlockingQueue<>(100);
```
lukestephenson-zendesk: Is `runQueue` used anywhere?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -231,18 +240,36 @@ public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) thro
...
+		return this.backgroundExecutor;
+	}
+
+	public void sendFallbackAsync(String topic, RowIdentity fallbackRecord, KafkaCallback callback, RecordMetadata md, Exception reason) {
+		// Note: this code may be executed from two different threads:
+		//  - client rejection: invoked synchronously from the  MaxwellKafkaProducerWorker thread (in sendAsync())
+		//  - broker rejection: invoked asynchronously from kafka's `kafka-producer-network-thread`
+		// We can't call `sendAsync` on kafka's own network thread, it causes a deadlock. So we
+		// invoke the send via an Executor with its own thread.
+		getBackgroundExecutor().submit(new Runnable() {
```
lukestephenson-zendesk: Can there be multiple concurrent calls to `sendFallbackAsync`?

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -231,18 +240,36 @@ public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) thro
...
 
-	public void sendFallbackAsync(String topic, RowIdentity fallbackRecord, KafkaCallback callback, RecordMetadata md, Exception reason) {
-		try {
-			ProducerRecord<String, String> record = makeFallbackRecord(topic, fallbackRecord, reason);
-			sendAsync(record, callback);
-		} catch (Exception fallbackEx) {
-			callback.onCompletion(md, fallbackEx);
+	private ExecutorService getBackgroundExecutor() {
+		if (this.backgroundExecutor == null) {
+			this.backgroundExecutor = Executors.newSingleThreadExecutor();
```
lukestephenson-zendesk: If possible it's good to give named thread factories to help with log debugging.  Not sure we have a named thread factory easily available from any of the dependencies.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: Ah nope, I was trying something out and forgot to back this out

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: Good point. Probably? It can be called from 2 threads as outlined, and I don't think there's any synchronization between them :/

Are you worried about `getBackgroundExecutor` making more than one pool due to races? I considered just using `ForkJoinPool.commonPool()`, which is easier but slightly cheekier if you're embedding maxwell in an app which uses that for its own purposes.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: I looked around for thread factories in scope (I like to `setDaemon(false)`), but there was nothing readily available, and I think the logs should make it reasonably clear.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
lukestephenson-zendesk: Was worried that the backgroundExecutor wasn't defined as `volatile` so could potentially break if two concurrent calls.  

More than 1 pool probably isn't too bad, I assume it would get GCed pretty quickly without a reference to it.

osheroff: believe that we can use lambas here for a tad of code cleanliness.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -136,7 +134,7 @@ public KafkaProducerDiagnostic getDiagnostic() {
 class MaxwellKafkaProducerWorker extends AbstractAsyncProducer implements Runnable, StoppableTask {
 	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellKafkaProducer.class);
 
-	private final KafkaProducer<String, String> kafka;
+	private final Producer<String, String> kafka;
```
osheroff: this change... are we sure this change is cool with older versions of kafka libraries?

timbertson-zd:
> throw it back into the queue could be a reasonably clean solution

@osheroff I agree, but as I mentioned in the description I'm pretty sure that approach can deadlock. It also delays fallback messages by pushing them to the end of the queue, which isn't ideal but probably acceptable.

What I'd really like is for the `queue` to still be blocking, but allow me to insert an item a the head (without blocking) in this one case. But I don't think there's a queue which works like that.

Unless... Does this `queue` actually need to be blocking? IIRC it's behind the inflightMessageList, which should do its own backpressure. Maybe we can just make this an unbounded queue, and then there's no risk of deadlock 🤔

osheroff:
> Unless... Does this queue actually need to be blocking? 

mmm.. yes, I think so...  otherwise when the producer stalls backpressures we won't backpressure the rest of the system and we'll just spool until we run out of memory.

However, it seems pretty reasonable to me to do a `queue.offer` with a tiny timeout, dropping the fallback on the floor in the unlikely event that we're completely jammed.

osheroff:
another option would be to add a special `fallbackMessageQueue` that the worker pulled from first and that never blocked.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: It's just an interface for the class, it's been there since 0.8 so I assume it's fine: https://kafka.apache.org/082/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html

timbertson-zd:
>  otherwise when the producer stalls backpressures we won't backpressure the rest of the system and we'll just spool until we run out of memory

Oh right, this queue sits in front of the inflight queue, not behind it :/

> another option would be to add a special fallbackMessageQueue that the worker pulled from first and that never blocked.

That seems pretty reasonable, I'll give it a go.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -177,6 +181,7 @@ public MaxwellKafkaProducerWorker(MaxwellContext context, Properties kafkaProper
 		this.ddlPartitioner = makeDDLPartitioner(hash, partitionKey);
 		this.ddlTopic =  context.getConfig().ddlKafkaTopic;
 		this.deadLetterTopic = context.getConfig().deadLetterTopic;
+		this.deadLetterQueue = this.deadLetterTopic == null ? null : new ConcurrentLinkedQueue<>();
```
osheroff: seems overly optimized, just init the thing.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -231,15 +253,18 @@ public void sendAsync(RowMap r, AbstractAsyncProducer.CallbackCompleter cc) thro
 
 		KafkaCallback callback = new KafkaCallback(cc, r.getNextPosition(), r.getRowIdentity(), value,
 				this.succeededMessageCount, this.failedMessageCount, this.succeededMessageMeter, this.failedMessageMeter,
-				this.deadLetterTopic, this.context, this);
+				this.topic, this.deadLetterTopic, this.context, this);
 
 		sendAsync(record, callback);
 	}
 
 	public void sendFallbackAsync(String topic, RowIdentity fallbackRecord, KafkaCallback callback, RecordMetadata md, Exception reason) {
```
osheroff: let's rename this `enqueueFallbackRow` or something.

src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
timbertson-zd: I don't care about the memory, this is just so we can skip `poll()`ing it if it's `null`. Otherwise we'll `poll()` an unused queue on every message when the feature is turned off. But if you're not worried, I'm good either way 🤷‍♂️

timbertson-zd:
> I'm wondering whether it makes sense to keep the original message marked outstanding (in InflightMessageList) until we complete the dead-letter message; shouldn't we just mark the original as having completed and move that queue along?

If you do that, then a shutdown at the wrong moment (or a transient kafka issue) would cause the dead letter to be dropped. This way we guarantee that _a_ message will be written somewhere. It costs us the added delay of a publish roundtrip, but as long as the inflight queue isn't at capacity, that won't actually affect the timing of any other message publish (they'll just get acked a bit later).

PR #1235, opened 2019-04-02
one more columns can be empty

Itisyou:
Producer_partition_columns supports multiple fields, as long as one exists.
We need a database-based column with a non-primary key to be processed uniformly, because some table field names are inconsistent and need to be processed compatibly.

### DIFF
--- a/config.properties.example
+++ b/config.properties.example
@@ -160,6 +160,8 @@ kafka.acks=1
 # specify what fields to partition by when using producer_partition_by=column
 # column separated list.
 #producer_partition_columns=id,foo,bar
+# & means id or uid ,if one of them is empty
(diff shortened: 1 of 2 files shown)


osheroff:
this is too specific, but maybe i could open it up so you can modify the partition key with the javascript filter?

Itisyou:
> this is too specific, but maybe i could open it up so you can modify the partition key with the javascript filter?

I've solved this problem by modifying the database fields. It's perfect that producer_partition_columns can support JavaSctipt expressions. If not, you can add this patch first and optimize it in the future

Itisyou:
great

osheroff:
solved by other PR

PR #1248, opened 2019-05-03
Added primary key output option

azhard:
- Updated google cloud dependancies
- Added output options to include list of primary key columns & list of primary keys in output of DML statements (updates and inserts)

### DIFF
--- a/docs/docs/dataformat.md
+++ b/docs/docs/dataformat.md
@@ -30,6 +30,8 @@ mysql> insert into test.e set m = 4.2341, c = now(3), comment = 'I am a creature
    "position":"master.000006:800911",
    "server_id":23042,
    "thread_id":108,
+   "primary_key": "[1]"
+   "primary_key_columns": "[id]"
    "data":{
       "id":1,
       "m":4.2341,
@@ -73,6 +75,16 @@ this field and `xid` to do so.  The data will look like:
 - row with `commit=true`, xid=142
 - row with no `commit`, xid=155
 - ...
+
+↳ `"primary_key":"[1]",`
+
+You only get this with --output_primary_key. List of values that make up the 
+primary key for this row.
+
+↳ `"primary_key_columns":"[id]",`
+
+You only get this with --output_primary_key_columns. List of columns that make 
+make up the primary key for this row.
 
 
 ### UPDATE
--- a/docs/docs/dataformat.md
+++ b/docs/docs/dataformat.md
@@ -30,8 +30,8 @@ mysql> insert into test.e set m = 4.2341, c = now(3), comment = 'I am a creature
    "position":"master.000006:800911",
    "server_id":23042,
    "thread_id":108,
-   "primary_key": "[1]"
-   "primary_key_columns": "[id]"
+   "primary_key": "[1, "2016-10-21 05:33:37.523000"]"
+   "primary_key_columns": "["id", "c"]"
    "data":{
       "id":1,
       "m":4.2341,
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -227,12 +227,17 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsub</artifactId>
-      <version>0.30.0-beta</version>
+      <version>1.66.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>23.6-jre</version>
```
osheroff: why did we pick up guava here?

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -184,6 +185,27 @@ public String toJSON(MaxwellOutputConfig outputConfig) throws Exception {
...
+		EncryptionContext encryptionContext = null;
+		if (outputConfig.encryptionEnabled()) {
+			encryptionContext = EncryptionContext.create(outputConfig.secretKey);
+		}
+
+		DataJsonGenerator dataWriter = outputConfig.encryptionMode == EncryptionMode.ENCRYPT_DATA
+			? json.getEncryptingGenerator()
+			: json.getPlaintextGenerator();
+
+		JsonGenerator dataGenerator = dataWriter.begin();
```
osheroff: this diff is weird, why did all this code move above there?

src/main/java/com/zendesk/maxwell/row/RowMap.java
```diff
@@ -1,5 +1,6 @@
 package com.zendesk.maxwell.row;
 
+import com.google.api.client.util.Lists;
```
osheroff: not a fan of this dependency

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
```diff
@@ -51,7 +52,7 @@ public void testGetRowIdentity() {
 
 		RowMap rowMap = new RowMap("insert", "MyDatabase", "MyTable", TIMESTAMP_MILLISECONDS, pKeys, POSITION);
 
-		rowMap.putData("id", "9001");
+		rowMap.putData("id", 9001);
```
osheroff: why should these tests change?  doesn't seem right to me...

osheroff:
overall this is a fine feature and implemented dilligently enough (thx for doing documentation the first time!)

I'm happy to merge it after code review issues are addressed.

pom.xml
azhard: Removed.

src/main/java/com/zendesk/maxwell/row/RowMap.java
azhard: We are using the dataWriter to write the primaryKeys and primaryKeyColumn objects but we need to write them before the excludeColumns logic happens. If we don't, any of the columns that make up the primary key that are also in the exlcudeColumns list will not be outputted.

src/main/java/com/zendesk/maxwell/row/RowMap.java
azhard: Removed.

src/test/java/com/zendesk/maxwell/row/RowMapTest.java
azhard: The test was changed to show how the "primary_key" is outputted when dealing with non string values. ie. "primary_key": [9001, "foo"] vs. "primary_key": ["9001", "foo"].

azhard:
@osheroff addressed all your comments, let me know if anything else needs fixing!

docs/docs/dataformat.md
```diff
@@ -30,6 +30,8 @@ mysql> insert into test.e set m = 4.2341, c = now(3), comment = 'I am a creature
    "position":"master.000006:800911",
    "server_id":23042,
    "thread_id":108,
+   "primary_key": "[1, "2016-10-21 05:33:37.523000"]"
```
osheroff: these two aren't quite right but I'll fix after merge.

osheroff:
nice work.  merging.

PR #1253, opened 2019-05-20
Filter column names fix

kmehomiyski:
Fix column based filters (either by column name or by column value) by using the original column names, instead of lower case version, when getting table's column list from TableColumnList class. The column list is used in Filter.couldIncludeFromColumnFilters() method and only there, which forces the method to always use lower case version of the column. On the other hand, the Filter.includes() method uses the original column names as reported by "RowMap.getData()" method. This leads to inconsistent syntex requirement for the filters, which in turn brakes column based filtering.


osheroff:
Hi.  

Column names are not case sensitive anywhere in mysql, so I think a better fix here would be to lower-case the column names whenever we're using them in the filter.

kmehomiyski:
Hi,

OK I will into it today and change the MR. Thanks for the feedback!

Issue #1245, opened 2019-04-30
When database add a new table maxwell crash with Couldn't find table exception


zengyongping:
When MySQL database add a new table, maxwell process crash with Couldn't find table exception.
#######################logs below######################
15:00:17,988 INFO  BinlogConnectorLifecycleListener - Binlog disconnected.
15:00:17,990 INFO  TaskManager - Stopping 4 tasks
15:00:17,990 ERROR TaskManager - cause:
**java.lang.RuntimeException: Couldn't find table hj_new_model in database XXXXXX**
        at com.zendesk.maxwell.replication.TableCache.processEvent(TableCache.java:33) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getTransactionRows(BinlogConnectorReplicator.java:462) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.getRow(BinlogConnectorReplicator.java:568) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.replication.BinlogConnectorReplicator.work(BinlogConnectorReplicator.java:174) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.util.RunLoopProcess.runLoop(RunLoopProcess.java:27) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.Maxwell.startInner(Maxwell.java:224) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.Maxwell.start(Maxwell.java:156) ~[maxwell-1.20.0.jar:1.20.0]
        at com.zendesk.maxwell.Maxwell.main(Maxwell.java:245) ~[maxwell-1.20.0.jar:1.20.0]

OpenOpened:
Please check whether the hj_new_model table is in your maxwell database schemas table?

zengyongping:
> Please check whether the hj_new_model table is in your maxwell database schemas table?

the hj_new_model table not in tables、schemas table，I manual insert table information about hj_new_model.But How maxwell can automatic capture new add tables metadata info?

OpenOpened:
The information in the schame table comes from the data generated by binlog, make sure maxwell can capture binlog build statements or schema changes.

zengyongping:
> generated



[...]

PR #1265, opened 2019-05-31
Fix potential memory leak caused by blacklist table cache

kevinclcn:
Maxwell often crashes when temporary tables are created, see this issue: https://github.com/zendesk/maxwell/issues/1245. I want to use blacklist filter to ignore the dynamic created temp tabled. This change is to fix a potential memory leak caused by blacklist table cache. The blacklisted tables would only be added to the cache but without being removed. Since the blacklist table cache is used in no where, I removed it here.


osheroff:
No, I won't accept this, but I would accept a patch that makes this into an LRU cache of one thousand or ten thousand tables....

kevinclcn:
@osheroff Sure, I can add a LRU cache here. But just for curious, I don't find the blacklistedTableCache is used in anywhere, why can't I remove it?

osheroff:
ah, no, good point.  It's been awhile since this code was used.

Issue #1274, opened 2019-06-19
Latest Docker version v1.22.2 fails with pubsub as producer


DPGrev:
When running maxwell with pubsub as a producer using the latest docker version: v1.22.2 I get the following error:

```
Exception in thread "main" java.lang.NoSuchMethodError: com.google.common.base.Preconditions.checkArgument(ZLjava/lang/String;CLjava/lang/Object;)V
	at io.grpc.Metadata$Key.validateName(Metadata.java:630)
	at io.grpc.Metadata$Key.<init>(Metadata.java:638)
	at io.grpc.Metadata$Key.<init>(Metadata.java:568)
	at io.grpc.Metadata$AsciiKey.<init>(Metadata.java:743)
	at io.grpc.Metadata$AsciiKey.<init>(Metadata.java:738)
	at io.grpc.Metadata$Key.of(Metadata.java:594)
	at io.grpc.Metadata$Key.of(Metadata.java:590)
	at com.google.api.gax.grpc.GrpcHeaderInterceptor.<init>(GrpcHeaderInterceptor.java:61)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.createSingleChannel(InstantiatingGrpcChannelProvider.java:180)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.createChannel(InstantiatingGrpcChannelProvider.java:164)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.getTransportChannel(InstantiatingGrpcChannelProvider.java:156)
	at com.google.api.gax.rpc.ClientContext.create(ClientContext.java:157)
	at com.google.cloud.pubsub.v1.stub.GrpcPublisherStub.create(GrpcPublisherStub.java:161)
	at com.google.cloud.pubsub.v1.Publisher.<init>(Publisher.java:152)
	at com.google.cloud.pubsub.v1.Publisher.<init>(Publisher.java:79)
	at com.google.cloud.pubsub.v1.Publisher$Builder.build(Publisher.java:637)
	at com.zendesk.maxwell.producer.MaxwellPubsubProducerWorker.<init>(MaxwellPubsubProducer.java:138)
	at com.zendesk.maxwell.producer.MaxwellPubsubProducer.<init>(MaxwellPubsubProducer.java:97)
	at com.zendesk.maxwell.MaxwellContext.getProducer(MaxwellContext.java:333)
[...]

PR #1275, opened 2019-06-19
Added guava to pom file

azhard:
The error seen in issue #1274 is due to guava needing to be included in the project's pom file.


osheroff:
have you tested the fix?

azhard:
Yes it worked for me, in my initial MR I had this as well but removed it after one of your comments because I forgot we needed it to fix this bug.

PR #1302, opened 2019-07-30
Add row type (delete/insert/update) to fallback error message

ThuTrinh:
@zendesk/goanna 
Adding row type change to the fallback message.

JIRA: https://zendesk.atlassian.net/browse/MELINF-982

Screen Shot 2019-07-30 at 3 18 56 pm

### DIFF
--- a/src/main/java/com/zendesk/maxwell/row/RowIdentity.java
+++ b/src/main/java/com/zendesk/maxwell/row/RowIdentity.java
@@ -8,9 +8,9 @@ import java.io.Serializable;
 import java.util.*;
 
 public class RowIdentity implements Serializable {
-	private String database;
-	private String table;
-	private String rowType;
+	private final String database;
+	private final String table;
+	private final String rowType;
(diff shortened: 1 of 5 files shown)


src/main/java/com/zendesk/maxwell/row/RowIdentity.java
```diff
@@ -10,12 +10,14 @@
 public class RowIdentity implements Serializable {
 	private String database;
 	private String table;
+	private String rowType;
```
lukestephenson-zendesk: can this be tagged `final`?

ThuTrinh: just checked usages, all three can :)

lukestephenson-zendesk:
This looks OK, but the tests are failing.  Not sure if it's related to this change.

ThuTrinh:
Looks like the test failures are due to a separate issue related to bootstrapping. I can see similar tests failing in other PRs.

PR #1308, opened 2019-08-06
Remove race condition from bootstrapper integration tests

ThuTrinh:
Previously the MaxwellTestSupport would interrupt the thread running Maxwell as an attempt to speed up the tests.
Unfortunately the all bootstrap tasks don't always complete before the threat is interrupted causing non deterministic results.

Now, we don't interrupt the test. Instead we introduce a way to detect that the bootstrap tasks have all been completed, before performing the asserts.

@zendesk/goanna @osheroff

### DIFF
--- a/src/main/java/com/zendesk/maxwell/bootstrap/BootstrapController.java
+++ b/src/main/java/com/zendesk/maxwell/bootstrap/BootstrapController.java
@@ -4,7 +4,6 @@ import com.zendesk.maxwell.producer.AbstractProducer;
 import com.zendesk.maxwell.row.RowMap;
 import com.zendesk.maxwell.row.RowMapBuffer;
 import com.zendesk.maxwell.util.RunLoopProcess;
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import snaq.db.ConnectionPool;
--- a/src/main/java/com/zendesk/maxwell/bootstrap/BootstrapController.java
+++ b/src/main/java/com/zendesk/maxwell/bootstrap/BootstrapController.java
@@ -4,6 +4,7 @@ import com.zendesk.maxwell.producer.AbstractProducer;
 import com.zendesk.maxwell.row.RowMap;
 import com.zendesk.maxwell.row.RowMapBuffer;
 import com.zendesk.maxwell.util.RunLoopProcess;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import snaq.db.ConnectionPool;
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/bootstrap/BootstrapController.java
```diff
@@ -4,6 +4,7 @@
 import com.zendesk.maxwell.row.RowMap;
 import com.zendesk.maxwell.row.RowMapBuffer;
 import com.zendesk.maxwell.util.RunLoopProcess;
+import org.apache.commons.lang3.StringUtils;
```
lukestephenson-zendesk: nit. Doesn't look like the new import is used

src/test/java/com/zendesk/maxwell/MaxwellTestSupport.java
```diff
@@ -232,30 +234,46 @@ public void run() {
...
 
-					if ( r.toJSON(config.outputConfig) != null )
-						list.add(r);
+			if (bootstrapComplete && replicationComplete) {
+				// For sanity testing, we wait another 2s to verify that no additional changes were pending.
+				// This slows down the test suite, so only enable when debugging.
+				boolean checkHasNoPendingChanges = false;
+				if (checkHasNoPendingChanges) {
+					long deadline = System.currentTimeMillis() + 2000;
+					while(true) {
```
lukestephenson-zendesk: could this just be `while(System.currentTimeMillis() <= deadline)`?  or a `do /while`?

src/main/java/com/zendesk/maxwell/bootstrap/BootstrapController.java
```diff
@@ -83,6 +84,21 @@ public synchronized void setCurrentSchemaID(long schemaID) {
...
+		try ( Connection cx = maxwellConnectionPool.getConnection() ) {
+			PreparedStatement s = cx.prepareStatement("select count(id) from bootstrap where is_complete = 0 and client_id = ?");
+			s.setString(1, this.clientID);
+
+			ResultSet rs = s.executeQuery();
+
+			while (rs.next()) {
+				return rs.getInt(1) > 0;
+			}
+		} catch (Exception ex){}
```
lukestephenson-zendesk: Is it possible for this code to live in `src/test/java` since it is purely for test purposes?  Or if it is intended for wider use and needs to live in `src/main/java`, perhaps the details of the swallowed Exception should be logged?

osheroff:
Looks good, thanks! :+1

src/test/java/com/zendesk/maxwell/MaxwellTestSupport.java
```diff
@@ -221,7 +221,7 @@ public void run() {
 		}
 
 		callback.afterReplicatorStart(mysql);
-		maxwell.context.getBootstrapController(null).interrupt();
+		BootstrapController bootstrapController = maxwell.context.getBootstrapController(null);
```
lukestephenson-zendesk: `bootstrapController` is never used.  Can this call be removed (or at least the variable assignment)?

src/test/java/com/zendesk/maxwell/MaxwellTestSupport.java
```diff
@@ -305,4 +324,19 @@ public static void requireMinimumVersion(MysqlIsolatedServer server, MysqlVersio
 		// skips this test if running an older MYSQL version
 		assumeTrue(server.getVersion().atLeast(minimum));
 	}
+
+	private static int getIncompleteBootstrapTaskCount(Maxwell maxwell, String clientID) throws SQLException {
+		try ( Connection cx = maxwell.context.getMaxwellConnection() ) {
```
lukestephenson-zendesk: This `try` block caught an Exception before it was moved into the tests.  Was that required?

If no exceptions are going to be caught, I don't think there is a need for the `try` (my java is very rusty though).

timbertson-zd: Still useful, it's a try-dispose: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

lukestephenson-zendesk: Thanks.  That's a new java addition since my java days.

PR #1310, opened 2019-08-12
Increase docker mvn version to maven:3.6-jdk-8

DPGrev:
The Maven:3.5-jdk-8 image does not get updated anymore. This will increase the java 8 version to 8u222. This introduces better functionality for running java in docker containers. 

I.e one can set the memory used by the docker container by setting the `-XX:+UseContainerSupport` and `-XX:MaxRAMPercentage`, `-XX:InitialRAMPercentage` `-XX:MinRAMPercentage`.


osheroff:
thanks!

PR #1315, opened 2019-08-16
Spanner support

adkhare:
We at Mailchimp are wanting some feature around this. Purpose is to have a test data replicated over to Spanner so that we can use it for future


osheroff:
It's really cool that you guys are doing this.  maxwell -> something else that I didn't ever anticipate is great.  However, I'm not sure if doing a maxwell producer around this is correct:

- how would one guarantee that the spanner schema is correctly synchronized with the maxwell schema?  
- what does it do if we find an out of sync schema?  
- maxwell really only provides at-least-once guarantees, duplicate rows here are a possible issue. 
- is one-row-at-a-time going to be fast enough or would we really need some sort of batching?

i mean i guess if the attitude is "whatever.  the code works, let it crash" I'm sort of ok with that.  I can't really see wholeheartedly recommending folks do this this way though, seems like what you really want is for maxwell to output to an intermediate buffer and then do this in application logic.

PR #1318, opened 2019-08-16
Add buffer memory configuration flag.

DPGrev:
This pr will make to memory size used by the buffered list configurable to more finely tune the memory usage of maxwell.


osheroff:
Hi there!  sorry for the delay.  

This PR is generally fine.  You can remove the bits where you pass this flag through `Recovery` -- the recovery code doesn't really output any rows so it doesn't buffer any of them to disk.

DPGrev:
@osheroff I have removed the flags from recovery.

DPGrev:
@osheroff Were you able to look at the changes?

PR #1314, opened 2019-08-14
Upgrade jackson-databind to 2.9.9.2

carrotpaul:
Hi there! I've been running Maxwell in production, and my vulnerability scanner found a critical vulnerability ([CVE-2019-14379](https://nvd.nist.gov/vuln/detail/CVE-2019-14379)) against the `jackson-databind` version Maxwell is currently using.

This PR bumps the version to 2.9.9.2, which contains a patch for that vulnerability.

PR #1319, opened 2019-08-21
Bump up jackson-databind version  to 2.9.9.3

anitanaseri:
Bump up com.fasterxml.jackson.core:jackson-databind version in maxwell to 2.9.9.3 for security patch


anitanaseri:
Close due to #1314

PR #1322, opened 2019-08-26
Added release instructions and fixed release script

ThuTrinh:
cc @zendesk/goanna @osheroff 

## Description
- Updated the release script so that the nexus part works when running the script from the build folder
- Added some release instructions to help the next person out


build/README.md
```diff
@@ -0,0 +1,97 @@
...
+    login github-username
+    password github-token
+```
+
+## Releasing a version
+To release a new version simply cd into the build folder and run the release script. This script will automatically bump up the version, tag and release to maven. 
+
+```shell
+cd ./build
+bundle exec ./release
```
timbertson-zd: It needs `<the-new-version-number>` as an argument, too

build/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Note: this script is not idempotent so if any errors occure, you will need to run each step manually. The release script takes the step as an option.
```
timbertson-zd:
```suggestion
Note: this script is not idempotent so if any errors occur, you will need to run each step manually. The release script takes the step as an option.
```

PR #1337, opened 2019-09-12
Upgrade mysql-connector-java to 8.0.17

lumixen:
Hola!
As we decided to use maxwell as an embedded library together with spring boot 2 we've faced with conflicting dependencies: spring boot data uses mysql connection 8.X+ while maxwell still relies on 5.X. As a result - `ClassNotFoundException` because 8.X+ driver deprecated and removed `MySQLIntegrityConstraintViolationException`.
This small PR just upgrades mysql driver to `8.0.17` and replaces deprecated exception with one that comes from the `java.sql` package.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <dependency>
       <groupId>mysql</groupId>
       <artifactId>mysql-connector-java</artifactId>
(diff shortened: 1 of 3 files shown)


osheroff:
thanks!

PR #1342, opened 2019-09-23
Add pubsub retry and batch config

DPGrev:
This PR adds two new functions to the pubsub producer. 

**Retry Config:**
As stated by google, delivering pubsub messages might fail. We noticed that in the extreme cases the default values for retrying failed and thus the message was not delivered. Thus we would like to set custom values for the retry config. The default values for these configs are set to the same default values as set by google. See [Retry settings pubsub docs](https://cloud.google.com/pubsub/docs/publisher#retry)

**Batch Config:**
Pubsub supports batching messages together this increases throughput however it also increases latency. In order to increase the throughput of pubsub we would like this to configurable. Default value are set to the same default values as set by google. See [Batchn settings pubsub docs](https://cloud.google.com/pubsub/docs/publisher#batching)


osheroff:
thanks for the PR!

I believe Maxwell's default command-line options are getting far too long, `--help` is pages and pages long.  I know it's outside of the scope of this PR, but what do you think the right thing to do is?  Should we:

- limit some of the options to only be in config.properties?
- hide some of the options behind like `--help pubsub` or `--help kafka`?
- something else?

DPGrev:
If only certain options are available in the config.properties it would mean that when people use docker to run maxwell they would need to modify the DockerFile to load in the config.properties, or it should be mounted within a volume. Thus this would introduce an extra step in running maxwell.

I like the `--help pubsub` and `--help kafka` option since we could use the headers as described in the docs for this. For example:

- Mysql options
- Producer options
- Kafka
- Etc etc

I think that this is a decent step in improving the docs.

osheroff:
sounds good.  Thanks for the PR, merging.

Issue #1307, opened 2019-08-03
Redis Producer Options


acrois:
Currently:
LPUSH
PUBLISH

Future:
XADD - As of July 23, [Jedis officially supports Redis streams](https://github.com/xetorthio/jedis/releases/tag/jedis-3.1.0). This means bumping dependency for Jedis to 3.1.0 from 2.9.0

Additional:
RPUSH - Same operation as LPUSH, different end of the list

I will provide PR after some time.

osheroff:
PR merged

PR #1348, opened 2019-09-29
Redis Producer Options #1307

acrois:
1. Bumping version of Jedis from 2.9.0 to 3.1.0.
2. Adds support for streams through **redis_type = xadd**
3. Adds configuration option **redis_stream_key** for the stream name
4. Adds configuration option **redis_stream_json_key** for the message field name
5. Treats **redis_stream_key**, **redis_pubsub_channel**, and **redis_list_key** as "channel" internally, based on **redis_type**.
6. Adds documentation for the new options
7. Creates a new, simplified constructor, in favor of reading configuration options from the **MaxwellContext** that is passed into **MaxwellRedisProducer**.
8. Deprecates old constructor for backwards compatibility. **No functionality has changed**
9. Delegates control of serialization to sendToRedis() by changing signature from **sendToRedis(String msg)** to **sendToRedis(RowMap msg)**
10. Moves the debug code into sendToRedis, does not change behavior

Note that stream messages are encoded as 1D arrays but interpreted as K(even)/V(odd) pairs in many client implementations in accordance with the Redis official documentation, for at least some basic support, all it needs to do is set one field with the message JSON. In the future it would be **nice to have** additional support for flattened, dot-notated object K/V mapped message bodies.

Addresses issue #1307

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
@@ -18,7 +18,7 @@ public class MaxwellRedisProducer extends AbstractProducer implements StoppableT
 	private final Jedis jedis;
 
 	@Deprecated
-    public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
+	public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
 		this(context, redisType);
 	}
 
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
@@ -6,42 +6,98 @@ import com.zendesk.maxwell.util.StoppableTask;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.StreamEntryID;
 import redis.clients.jedis.exceptions.JedisConnectionException;
+import java.util.HashMap;
+import java.util.Map;
 
 public class MaxwellRedisProducer extends AbstractProducer implements StoppableTask {
 	private static final Logger logger = LoggerFactory.getLogger(MaxwellRedisProducer.class);
 	private final String channel;
-	private final String listkey;
-	private final String redistype;
+	private final String redisType;
 	private final Jedis jedis;
 
-	public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
+	@Deprecated
+    public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
+		this(context, redisType);
+	}
+
+	public MaxwellRedisProducer(MaxwellContext context, String redisType) {
 		super(context);
 
-		channel = redisPubChannel;
-		listkey = redisListKey;
-		redistype = redisType;
+		if (this.context.getConfig().redisListKey != null) {
+			channel = context.getConfig().redisListKey;
+		}
+		else if (this.context.getConfig().redisStreamKey != null) {
+			channel = context.getConfig().redisStreamKey;
+		}
+		else {
+			channel = this.context.getConfig().redisPubChannel;
+		}
+
+		this.redisType = redisType;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -6,42 +6,98 @@
...
 	private static final Logger logger = LoggerFactory.getLogger(MaxwellRedisProducer.class);
 	private final String channel;
-	private final String listkey;
-	private final String redistype;
+	private final String redisType;
 	private final Jedis jedis;
 
-	public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
+	@Deprecated
+    public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
```
osheroff: indentation off here.

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -6,42 +6,98 @@
...
+					message.put(jsonKey, messageStr);
+				}
+
+				// TODO timestamp resolution coercion
+				// 			Seconds or milliseconds, never mixing precision
+				//      	DML events will natively emit millisecond precision timestamps
+				//      	CDC events will natively emit second precision timestamp
+				// TODO configuration option for if we want the msg timestamp to become the message ID
+				//			Requires completion of previous TODO
+				jedis.xadd(this.channel, new StreamEntryID(), message);
```
osheroff: it's strange to me that Jedis forces you to generate your own `StreamEntryID` -- according to https://redis.io/commands/xadd the vastly preferred method of ID-ing these things is to just pass "*" to redis and let it generate its own id.  In particular, I'd worry about:

> When a user specified an explicit ID to XADD,  ... the user must specify an ID which is greater than any other ID currently inside the stream, otherwise the command will fail. 

which means that two producers sending to the same stream can and will screw each other up. 

...

Ah.  Ok.  further source code reading indicates that you should be passing `StreamEntryID.NEW_ENTRY` instead of what you're doing.

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -6,42 +6,98 @@
...
+				Map<String, String> message = new HashMap<>();
+
+				String jsonKey = this.context.getConfig().redisStreamJsonKey;
+				
+				if (jsonKey == null) {
+					// TODO dot notated map impl in RowMap.toJson
+					throw new IllegalArgumentException("Stream requires key name for serialized JSON value");
+				}
+				else {
+					message.put(jsonKey, messageStr);
```
osheroff: just curious; is there a reason we're passing the whole thing as a JSON blob rather than using the internal redis KV store that's in the message?

acrois: Since you're curious-

I have a use case where the JSON body is the value part of the first K/V pair and nothing more. So I'll justify why support JSON body as first K/V pair instead of flattened K/V pair from a Maxwell integrator POV.

In practice, an integration with a Stream or Queue will introduce an interface that eliminates the concern of message source to message consumers. What if I wrote code ready for the case: No Redis. Yes Kafka?

For example, porting an implementation of Maxwell messages from list (used as queue) to stream is almost as simple as replacing a few Redis commands since a JSON body is to be read either way.

Flatting creates the same result. In production already, it should be a backwards-compatible feature as K/V (dot-notated?) flat mapping will [preferably] obsolete this practice. It requires implementing expansion and contraction on both the message's touch points: Maxwell and the Consumer via dot-notated keys, which is not (as of writing) implemented in client libraries anyway. This implementation path is most likely, and should be supported first to obtain the benefits of Redis streams.

Myself, I have yet to check the performance differences between flattened K/V and "json(value[0])" - so I know the performance of the decoding of JSON but not the impact on performance of the underlying data storage both redis, and protocol-compatible, disk-based K/V store like WiredTiger or RocksDB (implemented through ardb).


All that being said, this can be supported with little maintenance.

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
acrois: Thanks for pointing that out. Would have produced:
ERR The ID specified in XADD is equal or smaller than the target stream top item

PR #1350, opened 2019-10-03
Add support for age slo metrics

ThuTrinh:
Add support for setting message age slo threshold. When the message age exceeds this threshold the slo violation metric will be recorded.

🍐 @kayison

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
@@ -1,6 +1,7 @@
 package com.zendesk.maxwell.producer;
 
 import com.codahale.metrics.Gauge;
+import com.zendesk.maxwell.MaxwellConfig;
 import com.zendesk.maxwell.MaxwellContext;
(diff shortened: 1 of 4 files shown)


ThuTrinh:
@lukestephenson-zendesk we did look into it and it wasn't an easy task. We spoke to @timbertson and decided it wasn't worth the trouble.

src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
```diff
@@ -33,8 +36,14 @@ public void markCompleted() {
 				if (message != null) {
 					context.setPosition(message.position);
 					long currentTime = System.currentTimeMillis();
-					messagePublishTimer.update(currentTime - message.sendTimeMS, TimeUnit.MILLISECONDS);
-					messageLatencyTimer.update(Math.max(0L, currentTime - message.eventTimeMS - 500L), TimeUnit.MILLISECONDS);
+					long age = currentTime - message.sendTimeMS;
```
omgbob: Nitpick: would it be better to name this `messageAge` instead?

PR #1365, opened 2019-10-29
Avoid maxwell crash caused by non visible tables

kevinclcn:
Since permissions of Alicloud RDS are controlled strictly, Maxwell cannot be granted full permissions. It frequently crashes when we add new user or new table to another schema of the RDS instance.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/TableCache.java
+++ b/src/main/java/com/zendesk/maxwell/replication/TableCache.java
@@ -1,13 +1,16 @@
 package com.zendesk.maxwell.replication;
 
-import java.util.HashMap;
-
 import com.zendesk.maxwell.filtering.Filter;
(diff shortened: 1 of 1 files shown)


kevinclcn:
Hi @osheroff , not sure why the UT is not passed on MySQL 8.0, can you help on it?

osheroff:
the correct solution here is to use a blacklist instead (if possible).  As to your second question, I just don't understand it.

kevinclcn:
@osheroff Sorry for late replay. We usually create several schemas in the same ali-rds instance and the schemas and tables are created dynamically, so the blacklist does not work.

My second questions is that my code cannot pass against the unit tests in mysql 8.0.  I don't know why. see the link https://travis-ci.org/zendesk/maxwell/jobs/604247709

Issue #1362, opened 2019-10-23
redis_pub_channel doesn't have effect


KimPeeters:
When changing the redis_pub_channel all the messages are still published to "maxwell"

I've tested it with v1.23.2 and the latest master. Both don't seem to change the channel for me.

```
bin/maxwell --jdbc_options "serverTimezone=Europe/Amsterdam" --host xxxx --user maxwell --password xxxx --producer redis --redis_host 10.0.254.1 --redis_pub_channel testting
```
Changing redis_host is does work.

Any help on this would be appreciated.

osheroff:
i took a look here, the logic in this code in super funky/not too smart.  as a workaround you can just set `redis_list_key`, I'll fix this next release.

KimPeeters:
Yeah, noticed that yesterday too. Also saw that rpush is documented but wasn't implemented.

acrois:
Honestly, having worked in the Redis producer, it needs to be reimplemented and have tests if we're going to survive in production.

I would call the version in master: "Just good enough to do what we want".

elmehalawi:
FYI this also happens with `redis_type=xadd`. Basically `redis_stream_key` is ignored, and I needed to also set `redis_list_key`.

osheroff:
cleaned up these params a few releases ago.

PR #1366, opened 2019-10-29
Fixing redis commandline params

KimPeeters:
See also #1362. Added rpush, combined `--redis_pub_channel`, `--redis_list_key` and `--redis_stream_key` to `--redis_key`.


osheroff:
thanks for the PR, Kim.  Any chance I could get you to deprecate the old command line parameters instead of removing them in one fell swoop?

KimPeeters:
Sure, will look into it.

PR #1372, opened 2019-11-07
Fix pubsubRpcTimeoutMultiplier check

DPGrev:
The pubsub settings should be optional since defaults are already set.


osheroff:
the multiplier check makes sense but if people are passing these arguments, they need a value...  unless I'm missing something...

DPGrev:
@osheroff You are correct, apologies.

osheroff:
cool.

Issue #1251, opened 2019-05-19
Add a field to the Bootstrap table to expose the purpose of Bootstrap to the consumer.


tkaven:
Add a field to the Bootstrap table to expose the purpose of Bootstrap to the consumer.

Add this key-value pair at the top of the serialized json

In this way, consumers can do relevant processing according to this field

Very practical a function, can achieve it, thank you very much!!

tkaven:
Could you please help to implement this function? Thank you so much!

osheroff:
no, but I'd accept a PR (adding a `comment` field to `maxwell.bootstraps`)

tkaven:
> no, but I'd accept a PR (adding a `comment` field to `maxwell.bootstraps`)

ok, ill try, thank you..

I really need this feature because my clients consume maxwells data directly, and without this field the consumption logic of multiple consumers cannot be distinguished

tomcollinsproject:
I could really use this feature as a way of determining which process kicked off a bootstrap and whether to process the events or not. I'd like to submit a PR. I assume this should be added to the `RowMap` by the `SynchronousBootstrapper`. Would you have any issues with it being a first class property of `RowMap` (not in something like `extraAttributes`)? I use a custom RabbitMQ producer, so once exposed to the `RowMap`, I can take it from there. This property could be added to the various producers over time, however appropriate for that producer type.

Also, while I'm in there, would you have any issue with me changing the `bootstrap.where_clause` from a `varchar(255)` to a `text` type? I've needed that for a bit now and I'm pretty sure I've seen that request from others here.

Thanks!

osheroff:
> a "comment" wouldn't quite represent what I'm doing here, but eh.

you can just call it something opaque like `data` or `info` for all i care.

> Would you have any issues with it being a first class property of RowMap (not in something like extraAttributes)? I use a custom RabbitMQ producer, so once exposed to the RowMap, I can take it from there. 

I'm fine with it being first-class, but please make it a default part of the row serialization so that it's available to all producers.

> Also, while I'm in there, would you have any issue with me changing the bootstrap.where_clause from a varchar(255) to a text type? I've needed that for a bit now and I'm pretty sure I've seen that request from others here.

go right ahead.

PR #1373, opened 2019-11-11
Added bootstrap comment field.

tomcollinsproject:
Added a comment field to bootstrap records. Also, changed the bootstrap where_clause field from a varchar(255) to a text type. Reference: https://github.com/zendesk/maxwell/issues/1251

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ config/*.yml
 .classpath
 .project
 .settings/
+.factorypath
+.sts4-cache/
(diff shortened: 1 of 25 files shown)


osheroff:
cool, mostly good.  couple of things;

1.  need to upgrade existing installations by altering tables to have the `text` column and to include the `comment` field.  See `SchemaStoreSchema#upgradeSchemaStoreSchema`.

2. Let's write a test that tests these code paths.

tomcollinsproject:
Ugh, sorry for missing how you manage schema migrations. It was right there in the file I was looking at, too 🤦‍♂ . As for tests, again, sorry if I missed something obvious, but is there a trick to getting the integration tests working locally? I see reference to a `onetimeserver` file in `MysqlIsolatedServer`. Maybe I'm missing some bit of config..?

PR #1375, opened 2019-11-12
Added dynamic channel support for Redis.

FanAs:
It is similar behaviour that Kafka topics have.
It can be used, for example, for implementing reliable queues that will be parallel instead of putting all changes to single redis list.


osheroff:
looks good, thanks for the pr.

PR #1388, opened 2019-11-21
Add schema id to bootstrap-start and bootstrap-complete

DPGrev:
It would be nice to have the `schema_id` added to the `bootstrap-start` and `bootstrap-complete` messages.

- This would increase the consistency of the bootstrap messages since when `output_schema=true` is set every bootstrap event will contain the `schema_id`
- Having the `schema_id` in `bootstrap-start` and `bootstrap-complete` would add more identifiable information. For example it would help with the following use case:

If a schema change happens on the mysql instance and the messages are ingested into a data warehouse that does not support changing schemas one can create a new table and bootstrap the whole table again.  If two schema changes to the same table happen consecutively once can use the `schema_id` in `bootstrap-complete` to identify which bootstrap for which schema change finished.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
+++ b/src/main/java/com/zendesk/maxwell/bootstrap/SynchronousBootstrapper.java
@@ -37,7 +37,7 @@ public class SynchronousBootstrapper {
 
 	public void startBootstrap(BootstrapTask task, AbstractProducer producer, Long currentSchemaID) throws Exception {
 		performBootstrap(task, producer, currentSchemaID);
-		completeBootstrap(task, producer);
+		completeBootstrap(task, producer, currentSchemaID);
 	}
 
 	private Schema captureSchemaForBootstrap(BootstrapTask task) throws SQLException {
@@ -55,7 +55,7 @@ public class SynchronousBootstrapper {
 		Database database = findDatabase(schema, task.database);
 		Table table = findTable(task.table, database);
 
-		producer.push(bootstrapStartRowMap(table));
+		producer.push(bootstrapStartRowMap(table, currentSchemaID));
 		LOGGER.info(String.format("bootstrapping started for %s.%s", task.database, task.table));
 
 		try ( Connection connection = getMaxwellConnection();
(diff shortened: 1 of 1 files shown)


osheroff:
so I think maybe at one point in time `schema_id` existed in the bootstrap, but I took it out a little while ago; because bootstrapping is now a completely separate process from replication, an odd situation could occur:

- maxwell is lagged by (let's say) 2 days.  The rows its processing are at `schema_id = 10` 
- a bootstrap is requested
- at this point, the bootstrap is doing a `SELECT *` from the current table; the schema here is ahead of the replication thread and has DDL changes.

so in that world it'd be fairly misleading to say it had `schema_id == 10`.  I'm not sure if I'm being overly paranoid here, but I hate giving out misleading or inaccurate information like this.

DPGrev:
You are right. This would indeed result in the wrong `schema_id`. The scenario that I just described, would fail. This is indeed something to handle on the consumer end. 

However, as of the current version of maxwell the `schema_id` is still included in the `bootstrap-insert` events. Hence, the reason for this PR. Would you be willing to reconsider including the `schema_id` for now until a better solution has been found.

For example, whilst I think json is a good message format it is schemaless. We could look into adding support for the avro message format. Which does include the schema and thus there would be no need for the consumer to track the `schema_id` at all. Another option would be to add a hash of the mysql schema or add the types for the columns to the event. The latter option, however, will increase the event size quite a lot.

osheroff:
Avro has been talked about.  Someone maybe tried a PR at some point.  It’d be a thing where there would need to be a separate stream with just the Avro schemas, and a consumer would have to read that all the way through.  Or some solution to retrieve the schema.  I could never figure out a really elegant way to do it, so always settled for worse-is-better JSON.

Given that we’re already handing out dubious schema_ids, I guess it’s fine to put it in the start and complete events.

DPGrev:
We could add functionality for this iteratively. As a start, we could just put the schema in the AVRO byte stream which is one of the standard use cases for AVRO. This should work for al the producers which support bytestreams. 

The next step would be to actually add support for a schema registry and we replace the schema itself for a schema identifier. This would integrate particularly well with kafka. As an alternative we could also place the AVRO schemas and corresponding identifier to the maxwell DB to give users we do not want to use a schema registry, still an option to use schema identifiers. This is, of course, al optional.

PR #1392, opened 2019-11-25
Add schema to event

DPGrev:
After discussing in https://github.com/zendesk/maxwell/pull/1388 this would be a good first step to also support a format that contains schema. 

This is also useful for debugging.


DPGrev:
Hi @osheroff,

For us this use case has come up again. Was there a reason why that PR was not reviewed? I.e do we think it is a useful future to have the option to add the schema to the message?

If so, I would be willing to revisit this issue.

osheroff:
sorry, i have no idea why this dropped off.  Fix the merge conflicts and we can start looking.  I think there will be work to do to ensure that the schema types are reasonably consistent.

PR #1394, opened 2019-11-28
add the ability to have the type in the routing key for rabbitmq prod…

paulhoang-hee:
…ucer

This is rather a cheeky PR, its a small tweak to an existing feature for the RabbitMQ routing keys.

It would be really helpful for our team to have the ability to include the type of message in the routing key as we route our messages in the format [system].[entity(table)].[event(type)].

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/RabbitmqProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/RabbitmqProducer.java
@@ -6,6 +6,7 @@ import com.rabbitmq.client.ConnectionFactory;
 import com.rabbitmq.client.MessageProperties;
 import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.row.RowMap;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


osheroff:
i'm fine with this.  chance you could refactor the rabbitmq and kafka/kinesis producer to use common code for interpolation here so that everyone can bask in the glory of your cheeky feature?

paulhoang-hee:
haha, thanks for getting back so soon! sure i'll see if i can make the same changes for the kafka/kinesis producers

paulhoang-hee:
Hi @osheroff, I've done another commit, this includes the Kafka topic change but I couldn't find anything relating to the topics for Kenisis.

You've not asked for it but I've also made the change to the Redis producer to do the same.

osheroff:
ah @paulhoang-hee, sorry, I was not specific enough.  Can you make it so that the kafka and the redis producer actually share code to do the interpolation?  thx!

paulhoang-hee:
Hi @osheroff, no need to apologise. I've pushed another commit so that the Redis and Kafka producers share the same code. I was initially going to pull up the method into the Abstract class but then that would affect the 5+ or so other classes that extend it. I then thought about moving it to a `utils` class but that felt wrong as it would just be a single class with no other methods.

So I ended up created an interface with a default method. If this doesn't make sense please let me know. I also found it hard naming it so if you want that changed too, please let me know (naming things is hard)

osheroff:
why not just an instance method on `RowMap`?  Maybe not *exactly* the correct place to put it but probably good enough.

paulhoang-hee:
Hi @osheroff . I've moved the implementation to the RowMap class and deleted the previously created Interface.

Most of it went smoothly but the `MaxwellKafkaProducer` fallback flow was a little troublesome. It didn't have access to the RowMap instance so I used the queue's `peek` method to get one. Is that ok?

paulhoang-hee:
Just noticed travis came back with a broken build, guess I'll have to fix that first

Issue #1419, opened 2020-01-16
message.publish.age is (mostly) 0 after 1.24.0


ylebedeva:
After upgrading Maxwell from 1.22.4 to 1.24.0 (huge thanks for the updates!) we saw one of the metrics we monitor starting to behave oddly.

`message.publish.age` is now equal to 0 most of the time. On prior versions it reported age of messages seen in ms. It occasionally reports large spikes. See documentation https://github.com/zendesk/maxwell/blob/master/docs/docs/monitoring.md#metrics

I believe this commit changed its behaviour and particular values it is reporting: https://github.com/zendesk/maxwell/commit/2f3789b64098d9d507649efe25f91eb5f49f8595#diff-80a5ac0ad08c1ae8825e872470b1ac7bR42

<img width="1024" alt="Screen Shot 2020-01-16 at 11 13 25 AM" src="https://user-images.githubusercontent.com/2971259/72542010-74204600-3851-11ea-8794-d8b694fddc52.png">

osheroff:
merge & fixed in 1.24.1

PR #1420, opened 2020-01-16
Fix issue #1419 message.publish.age metric

ylebedeva:
Issue https://github.com/zendesk/maxwell/issues/1419
After updating from 1.22.4 to 1.24.0 we saw this metric report unreasonable values: 0 most of the time with a few spikes. This change reverts what appears to be unintentional change to it in SLO violation counter feature https://github.com/zendesk/maxwell/pull/1350


osheroff:
ah, yes, @ThuTrinh looks like we missed the subtle definition of publish time vs overall time (kafka time vs. overall database time).  I'm gonna merge this, @ThuTrinh you may want to make sure that the "age" definition is defined how you wanted....

osheroff:
@ylebedeva thx for the fix!

PR #1428, opened 2020-02-10
handle kinesis large values

rotemseekingalpha:
add producer_max_value_size
remove largest value (> producer_max_value_size) recursively from RowMap.data


osheroff:
intersting approach.  I wouldn't want this on by default, but I could see it as an option; cut out large blobs but get the important bits in by default.

For me to merge this PR I think:
- there'd need to be tests.  This probably means extracting the code into a module somewhere.
- there'd need to be configuration to turn it on
- it should work for kafka too.

PR #1445, opened 2020-03-19
Add realpath to be able to resolve symlink

AyrtonRicardo:
In out scenario we use maxwell via symlink, for eg. `/usr/bin/maxwell`, when using only `dirname` do not get the correct path because of symlink, the realpath will get the real maxwell path to be able to resolve the dependencies when using the symlink created.

### DIFF
--- a/bin/maxwell
+++ b/bin/maxwell
@@ -2,7 +2,7 @@
 
 set -e
 
-base_dir="$(dirname "$0")/.."
+base_dir="$(dirname "$(realpath "$0")")/.."
 lib_dir="$base_dir/lib"
 lib_dir_development="$base_dir/target/lib"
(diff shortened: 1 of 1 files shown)


osheroff:
this doesn't work on my machine, I don't have any such `realpath` utility.  Can you come up with another solution?

AyrtonRicardo:
@osheroff can you tell me which linux version/distribution do you use?
 I will use add a condition later for that :)

osheroff:
osx, locally.

AyrtonRicardo:
@osheroff oh, ok! I completely forgot about mac been able to use `bash`, just thought about linux, and indeed `realpath` does not exists to mac.
Check done via `type` command to validate if `realpath` exists, if not exists just do as before.

osheroff:
ok.  seems to work well enough in the various cases i've run it through.  thx for the work!

PR #1449, opened 2020-03-24
fix - using the correct var to define redis channel when using rpush.

lucastex:
When we try to use `rpush` as `redis_type`, it wasn't replacing the `%{table}` and `%{database}` vars 
in the `redis_key` because it was using `this.channel` and not the local var `channel`. 

This was not happening when using `lpush`, only with `rpush`


osheroff:
thanks for the PR!

Issue #1442, opened 2020-03-16
MaxwellSQLSyntaxError on drop column with qualified table name


danieltmartin:
`ALTER TABLE t DROP t.foo` gives `com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: .`

Running Maxwell v1.24.1 against AWS Aurora.

osheroff:
fixed in 1450

PR #1450, opened 2020-03-24
issue #1442, more on fully-qualified column names

osheroff:


osheroff:
not sure why tests aren't running.  merging.

PR #1462, opened 2020-04-06
Upgrades the docker JDK to 11.

taavi:
We've been having performance issues connecting to MySQL using SSL, and upgrading to JDK11 had a HUGE impact (2-4x) on performance.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -356,6 +356,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>2.9.1</version>
+        <configuration>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -356,6 +356,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>2.9.1</version>
+        <configuration>
+            <source>8</source>
+        </configuration>
```
taavi, line 361: This was required to allow the docs to build under JDK11.

osheroff:
okey doke.

PR #1469, opened 2020-04-19
add support to convert underscore naming style of fields in JSON output to camel case

FrankChen021:
in production, most of the fields in MYSQL tables are using underscore style naming strategy。And maxwell currently uses these raw names in its json output. 

````
CREATE TABLE `book` (
  `pk_id` bigint(20) NOT NULL AUTO_INCREMENT ,
  `cover_id` varchar(64) ,
  `cover_raw_link` varchar(1024) ,
  `link_to` varchar(64) ,
  `created_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `updated_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`pk_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;

````

in applications, such as java or c#, these underscore naming strategy are not widely used, so to convert maxwell's json output to java beans, we need to map these names to the name of fields in java beans.

````
public class Book {
        [JsonProperty("cover_id")] //<=======explicit map the property to the name in JSON
        public string CoverId { get; set; }

        [JsonProperty("pk_id")] //<=======explicit map the property to the name in JSON
        public long? PkId { get; set; }
}
````

or in java 

````
public class User {
        @JsonProperty("cover_id") //<=======explicit map the property to the name in JSON
        private string coverId;

        @JsonProperty("pk_id") //<=======explicit map the property to the name in JSON
        private Long PkId;
}
````

these mapping definitions are so boring that it brings more details in the underlying database definition.

2nd, most of ORM tools generate camel case style beans/pojos for us although it's underscore style in database. Because of the conflict of name styles in underlying database and these beans/pojos , we can not reuse those auto geneated beans/pojos for deserialization of JSON generated by maxwell.

So,  in maxwell's JSON ouput, if we can turn those underscore style names into camel case style names, which are widely used in java/c# beans/pojos, it will dramatically simplify the object definition in applications. 

if `underscore_to_camelcase` style is applied, all the names of fields are camel case style instead of underscore style. eg:

````
{
    "database": "book",
    "table": "book",
    "type": "update",
    "ts": 1587292391,
    "xid": 1200,
    "commit": true,
    "data": {
        "pkId": 417300,     //<===instead of pk_id
        "coverId": "123",   //<===instead of cover_id
        "coverRawLink": "22",//<===instead of cover_raw_link
        "linkTo": "new_value"//<===instead of link_to,
        "updatedAt": "2020-04-19 10:25:23"
    },
    "old": {
        "linkTo": "old_value",//<===instead of link_to
        "updatedAt": "2020-04-19 10:31:36"//<===instead of updated_at
    }
}
````

and for deserialization, we can define our data/old object as 

````
public class Book  {
    private Long pkId;
    private String coverId;
    private String coverRawLink;
    private String linkTo;
    private Date createdAt;
    private Date updatedAt;
}
````

or reuse the pojos generated by ORM tools.

this PR adds a configurable parameter to allow us whether to enable the underscore to camel case conversion. In default, it's not enabled for compatibility. 

And from the implementation perspective view, there's an interface defined to support more name conversion style, such as PASCAL style , in the future.

### CODE REVIEW

**Comment by osheroff** [comment]
seems like a reasonable enough feature, but can you:

1.   Fix whitespace in the PR: maxwell uses tabs, not spaces.  
2.  Undo any unnecessary changes to `import` sections (I have no idea why everyone's editor insists on fucking with code in this way.  it's terrible).
3.  Let's get rid of having all these interfaces and "enterprisey" java features; we can use an interface when we really really need one.  I'd like to see this code change as a simple `if (output_config.camelCase ) { }` block, not a large amount of interfaces and implementations. 

Thanks!
lemme know if you have any questions.

**Comment by FrankChen021** [comment]
@osheroff it's ok with the first two opinions. 

for opinion 3, i think it's better to leave the interface here, because there might be more strategies in the future. For example, in the c# world, PASCAL naming style is heavily used, with the interface here, we cadd just add a new implementation(I'm working on it) without any changes in `RowMap`.

**Comment by osheroff** [comment]
Sorry frank, I’m going to be firm here and say I need the code simplified.  I try very hard to follow “YAGNI” principles and keep everything as simple as possible.  

https://en.wikipedia.org/wiki/You_aren't_gonna_need_it?wprov=sfti1

**Comment by FrankChen021** [comment]
@osheroff a simplified version has been committed, pls check it.

File: src/main/java/com/zendesk/maxwell/row/FieldNameStrategy.java [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+		StringBuilder newName = new StringBuilder(oldName.length());
+		for (int i = 0, len = oldName.length(); i < len; i++) {
+			char c = oldName.charAt(i);
+			if (c == '_' ) {
+				if ( newName.length() > 0 //ignore the leading underscore 
+				  && i + 1 < len //this underscore is not the last character 
+				  ) {
+					char afterUnderscore = oldName.charAt(++i);
+					
+					newName.append(Character.isLowerCase(afterUnderscore)	? (char) (afterUnderscore - 'a' + 'A')
```
osheroff, line 40: how would this work with non-ascii charaters?  Do they simply fail the `isLowerCase` check?  can you add a test around it?

src/main/java/com/zendesk/maxwell/row/FieldNameStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+
+	private String strategyName;
+
+	public FieldNameStrategy(String strategy) {
+		this.strategyName = strategy;
+	}
+	
+	public String apply(String oldName) {
+		if ( UNDERSCORE_TO_CAMEL_CASE.equals(strategyName)) 
+			return caches.computeIfAbsent(oldName, k -> underscore2Camel(k));
```
osheroff, line 25: wondering how necessary the underscore-to-camel-case cache is.  My guess is that if you measured it, the cost of `String#hash` would be around the same as the cost of `underscore2Camel`.  Or close enough.

osheroff: also please *do* cache the result of `UNDERSCORE_TO_CAMEL_CASE.equals(strategyName)` in the constructor so you only have to do that string comparison once.

FrankChen021: > wondering how necessary the underscore-to-camel-case cache is. My guess is that if you measured it, the cost of `String#hash` would be around the same as the cost of `underscore2Camel`. Or close enough.

time consumption maybe the same, but cache prevents lots of temporary objects allocation and gc

src/main/java/com/zendesk/maxwell/row/FieldNameStrategy.java
FrankChen021: there're several ways to handle non-ascii characters,
1. discard all invalid characters for a variable name. 
2. keep them, leave them to user.
3. replace them one by one according some fixed rules

There're no perfect ways to handle it. if we discard invalid characters, there's some cases which will fail. `...` for example, is a valid name for a table or a column in mysql, after discarding, it turns an empty string.

JOOQ replaces `.` with underscore, and `!` with `_21`. if we apply some rules like that, it's too complex.

MyBatis replaces all underscore with empty chars,
````
  org.apache.ibatis.reflection.MetaClass

  public String findProperty(String name, boolean useCamelCaseMapping) {
    if (useCamelCaseMapping) {
      name = name.replace("_", "");
    }
    return findProperty(name);
  }
````

So my suggestion is to keep those characters, leave them to user. In code, user can still use `JsonProperty` annotation to map the field name in json to a rightful variable name.

File: src/main/java/com/zendesk/maxwell/row/FieldNameStrategy.java (hunk 2 above)
Comment (FrankChen021) [reply src/main/java/com/zendesk/maxwell/row/FieldNameStrategy.java:29]: > also please _do_ cache the result of `UNDERSCORE_TO_CAMEL_CASE.equals(strategyName)` in the constructor so you only have to do that string comparison once.

i've fixed it, please check it

File: src/main/java/com/zendesk/maxwell/row/RowMap.java [hunk 3]
```diff
@@ -123,66 +126,72 @@ private void writeMapToJSON(
...
+	private String applyNamingStrategy(String name) {
+		return this.fieldNameStrategy.apply(name);
+	}
+
 	public String toJSON() throws Exception {
 		return toJSON(new MaxwellOutputConfig());
 	}
 
 	public String toJSON(MaxwellOutputConfig outputConfig) throws Exception {
+		this.fieldNameStrategy = new FieldNameStrategy(outputConfig.namingStrategy);
```
osheroff, line 144: sorry I didn't notice the first time through: there's no need for this to be an instance variable.  Just make it a local variable please (and pass it into `writeMapToJSON`)

FrankChen021: yes, there's no need to declare an instance variable. i've updated the code.

osheroff:
ok.  thanks for the revisions.  👍

PR #1480, opened 2020-05-06
Redis reconnection handling

lucastex:
Hi @osheroff 

This pull request recreates the Jedis connection when the jedis lost its connection instead of just invoking the `connect()` method. 

I've noticed that after reconnection failed for three times, the messaged failed but it was not incrementing the maxwell's failed messages internal counter. I've moved this outside the loop with a control variable `sentToRedis` and removed the succes incrementing to the same block. What do you think?

If you prefer, I can sent another with only the reconnecting issue.


osheroff:
looks good to me!  did you test this code?  did it fix the issue?

lucastex:
Thanks @osheroff ! 

Yes, I've tested and it is running great in our DEV / QA and couldn't reproduce the problem. I'll check again in the production environment after the release. 

We've faced this issue only a few few times in DEV / QA, it happens mainly in our prod env.

Thanks a lot for the opportunity on contributing. Waiting for your next steps.

osheroff:
next steps, I'm merging.  thanks for the work!  I'll get a release out tomorrow-ish, I don't think there's anything else in the pipeline but I can easily do a patch release for this one.

lucastex:
Thanks a lot! As soon you release it, will test it here in production.

osheroff:
released in 1.26.1

lucastex:
That's it @osheroff 

Running in production since yesterday... I can see the errors happening, but right after the error, the connection restabilshes and keep sending to redis.

Thanks a lot for the help!

Issue #1497, opened 2020-05-24
Using maxwell with redis can corrupt messages if using bootstrapper in async mode


lucastex:
Hi @osheroff , me again :) 

After the reconnection problem solved in the other issue/pull request, our maxwell/redis environment is already running for some days in production.

In some cases, we had a chance to see "corrupted" messages, that was in form of an invalid json, and even with redis commands inside of it. You can see one example here:

```
"{\"database\":\"moskit\",\"oskit\",\"table\":\"project\",\"type\":\"bootstrap-insert\",
\"ts\":1590008497,\"data\":{\"id\":47908,\"created_by_id\":10113,\"customer_id\":7490,
\"date_created\":\"2019-11-28 18:38:38\",\"date_deleted\":null,\"last_updated\":\"2019-11-28 18:38:38\",
\"name\":\"Nome Projeto\",\"prevision_close_date\":null,\"responsible_id\":10113,\"source\":\"MANUAL\",
\"summary_id\":47908,\"step_id\":344,\"archived\":0,\
"archived_date\":null}}\r\n*3\r\n$5\r\nLPUSH\r\n$41\r\ncdc:search-service-worker:project:pending\r\n$389\r\n
{\"database\":\"moskit\",\"table\":\"p"
```
_line breaks on me, to make it better to read here in github_

We can see some problems in the message above:
- after the first "database:moskit", we can see another "oskit" property alone
- the message is not ended correctly: "table: p" is the end of the message
- after the end of the "correct" message, we can see a `\r\n*3\r\n$5LPUSH <queue_name>`. This is for sure the next redis command that would be sent after this.

Very strange at the first time, but after digging a little bit, I realized that every time this happens, was when we're running a bootstrap task. Then I found the same issue in some other projects that uses redis, and all of them, pointed to the same problem, that is that a Jedis instance is not thread safe as we can see here: https://github.com/xetorthio/jedis/wiki/Getting-started#using-jedis-in-a-multithreaded-environment

> You shouldn't use the same instance from different threads because you'll have strange errors. And sometimes creating lots of Jedis instances is not good enough because it means lots of sockets and connections, which leads to strange errors as well. A single Jedis instance is not threadsafe! To avoid these problems, you should use JedisPool, which is a threadsafe pool of network connections. You can use the pool to reliably create several Jedis instances, given you return the Jedis instance to the pool when done. This way you can overcome those strange errors and achieve great performance.

[...]

PR #1498, opened 2020-05-24
Using JedisPool instead of a direct Jedis instance

lucastex:
This fixes https://github.com/zendesk/maxwell/issues/1497
Description in the issue

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
@@ -35,15 +35,13 @@ public class MaxwellRedisProducer extends AbstractProducer implements StoppableT
 		this.interpolateChannel = channel.contains("%{");
 		this.redisType = context.getConfig().redisType;
 
-		if (jedisPool == null) {
-			jedisPool = new JedisPool(
-				createRedisPoolConfig(), 
-				context.getConfig().redisHost, 
-				context.getConfig().redisPort,
-				Protocol.DEFAULT_TIMEOUT,
-				context.getConfig().redisAuth, //even if not present jedispool will handle a null value 
-				context.getConfig().redisDatabase); //even if not present jedispool will handle a null value 
-		}
+		jedisPool = new JedisPool(
+			createRedisPoolConfig(), 
+			context.getConfig().redisHost, 
+			context.getConfig().redisPort,
+			Protocol.DEFAULT_TIMEOUT,
+			context.getConfig().redisAuth, //even if not present jedispool will handle a null value 
+			context.getConfig().redisDatabase); //even if not present jedispool will handle a null value 
 	}
 
 	private JedisPoolConfig createRedisPoolConfig() {
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
@@ -7,6 +7,9 @@ import com.zendesk.maxwell.util.StoppableTask;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisPool;
+import redis.clients.jedis.JedisPoolConfig;
+import redis.clients.jedis.Protocol;
 import redis.clients.jedis.StreamEntryID;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import java.util.HashMap;
@@ -17,7 +20,8 @@ public class MaxwellRedisProducer extends AbstractProducer implements StoppableT
 	private final String channel;
 	private final boolean interpolateChannel;
 	private final String redisType;
-	private Jedis jedis;
+
+	private static JedisPool jedisPool;
 
 	@Deprecated
 	public MaxwellRedisProducer(MaxwellContext context, String redisPubChannel, String redisListKey, String redisType) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -129,9 +143,7 @@ public void push(RowMap r) throws Exception {
 				break;
 			} catch (Exception e) {
 				if (e instanceof JedisConnectionException) {
-					logger.warn("lost connection to server, recreating redis connection...", e);
-					jedis.disconnect();
-					initRedisConnection();
+					logger.warn("lost connection to server, will try again with another connection from pool", e);
```
osheroff, line 146: does the pool throw away the connection and re-establish?  or is there a possibility of running out of pools?  

It'd be wonderful to establish an integration test for these behaviors, but that would blow out the scope of the PR greatly.  In lieu of that, is there an easy way to trigger `JedisConnectionException` manually and verify that the connection pool does the right thing and throws the connection away?

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -31,20 +35,27 @@ public MaxwellRedisProducer(MaxwellContext context) {
...
+				Protocol.DEFAULT_TIMEOUT,
+				context.getConfig().redisAuth, //even if not present jedispool will handle a null value 
+				context.getConfig().redisDatabase); //even if not present jedispool will handle a null value 
 		}
 	}
 
-	private void initRedisConnection() {
-		jedis = new Jedis(context.getConfig().redisHost, context.getConfig().redisPort);
-		jedis.connect();
+	private JedisPoolConfig createdRedisPoolConfig() {
```
osheroff, line 49: nitpick, maybe just name this `createRedisPoolConfig` or event `redisPoolConfig()`

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -31,20 +35,27 @@ public MaxwellRedisProducer(MaxwellContext context) {
...
 
-		initRedisConnection();
-
-		if (context.getConfig().redisAuth != null) {
-			jedis.auth(context.getConfig().redisAuth);
-		}
-
-		if (context.getConfig().redisDatabase > 0) {
-			jedis.select(context.getConfig().redisDatabase);
+		if (jedisPool == null) {
```
osheroff, line 38: I don't think we have to worry about / support the case where the call to `super` has mucked with `jedisPool` (it's deprecated), in other words can probably just eliminate the `null` check here.

osheroff:
good to go after we verify or test how `JedisPool` handles broken connections

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
lucastex: sorry for the typo

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
lucastex: ok!

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
lucastex: Yes it does. Since Jedis is a Closeable object, when the try block finishes, the connection is returned to the pool. As the pool is configured to `setTestOnBorrow(true)` the connection is recreated if is no good anymore.

I've tested here throwing the JedisConnectionException occasionally (time % 10), I could see a lot of errors (expected) but no downtime, all messages succeeded in the next atempt.

And I've measured the pool health with active connections and idle connections. The pool starts with 1/0 with a regular usage (binlog streaming). After starting a new bootstrap task, the pool stays with 2/0, when the bootstrap task finishes it stays with 1/1 till the idle connection drops.

IMHO, everything working great here.

lucastex:
There you go @osheroff !

src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
osheroff: great, thanks for the dilligence

lucastex:
Hi @osheroff  , everything is great here with the new version. We haven't faced the corrupted message not even once! 🎉🎉🎉

The only thing is that we can see that the bootstrap speed is slower than before. This seems to be linked to the fact that redis connection is being tested everytime the pool borrows it (just my thinking, not tested). Or even if you have any idea what this could be related...

I'll think on something to work on this in the future, and keep you updated!
Have any ideas?

Check these two images:
- First one the bootstraps that ran before and after the version change
- Second image, the same bootstrap before and after (same ammount of rows)

Screen Shot 2020-05-28 at 09 11 24

osheroff:
well, there's really only a couple of options, it's either the overhead of test-on-checkout or of the pool itself.  very likely it's the test-on-checkout. 

classically here you just turn off test-on-checkout and put the ugly reconnect logic back in :)

another option is to leave it on and just have the producer reserve the connection for its entire lifecycle, but I'm not sure that that's fundamentally different than turning off test-on-checkout...

lucastex:
Hi @osheroff 

I don't think turn off test-on-checkout would be the best approach since it is a safety condition, right? :) 

I've increased the velocity a lot bringing the redis instance for this project internally to our k8s instead of using the regular AWS Managed redis we were using before.

What else I did, is to test maxwell with another redis client (Redisson, that I've mentioned before when we started to dig on this problem, remember?). Just changing the client, I've achieved +20% speed on bootstrap. I'll open a new pull request with it later so we can discuss it better there.

Thanks a lot @osheroff

lucastex:
Done!  https://github.com/zendesk/maxwell/pull/1507

osheroff:
> I don't think turn off test-on-checkout would be the best approach since it is a safety condition, right? :) 

you can win a lot of speed by simply assuming the connection is in a good state and then rescuing the errors when it's not.  I don't think it's *unsafe* to avoid ping-on-checkout, but you do have to write more less-than-fun error handling code.

lucastex:
Hum, the problem is that redis won't throw an error. it will just send corrupted messages to my queue. The error will only "exists" in my consumer application, that is reading from redis.

In this issue https://github.com/zendesk/maxwell/issues/1497, it shows the message that was pushed to the redis queue, but that message haven't thrown an error while in maxwell, got it?

osheroff:
Sorry, I'm not sure I'm being clear enough.

Your issue in #1497 was caused by a lack of thread safety.  That's solved by bringing in `JedisPool`.  But the speed problem is likely caused by the line:

```
		poolConfig.setTestOnBorrow(true);
```

So what I'm suggesting is that we keep using `JedisPool`, but eliminate the `setTestOnBorrow` line.  This will create a solution with thread safety, where they won't corrupt messages, but instead of jedis pinging the server before every command, it will throw an error if/when the connection is dropped by the server.

lucastex:
Oh, got it now, sorry for the missunderstood. I'll do some tests on how jedis would behave when a connections fails using the jedispool (without `setTestOnBorrow(true)`) and get back to you.

PR #1507, opened 2020-05-30
Redis client implementation alternative to jedis (redisson)

lucastex:
Hi @osheroff 

This pull request adds an alternative redis client to use while using maxwell with redis publishing. I have not replaced jedis, since it is being used already and proven to deliver what it need to deliver.

The main difference between jedis/redisson is that redisson is threadsafe allowing us to share the same client instance between the bootstrap tasks thread and the main thread (streaming changes).

The way it is implemented, allow the user to choose between `jedis` and `redisson` client, using the new parameter `redis_client`. The default value it is still `jedis`, and the behaviour will not change for existing users/configs. If you feel that we don't need another redis client implementation and this could replace jedis, just tell me and I'll do it, ok?

The implementation is quite the same as the existing `MaxwellRedisProducer` just using another client. Using this implementation, I got bootstrap tasks done more than 20% faster, as this two examples shows:

Screen Shot 2020-05-30 at 18 26 55
 
@osheroff , this is not written in the documentation yet, I'll leave it to do later, as soon I hear from you what do you think of this pull request and if you'd like to continue with both redis clients or just one.

Thanks a lot, looking forward to hear from you


osheroff:
question, how does redisson compare against jedis without the ping-on-checkout code?  I'd love to have the fastest possible solution here....

lucastex:
Hi @osheroff 

The numbers are a little bit confusing right now. All the tests are in my local development machine, using a docker-compose that runs my database/redis/maxwell instances.

I realized that the lower numbers was when running from the packaged jar directly, instead of running from a docker image.

I did some tests here with jedis in the current image 1.26.3, and the initial one, 1.26.1. 
Then, I've run locally with the packaged jar, with jedis and redisson, and with a docker image built locally (`docker build -t maxwell_dev:local .`) with jedis and redisson.

From all the numbers, running from the jar with jedis was the worst approach, while running on redisson or jedis on docker gave me better results. I know in production this will certainly change, but can't tell for sure how much it will.

Can you run a benchmark on your environment so we can have more results to compare?

Here are the numbers:

Screen Shot 2020-05-31 at 10 07 49

lucastex:
Hi @osheroff ,

Did some other tests here, using the redisson async alternative, that have a connection pooling inside of it. This alternative can handle even more pushs to redis, since have other connections inside of it waiting to be used.

The initial redisson pool is configured to handle 64 connections, I've made tests using the default value, and using a smaller value, 10 in my case, and got almost the same results, which lead me to think that we're reaching the row input throughput in the process.

With this 10 connections in the pool, I got faster results than the initial jedis version (maxwell v1.23.1) that had the initial problem of corrupting messages and not handling broken connections.

The initial version (maxwell v1.23.1) got `383 rows/sec` for my 100k records bootstrap, while this redisson async version got `438 rows/sec` running from the jar and `502  rows/sec` running from a docker image that I've built locally.

Screen Shot 2020-06-01 at 07 37 31

In this last commit, this async version is written as another redis_client option `redis_client=redisson_async`, since we're still discussing this and I don't know which path you'll prefer.

We can leave all 3 implementations (which I think is not necessary), we can leave one jedis and one redisson, or we can leave only the fastest one. Or we can leave both for some versions and remove one later in another versions. This is your call since I'm just contributing here!

If we go in this async path, I can define more options for the user config, so he can define the internal redisson pool size or other configs. And of course, I'll add some docs explanation on this.

I'm very excited with these numbers, and hope we can see them in production!

Thanks for the opporunity @osheroff

osheroff:
async seems like it opens up the possibility for out-of-order rows, doesn't it?  

That's sort of fine for the bootstrapping process but seems like a pretty dangerous window to open for the regular producer.

lucastex:
Hum, thinking like this, even a pool with 2 connections could ran into out-of-order messages, right? Maybe this is the point on sharing the same redis connection between bootstrap/streaming data.

But Jedis wouldn't allow this, we really need a threadsafe lib, with one and only one connection, but that can be shareable between threads.

Is this correct? If so, I can try to work on something. Besides Redisson we have Lettuce, another java redis client that is threadsafe and could be used.

lucastex:
Or different entrypoints for bootstrap / streaming data. Do you think this is possible (looking for the maxwell architecture)? 

If so, on bootstrap tasks we could achieve a higher throughput with multiple connections, while in streaming data, we can guarantee the message order and integrity.

osheroff:
> Or different entrypoints for bootstrap / streaming data. Do you think this is possible (looking for the maxwell architecture)?

we could probably switch queues based just on the message type of the `RowMap`; if it's a `bootstrap-insert`, we go async.

However,  I'm not quite sure this is worth it for the speed boost.  What would *really* get you a speed boost, without sacrificing thread safety, is moving to an aysnc consumer / pipelining model:

- clients push to the producer, which puts it in a queue and returns immediately
- a producer thread waits somewhere between 1 and 10ms for rows to enter the queue.
- producer grabs all the rows in the queue and starts a redis pipeline operation, batching them all up at once.

this should greatly improve the overall throughput of the system, at the tradeoff of a touch of latency in the case of a very low output system.

lucastex:
Hi @osheroff , since we're not going on the other redis client solution, I think this PR can be closed right? I'll still try to get some extra time here to test the jedis pool without testing the connection everytime it is borrowed from the pool, but if / when this happens, I think we can continue on the other issue we're discussing and with another pull request, what do you think?

PR #1516, opened 2020-06-08
ensure bootstrap's connection URI to use jdbcoptions

erichll:


osheroff:
can you clarify what problem you're trying to fix?  the PR is fine, I just want to know precisely the trouble you're having...

erichll:
> clarify



> can you clarify what problem you're trying to fix? the PR is fine, I just want to know precisely the trouble you're having...

When running bootstrap without set "serverTimezone=Asia/Shanghai" will produce "java.sql.SQLException: HOUR_OF_DAY: 2 -> 3"  in infinite loop.

PR #1518, opened 2020-06-08
Mysql NOW function precision support

lukestephenson-zendesk:
The now function optionally supports a precision.

https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_now

Maxwell was getting stuck on a binlog event which was creating a table with 
```
`created_at` datetime(6) DEFAULT NOW(6) NOT NULL
```


lukestephenson-zendesk:
cc @zendesk/goanna

Issue #1525, opened 2020-06-23
Techniques for joining maxwell data with application context


jl-gogovapps:
Hello, 
I've started experimenting with Maxwell for CDC, with the hope that I can use it to generate Audit logs.  
One requirement for audit logs is understanding who within the application made the change.  

I was hoping that the `thread_id` would be a viable way to create this linkage, and some have suggested publishing application context to a blackhole table.  This sounded promising, but seemed to collapse for me in the context of Connection Pooling, where the connection_id (thread_id for mysql) is reused.

I could refactor all my database tables to have a "last_modified_by_id", but I'm trying to avoid that, since it doesn't help for hard deletes, and requires quite a bit of overhaul across a very legacy application.  

I've been trying to determine if the transaction_id (xid) is accessible by an application (Sinatra/rails) in a relatively opaque way, does anyone have any experience here that might help me out?

Thanks in advance
JL

osheroff:
I looked into this, under the hood, down into that deep deep cave that is mysql, and I don't have good news.  There's plenty of features that surround this, but nothing that seems to solve it.  `information_schema.innodb_trx` has a transaction id, sure, but it doesn't correspond to anything in the binlog.  If you're using XA style transactions there's a host of features lining up to give you back *that* XID.  But nothing baked in that seems to grant you the holy rites of passage that are knowing something correlated to the backend.

Here's some hacky ideas, you've probably thought of a lot of them:

- at the top and bottom of a web request, write the current thread_id into the BLACKHOLE table, along with whatever unique identifier you plan to put in your audit logs.  Assuming that you retain a connection for the duration of your request (which I believe activerecord does out of the box), everything done by that thread between the blackhole markers will be audited.
- do the same but do it per-transaction, hooking in right after your mysql client issues a BEGIN statement.  
- probably a bunch of variations on that theme.

this really *should* be available.

jl-gogovapps:
@osheroff thank you so much for the response, I greatly appreciate it.

I went quite a distance with the Blackhole approach:
- Set up the blackhole table
- Created a concern (ruby) that would send thread ids and context data (a concern seemed like a relatively scalable way to do this across the system, with low lift)
- Created a kinesis stream
- Created a kinesis data analytics app to self join the stream based on the thread id over a window and produce a joined stream
- Created a KCL consumer to process the joined stream, manage audit tables, and generate audit logs

[...]

PR #1533, opened 2020-06-29
Allow Publishers to produce messages by thread ID

jl-gogovapps:
Based on discussion #1525 
```
21:11:40,738 DEBUG KinesisCallback - ->  key:820361, shard id:shardId-000000000000, sequence number:49608083649042269680942954152460432258721642278002098178
21:11:40,738 DEBUG KinesisCallback -    {"database":"xxx","table":"xxx","type":"insert","ts":1593465100,"xid":26209465,"commit":true,"position":"mysql-bin-changelog.146334:257588","server_id":382330091,"thread_id":820361,"schema_id":12,"primary_key":[135449],"primary_key_columns":["id"],"data":{...}}
```

The update here is that the `key:820361` corresponds with the thread id

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Maxwell changelog
 
-### [v1.26.5](https://github.com/zendesk/maxwell/releases/tag/v1.26.5): "Producer partitioning by thread id"
-
-
- - Allows Kinesis Producer to partition Kinesis Records by thread id
-
-
 ### [v1.26.4](https://github.com/zendesk/maxwell/releases/tag/v1.26.4): "Now function precision support"
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Maxwell changelog
 
+### [v1.26.5](https://github.com/zendesk/maxwell/releases/tag/v1.26.5): "Producer partitioning by thread id"
+
+
+ - Allows Kinesis Producer to partition Kinesis Records by thread id
+
+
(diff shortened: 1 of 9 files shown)


osheroff:
looks good thx!  sorry to bug you but can you undo the changes that bump the version?  I have a scripts that do that upon release and this will mess them up

jl-gogovapps:
Ah, no problem, I should be able to get to this today.  Thanks!  
I've been running this in our staging environment since yesterday morning, haven't **observed** any issues so far

jl-gogovapps:
Just wanted to share this bit of code too, if someone were searching the depths of the internet like I was :D 

```
  class Cdc
    def initialize app
      @app = app
    end

    def generate_marker(env)
      request = Rack::Request.new(env)
      actor_id = request.env[:token]&.dig('data','user','id')
      context = {
        actor_id: actor_id,
        ip_address: request.forwarded_for || request.ip
      }
      ActiveRecord::Base.connection.execute(
        <<SQL
INSERT INTO cdc_thread_requests (thread_id, context) VALUES (CONNECTION_ID(), '#{JSON.generate(context)}')
SQL
      )
    end

    def should_generate_marker?(env)
      return false if env['REQUEST_METHOD'].downcase.to_sym == :options
      return false if ENV['RACK_ENV'] == 'test'
      true
    end

    def call env
      generate_marker(env) if should_generate_marker?(env)
      response = @app.call(env)
      generate_marker(env) if should_generate_marker?(env)
      response
    end
  end
```

In a sinatra app, for example:
```
use Cdc
```
after verifying and unpacking a JWT token into `env[:token]`

osheroff:
thx for paying it forward.

PR #1536, opened 2020-07-10
Add aws-java-sdk-sts@1.11.782 to support EKS IRSA

robwittman:
In order to support usage of EKS IRSA, the aws-java-sdk-sts module has to be included.

---
This change is [Reviewable](https://reviewable.io/reviews/zendesk/maxwell/1536)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -295,6 +295,11 @@
       <version>1.11.782</version>
     </dependency>
     <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-sts</artifactId>
+      <version>1.11.782</version>
+    </dependency>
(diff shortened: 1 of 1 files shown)


lukestephenson-zendesk:
Is this specific to a particular way of deploying maxwell?  Questioning if it needs to be a maxwell dependency.

robwittman:
AWS added the `WebIdentityTokenFileCredentialsProvider` to the credential chain, allowing the SDK to authenticate to AWS, notably using EKS IRSA. Without the `aws-java-sdk-sts` module, the `WebIdentityTokenFileCredentialsProvider` can't be used. I believe this is because we include individual portions of the SDK rather than all of it

PR #1539, opened 2020-07-15
Redis producer - support both sentinel and traditional Redis connection

FanAs:
Hello,
We are preparing High Availability solution for our system that requires Sentinel support (https://redis.io/topics/sentinel) for Redis.
I thought that it would be great to have it in Maxwell, so I created this PR.

Maxwell configuration will have following configuration parameters:
* `redis_sentinels` - as a list of sentinels in format host1:host1,host2:port2,...
* `redis_sentinel_master_name` - as a master name of the sentinel

Both options are non-required (to keep compatibility with `redis_host` and `redis_port`).

I added required information to documentation pages.

Best regards,
Artyom.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
@@ -7,13 +7,18 @@ import com.zendesk.maxwell.util.StoppableTask;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisPoolAbstract;
 import redis.clients.jedis.JedisPool;
+import redis.clients.jedis.JedisSentinelPool;
 import redis.clients.jedis.JedisPoolConfig;
 import redis.clients.jedis.Protocol;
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/producer/MaxwellRedisProducer.java
```diff
@@ -64,12 +84,20 @@ private String generateChannel(RowIdentity pk){
 		return channel;
 	}
 
+	private Jedis getJedisResource() {
+		if (jedisSentinelPool != null) {
+			return jedisSentinelPool.getResource();
+		}
+
+		return jedisPool.getResource();
+	}
```
osheroff, line 93: rather than these two `if` statements, can we just make `jedisPool` and instance of the common parent, `Pool<Jedis>`?

osheroff:
thanks for the PR!  Hey @timbertson any idea what the license and security checks that are failing here are?

Also, the circle CI configuration appears to be gone.  So that's worrying..

FanAs:
Hello @osheroff, thanks for a code review.
I didn't know about `JedisAbstractPool` - I replaced it with abstract pool.
And all checks are now passed (IDK why).

PR #1542, opened 2020-07-21
Filter columns

deverlex:
Allow white list of columns in the table


osheroff:
thanks for the PR!  I'd like to see:

- rename this into `output_column_filter` or something
- tests 
- make a speciailized version of `ColumnFilter` that always requires a column name; ie, you can say `*.*.col` but you can't just pass it `db.table` as that makes no sense.
- tests

PR #1563, opened 2020-08-17
Correct comparision of age against SLO to be ms based

lukestephenson-zendesk:
:pear: @thutrinh @kayison

The config is specified in seconds, but the timestamp compared to is in milliseconds.  This rectifies the issue by converting the specified config value into ms.

cc @zendesk/goanna

### DIFF
--- a/src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/AbstractAsyncProducer.java
@@ -36,12 +36,12 @@ public abstract class AbstractAsyncProducer extends AbstractProducer {
 				if (message != null) {
 					context.setPosition(message.position);
 					long currentTime = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


samzx (commented):
I suppose the config can't be in milliseconds because of the accuracy of the time stamps on the big log event?

lukestephenson-zendesk:
@samzx Config is in seconds because that is how it was documented originally. https://github.com/zendesk/maxwell/blob/4386e6daaf8784bdb192c85fb0663c70145c83f5/src/main/java/com/zendesk/maxwell/MaxwellConfig.java#L355

I would have preferred the config var was named to indicate the unit like `metricsSloSecs`, but that would be a breaking change at this point.

samzx (approved):
Ah - makes sense. 👍

Users upgrading may see increase in slo metric - release notes should let the user know how to fix.

ThuTrinh (approved):

PR #1568, opened 2020-09-01
[1567] Possible SRID fix?

FanAs:


osheroff:
looks fine, add a test somewhere in `DDLParserTest` and can merge..

FanAs:
Hello @osheroff,
I added a test that showed that implementation wasn't correct.

Could you please look at it one more time? It's one of column options, so I needed to add definition to `column_definitions.g4` with `SRID INTEGER_LITERAL`

PR #1584, opened 2020-09-28
schema compaction, try two

osheroff:
here we make sure we take a lock and respect other clients' schemas, and run the compactor in a thread that's isolated from the rest of the maxwell world.

still TODO: actually wire this thing up to configuration. more tests.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellContext.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellContext.java
@@ -13,6 +13,7 @@ import com.zendesk.maxwell.schema.MysqlSchemaCompactor;
 import com.zendesk.maxwell.schema.PositionStoreThread;
 import com.zendesk.maxwell.schema.ReadOnlyMysqlPositionStore;
 import com.zendesk.maxwell.util.C3P0ConnectionPool;
+import com.zendesk.maxwell.util.RunLoopProcess;
 import com.zendesk.maxwell.util.StoppableTask;
 import com.zendesk.maxwell.util.TaskManager;
 import org.slf4j.Logger;
@@ -239,6 +240,25 @@ public class MaxwellContext {
 		return this.terminationThread;
 	}
 
+	public Thread startTask(RunLoopProcess task, String name) {
+		Thread t = new Thread(() -> {
+			try {
+				task.runLoop();
+			} catch (Exception e) {
+				LOGGER.error("exception in thread: " + name, e);
+				try {
+					this.terminate(e);
+				} catch (Exception exception) {
+					exception.printStackTrace();
+				}
+			}
+		}, name);
+
+		t.start();
+		addTask(task);
+		return t;
+	}
+
 	public Exception getError() {
 		return error;
 	}
@@ -409,17 +429,8 @@ public class MaxwellContext {
 			currentSchemaID
 		);
 
-		this.bootstrapControllerThread = new Thread(() -> {
-			try {
-				this.bootstrapController.runLoop();
-			} catch (Exception e) {
-				e.printStackTrace();
-			}
-		}, "maxwell-bootstrap-controller");
-		bootstrapControllerThread.start();
+		this.bootstrapControllerThread = this.startTask(this.bootstrapController, "maxwell-bootstrap-controller");
 
-
-		addTask(this.bootstrapController);
 		return this.bootstrapController;
 	}
 
@@ -435,15 +446,7 @@ public class MaxwellContext {
 				this.getCaseSensitivity()
 		);
 
-		Thread compactorThread = new Thread(() -> {
-			try {
-				compactor.runLoop();
-			} catch (Exception e) {
-				e.printStackTrace();
-			}
-		}, "maxwell-schema-compactor");
-		compactorThread.start();
-		this.addTask(compactor);
+		this.startTask(compactor, "maxwell-schema-compactor");
 	}
 
 	public Filter getFilter() {
--- a/src/main/java/com/zendesk/maxwell/MaxwellContext.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellContext.java
@@ -9,6 +9,7 @@ import com.zendesk.maxwell.recovery.RecoveryInfo;
 import com.zendesk.maxwell.replication.*;
 import com.zendesk.maxwell.row.RowMap;
 import com.zendesk.maxwell.schema.MysqlPositionStore;
+import com.zendesk.maxwell.schema.MysqlSchemaCompactor;
 import com.zendesk.maxwell.schema.PositionStoreThread;
 import com.zendesk.maxwell.schema.ReadOnlyMysqlPositionStore;
 import com.zendesk.maxwell.util.C3P0ConnectionPool;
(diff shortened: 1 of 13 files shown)


osheroff:
well, here we are, 18 months after the original PR @jakubkulhan.  Maybe you're still interested, maybe you've shuffled off this mortal coil.  Hard to say what happens in such a time frame. 

@timbertson or @timbertson-zd or whoever maybe is minding the Maxwell shop over there at Zendesk, I wouldn't mind a code review on this one, given the generally dangerous nature of it all.  Yeah, it's off by default, but I'd like to try to think through the edge cases all the way before shipping it.

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
```diff
@@ -0,0 +1,197 @@
...
+				}
+			} finally {
+				releaseLock(cx);
+			}
+		}
+	}
+
+	private boolean shouldCompact(Connection cx) throws SQLException {
+		ResultSet rs = cx.prepareStatement(
+			"select count(*) as count from `schemas` where `server_id` = " + this.serverID
```
timbertson-zd, line 87: IIRC there's a bunch of reasons why a server might have a broken schema chain. Assuming we only compact the latest chain, this might keep triggering no-op compactions forever.

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
```diff
@@ -0,0 +1,197 @@
...
+	private Long chooseCompactedSchemaBase(Connection cx) throws SQLException {
+		if ( !shouldCompact(cx) ) {
+			return null;
+		}
+
+		ResultSet rs = cx.prepareStatement(
+			"select id, binlog_file, binlog_position, gtid_set, 0 as last_heartbeat_read "
+			+ " from `schemas` where `server_id` = " + this.serverID
+			+ " order by id desc limit 1"
+		).executeQuery();
```
timbertson-zd, line 103: Looks like this duplicates but doesn't fully replicate the logic in https://github.com/zendesk/maxwell/blob/e48b0fc957c33062b32d23c21ddfb6a26b39191d/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java#L578-L588

Can we share them?

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
```diff
@@ -0,0 +1,197 @@
...
+	public void doWork() throws Exception {
+		try ( Connection cx = maxwellConnectionPool.getConnection() ) {
+			cx.setAutoCommit(false);
+			try {
+				if (getLock(cx)) {
+					compact(cx);
+				}
+			} finally {
+				releaseLock(cx);
+			}
```
timbertson-zd, line 81: I don't know much about mysql locking, but presumably it's possible for the releaseLock to fail, on connection error etc.

 - is the lock auto-released when the connection dies?
 - should we explicitly shutdown the connection on error (and take it out of the pool) to ensure that?

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
```diff
@@ -0,0 +1,197 @@
...
+		slowDeleteFrom("columns", cx, schemaID);
+		slowDeleteFrom("tables", cx, schemaID);
+		slowDeleteFrom("databases", cx, schemaID);
+		cx.createStatement().executeUpdate("delete from `schemas` where id = " + schemaID);
+	}
+
+	private static final int DELETE_SLEEP_MS = 200;
+	private static final int DELETE_LIMIT = 500;
+
+	private void slowDeleteFrom(String table, Connection cx, long schemaID) throws SQLException {
```
timbertson-zd, line 184: Is this just to prevent a huge transaction? And I know these strings are constant-ish, but can you use a prepared statement for the table name + ID?

timbertson-zd (commented):
Are there related issues/ PRs? I feel like last time we looked too deep into this we realised it wasn't actually compacting when we expected, but my memory is pretty foggy.

These days @lukestephenson-zendesk's team owns maxwell, but I'm still around in an adjacent team and happy to help out with reviews.

Overall it looks pretty sane. I'm unsure about the purpose of the locking though.

lukestephenson-zendesk (commented):
Who wanted this feature?  I'm not exactly sure what motivated this work.

src/main/java/com/zendesk/maxwell/MaxwellContext.java
```diff
@@ -422,6 +423,29 @@ public synchronized BootstrapController getBootstrapController(Long currentSchem
...
+				this.config.clientID,
+				this.getServerID(),
+				this.getCaseSensitivity()
+		);
+
+		Thread compactorThread = new Thread(() -> {
+			try {
+				compactor.runLoop();
+			} catch (Exception e) {
+				e.printStackTrace();
```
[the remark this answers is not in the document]
osheroff: should probably terminate the process.  There's a lot of repeated code around this stuff, I'll see if I can refactor...

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
osheroff: anything not on the chain will get deleted on the first compaction, I think.

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
osheroff: mmm, no, one asks "most-recent-for-this-position" and one asks "most-recent-overall".

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
osheroff: these are good questions, I'll check.  My assumption was that the lock would be released on CX dying.  If mysql automatically releases the lock when the connection dies, isn't that (plus the `finally` statement`) good enough to ensure we don't starve?

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
osheroff: yeah mostly to prevent a huge transaction.  Mysql also does some weird shit with like compaction threads when you delete a ton of rows at once (although that may be a concern from a different time, don't remember).

I don't think I can inject a table name into a prepared statement.  IDs I can, but other than a marginal performance benefit I'm not sure of the value.

osheroff:
> Who wanted this feature? I'm not exactly sure what motivated this work.

This came from the community, someone who has a large amount of schema churn was getting overwhelmed by number of rows.  It's also not a terrible feature to have generally, a maxwell installation that grows unbounded for all time will eventually have very long startup times as the # of schema deltas grows.

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
```diff
@@ -0,0 +1,197 @@
...
+
+	private void releaseLock(Connection cx) throws SQLException {
+		PreparedStatement s = cx.prepareStatement(
+			"SELECT RELEASE_LOCK(?)"
+		);
+		s.setString(1, this.lockName());
+		s.execute();
+	}
+
+	public void doWork() throws Exception {
```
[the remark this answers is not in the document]
osheroff: probably not but it's following the signature of an interface

src/main/java/com/zendesk/maxwell/schema/MysqlSchemaCompactor.java
timbertson-zd: I think so, yeah. Unless there's exceptions in `releaseLock` that don't cause the connection to die, but that sounds unlikely and hard to fix anyway 🤷‍♂️

osheroff: > A lock obtained with GET_LOCK() is released explicitly by executing RELEASE_LOCK() or implicitly when your session terminates (either normally or abnormally). Locks obtained with GET_LOCK() are not released when transactions commit or roll back.

PR #1591, opened 2020-10-23
Maxwell HA

leoChaoGlut:
1.add com.101tec:zkclient
2.add MaxWellHA.java, and mutiple instance test passed
3.add maxwell-ha startup script

### DIFF
--- a/src/main/java/com/zendesk/maxwell/zk/ZkService.java
+++ b/src/main/java/com/zendesk/maxwell/zk/ZkService.java
@@ -1,0 +1,132 @@
+package com.zendesk.maxwell.zk;
+
+import org.I0Itec.zkclient.IZkStateListener;
+import org.I0Itec.zkclient.ZkClient;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.zookeeper.Watcher;
+import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.UUID;
+
+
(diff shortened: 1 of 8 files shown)


src/main/java/com/zendesk/maxwell/zk/ZkService.java
```diff
@@ -0,0 +1,134 @@
+package com.zendesk.maxwell.zk;
+
+import org.I0Itec.zkclient.IZkStateListener;
+import org.I0Itec.zkclient.ZkClient;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.zookeeper.Watcher;
```
lukestephenson-zendesk, line 6: Maxwell currently doesn't have any dependency on zookeeper. Given it does have a mysql dependency, why not use mysql locking as suggested by @osheroff rather than adding a dependency on zookeeper.

src/main/java/com/zendesk/maxwell/zk/ZkService.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.UUID;
+
+
+public class ZkService {
```
lukestephenson-zendesk, line 16: if zookeeper is what we want to couple maxwell to for leader election, would it be better to make use of an apache curator recipe for that purpose?

PR #1609, opened 2020-11-27
high availability via jgroups-raft

osheroff:
While many maxwell users at this point probably have rolled their own singleton-cluster-management solution (via kubernetes, consul, etcd,  etc, etc), having a baked-in HA story has been an oft-requested feature.  For many it's simply not feasible to tell them "go implement a giant PAAS architecture in order to pick up some high-availability".  

so here's a fairly dirty and cheap HA story using @belaban's excellent [jgroups-raft](https://github.com/belaban/jgroups-raft).  

Note that the implementation of the *lock* is imperfect (as according to https://github.com/belaban/jgroups-raft/issues/79) and http://belaban.blogspot.com/2020/11/i-hate-distributed-locks.html; in order to do this *correctly* we'd probably want to move the data in `maxwell`.`positions` into the raft cluster itself, and require a successful `write()` operation to the RAFT cluster to make progress.

Note also that at least for now, I don't care about the possibility of two simultaneous singletons; (a) the overlap will be brief, as the old, partitioned leader will soon notice the lack of quorum and step down, and (b) if the maxwell clients are configured correctly the server will enforce a singleton anyway when it checks the uniqueness of the server_id.  

Also; I tried atomix first in all of this.  It did not pass muster.  It did not pass much.  

cc @zendesk/goanna @timbertson (i think goanna is the Zendesk side of maxwell maint these days? )

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -317,12 +317,12 @@
     <dependency>
       <groupId>io.atomix</groupId>
       <artifactId>atomix</artifactId>
-      <version>3.0.6</version>
+      <version>3.1.6</version>
     </dependency>
     <dependency>
       <groupId>io.atomix</groupId>
       <artifactId>atomix-raft</artifactId>
-      <version>3.0.6</version>
+      <version>3.1.6</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -314,6 +314,16 @@
       <artifactId>simpleclient_servlet</artifactId>
       <version>0.5.0</version>
     </dependency>
+    <dependency>
+      <groupId>io.atomix</groupId>
+      <artifactId>atomix</artifactId>
+      <version>3.0.6</version>
+    </dependency>
+    <dependency>
+      <groupId>io.atomix</groupId>
+      <artifactId>atomix-raft</artifactId>
+      <version>3.0.6</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -318,7 +318,7 @@
     <dependency>
       <groupId>org.jgroups</groupId>
       <artifactId>jgroups-raft</artifactId>
-      <version>0.5.3-SNAPSHOT</version>
+      <version>1.0.0.Final</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -314,15 +314,11 @@
       <artifactId>simpleclient_servlet</artifactId>
       <version>0.5.0</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.jgroups/jgroups-raft -->
     <dependency>
-      <groupId>io.atomix</groupId>
-      <artifactId>atomix</artifactId>
-      <version>3.1.6</version>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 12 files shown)


src/main/java/com/zendesk/maxwell/MaxwellHA.java
```diff
@@ -0,0 +1,67 @@
...
+				isRaftLeader.set(false);
+			} else {
+				LOGGER.info("lost HA election, current leader: " + handle.leader());
+			}
+		});
+
+		ch.connect(this.clientID);
+		LOGGER.info("enter HA group, current leader: " +  handle.leader());
+
+		Thread.sleep(Long.MAX_VALUE);
```
lukestephenson-zendesk, line 65: If `maxwell.run()` fails to execute, I'm assuming in non HA mode the process dies.  But here it looks like the process would just hang.  Is that right?

lukestephenson-zendesk (commented):
I've not looked into raft much, but this seems reasonable.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -223,6 +222,10 @@ protected MaxwellOptionParser buildOptionParser() {
 		parser.accepts( "max_schemas", "Maximum schemas to keep before triggering a compaction operation.  Default: unlimited" ).withRequiredArg();
 		parser.section("operation");
 
+		parser.accepts( "ha", "enable high-availability mode via jgroups-raft" );
+		parser.accepts( "jgroups_config", "location of jgroups xml configuration file" ).withRequiredArg();
+		parser.accepts( "raft_member_id", "raft memberID.  (may also be specified in raft.xml)" ).withRequiredArg();
```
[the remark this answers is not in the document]
osheroff: yeah, they are actually totally option (even if ha is enabled) -- you can read `.withRequiredArg()` as saying "if `--jgroups_config` is there, it needs to be given a string (ie it's not just a flag argument)

src/main/java/com/zendesk/maxwell/MaxwellHA.java
osheroff: very good question, lemme study the code and/or run it and figure out the answer to this one.  I don't think we should attempt to continue if `maxwell.run()` crashes for whatever reason.

osheroff: utility-of-code-review of the year award there @lukestephenson-zendesk.  Updated so that HA process still crash on exception.

pom.xml
```diff
@@ -314,6 +314,12 @@
       <artifactId>simpleclient_servlet</artifactId>
       <version>0.5.0</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.jgroups/jgroups-raft -->
+    <dependency>
```
lukestephenson-zendesk, line 318: Not directly related to this change, but I find it frustrating that maxwell has so many dependencies, lots of which are not used in all use cases.  For those of us that have enterprise tools to do vulnerability scanning, it creates noise with dependencies that have vulnerabilities which we aren't even using.  I'd love to see maxwell become more modular (but appreciate that is a lot of effort).

lukestephenson-zendesk (approved):
Seems reasonable.

PR #1643, opened 2021-02-02
BinlogConnectorReplicator: add heartbeat detection [MELINF-2251]

timbertson-zd:
Since deploying v1.27.x, we've had a small but noticeable uptick in stuck maxwells - no events flowing, but maxwell itself is running with no errors, still reporting metrics etc.

My main suspect is https://github.com/zendesk/maxwell/pull/1548. That disables binlog replicator's dead connection logic because it clashes with maxwell's own functionality. But maxwell's functionality is pretty weak here - it only knows about active failures (i.e. exceptions), if the connection simply goes quiet it won't notice.

After reading https://github.com/shyiko/mysql-binlog-connector-java/issues/118 I think maxwell should use binlog-connector's heartbeating. We still don't want to enable BC's keepalive thread (because it conflicts with our reconnect logic), but we can enable heartbeating and then reimplement the heartbeat detection as part of our existing dead connection detection.

/cc @zendesk/goanna

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -36,6 +36,7 @@ replication_password           | STRING               | password on replication 
 replication_port               | INT                  | port on replication server                          | 3306
 replication_user               | STRING               | user on replication server                          |
 replication_ssl                | [SSL_OPT](#sslopt)   | SSL behavior for replication cx cx                  | DISABLED
+replication_heartbeat          | BOOLEAN              | enable replication heartbeats to detect stale connections | ENABLED
 replication_jdbc_options       | STRING               | mysql jdbc connection options for replication server| [DEFAULT_JDBC_OPTS](#jdbcopts)
 &nbsp;
 schema_host                    | STRING               | server to capture schema from.  See [split server roles](#split-server-roles) | *schema-store host*
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -30,13 +30,13 @@ master_recovery                | BOOLEAN              | enable experimental mast
 gtid_mode                      | BOOLEAN              | enable GTID-based replication                       | false
 recapture_schema               | BOOLEAN              | recapture the latest schema. Not available in config.properties. | false
 max_schemas                    | LONG                 | how many schema deltas to keep before triggering compaction operation | unlimited
+binlog_heartbeat               | BOOLEAN              | enable binlog heartbeats to detect stale connections | DISABLED
 &nbsp;
 replication_host               | STRING               | server to replicate from.  See [split server roles](#split-server-roles) | *schema-store host*
 replication_password           | STRING               | password on replication server                      | (none)
 replication_port               | INT                  | port on replication server                          | 3306
 replication_user               | STRING               | user on replication server                          |
 replication_ssl                | [SSL_OPT](#sslopt)   | SSL behavior for replication cx cx                  | DISABLED
-replication_heartbeat          | BOOLEAN              | enable replication heartbeats to detect stale connections | DISABLED
 replication_jdbc_options       | STRING               | mysql jdbc connection options for replication server| [DEFAULT_JDBC_OPTS](#jdbcopts)
 &nbsp;
 schema_host                    | STRING               | server to capture schema from.  See [split server roles](#split-server-roles) | *schema-store host*
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -36,7 +36,7 @@ replication_password           | STRING               | password on replication 
 replication_port               | INT                  | port on replication server                          | 3306
 replication_user               | STRING               | user on replication server                          |
 replication_ssl                | [SSL_OPT](#sslopt)   | SSL behavior for replication cx cx                  | DISABLED
-replication_heartbeat          | BOOLEAN              | enable replication heartbeats to detect stale connections | ENABLED
+replication_heartbeat          | BOOLEAN              | enable replication heartbeats to detect stale connections | DISABLED
 replication_jdbc_options       | STRING               | mysql jdbc connection options for replication server| [DEFAULT_JDBC_OPTS](#jdbcopts)
 &nbsp;
 schema_host                    | STRING               | server to capture schema from.  See [split server roles](#split-server-roles) | *schema-store host*
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -46,9 +46,8 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 	private final String maxwellSchemaDatabaseName;
 
 	protected final BinaryLogClient client;
-	private final long binlogHeartbeatInterval = 1000L;
-	private final float binlogHeartbeatIntervalAllowance = 5.0f;
 	private BinlogConnectorEventListener binlogEventListener;
+	private BinlogConnectorLivenessMonitor binlogLivenessMonitor;
 	private final LinkedBlockingDeque<BinlogConnectorEvent> queue = new LinkedBlockingDeque<>(20);
 	private final TableCache tableCache;
 	private final Scripting scripting;
@@ -151,7 +150,9 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 		 */
 		this.client.setKeepAlive(false);
 		if (mysqlConfig.enableHeartbeat) {
-			this.client.setHeartbeatInterval(this.binlogHeartbeatInterval);
+			this.binlogLivenessMonitor = new BinlogConnectorLivenessMonitor(client);
+			this.client.registerLifecycleListener(this.binlogLivenessMonitor);
+			this.client.registerEventListener(this.binlogLivenessMonitor);
 		}
 
 		EventDeserializer eventDeserializer = new EventDeserializer();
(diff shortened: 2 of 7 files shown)


lukestephenson-zendesk (approved):
Seems ok.

osheroff:
I like this approach, but do keep in mind that this is theory-based programming: losing the server CX and not noticing is a *possible* cause but perhaps not the only one;  if you can catch a node red-handed please try to get a `jstack` dump of it to confirm. 

Regardless, turning on master heartbeating seems like a pretty good idea anyway.  I'll put code-specific comments inline.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -41,11 +41,14 @@
...
 	public static final int BAD_BINLOG_ERROR_CODE = 1236;
 	public static final int ACCESS_DENIED_ERROR_CODE = 1227;
+	public static final int HEARTBEAT_FAILURE_ERROR_CODE = 1623;
 
 	private final String clientID;
 	private final String maxwellSchemaDatabaseName;
 
 	protected final BinaryLogClient client;
+	private final long heartbeatInterval = 5000L; // TODO configure
+	private final float heartbeatIntervalAllowance = 3.0f; // TODO configure, find a better name?
```
osheroff, line 51: I'd say let's not configure these, but instead just have a boolean flag that turns on/off the master heartbeating feature.  1 second  Interval and 5 second allowance seem fine?

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -249,6 +249,16 @@ private void checkCommErrors() throws ServerException {
...
+		long lastEventSeenAt = this.binlogEventListener.getLastEventSeenAt();
+		long lastEventAge = System.currentTimeMillis() - lastEventSeenAt;
+		long maxAllowedEventAge = Math.round(this.heartbeatInterval * this.heartbeatIntervalAllowance);
+		if (lastEventAge > this.heartbeatInterval * this.heartbeatIntervalAllowance) {
+			throw new ServerException(
+					"Last binlog event seen " + lastEventAge + "ms ago, exceeding " + maxAllowedEventAge + "ms allowance " +
+					"(" + this.heartbeatInterval + " * " + this.heartbeatIntervalAllowance + ")",
+					HEARTBEAT_FAILURE_ERROR_CODE,
+					"HY000");
+		}
```
osheroff, line 261: put in a separate function plz.

osheroff: I think that we should make an effort to reconnect rather than simply crashing.  see the logic in `ensureReplicatorThread` -- I'm sure we can hook into that code path and trigger a reconnect.

osheroff (changes requested):

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
timbertson-zd: Sure, less work for me 👍

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
timbertson-zd: I'll give it a go, but I reserve the right to an "I Told You So" if this ends up detecting an issue but unable to reconnect because computers are hard 😉

timbertson-zd:
> if you can catch a node red-handed please try to get a jstack dump of it to confirm.

Yeah, we're working on that for next time. Definitely agree that it's not necessarily the cause but it hopefully eliminates one possible cause. And it'll be useful data to see how often this triggers, once we deploy it.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
timbertson-zd: One thing to point out: the implementation treats "construction of this java object" as the first heartbeat, so it incidentally includes connection time in the first heartbeat. But we hardcode connection timeout to 5s as well anyway, so this shouldn't be an issue.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
osheroff: you can have an I-told-you-so right now, for all i care.  they're pretty cheap in this day and age.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
osheroff: good to know you're thinking of the corner cases.  I too was wondering if there were any places where we might timeout inelegantly waiting for the connection to come up or something. 

what does happen with a timing out server, I wonder.

timbertson-zd:
OK, updated with just a `replication_heartbeat` boolean to turn it on/off, it's enabled by default.

docs/docs/config.md
```diff
@@ -36,6 +36,7 @@ replication_password           | STRING               | password on replication
 replication_port               | INT                  | port on replication server                          | 3306
 replication_user               | STRING               | user on replication server                          |
 replication_ssl                | [SSL_OPT](#sslopt)   | SSL behavior for replication cx cx                  | DISABLED
+replication_heartbeat          | BOOLEAN              | enable replication heartbeats to detect stale connections | ENABLED
```
osheroff, line 39: let's leave this disabled by default for the world until Zendesk runs the code through its paces with it off.  Happy to enable it by default in a future release once it has proved itself.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -251,6 +252,30 @@ private void checkCommErrors() throws ServerException {
...
+	 * Returns true if connected with a recent event.
+	 * If binlog heartbeats are disabled, just returns
+	 * whether there is a connection.
+	 */
+	private boolean isConnectionAlive() {
+		if (!isConnected) {
+			return false;
+		}
+		boolean isAlive = true;
+		if (client.getHeartbeatInterval() > 0L) {
```
osheroff, line 265: would be nice to check `enableHeartbeat` here instead.

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -251,6 +252,30 @@ private void checkCommErrors() throws ServerException {
...
+			long lastEventSeenAt = binlogEventListener.getLastEventSeenAt();
+			long lastEventAge = System.currentTimeMillis() - lastEventSeenAt;
+			long maxAllowedEventAge = Math.round(binlogHeartbeatInterval * binlogHeartbeatIntervalAllowance);
+			isAlive = lastEventAge <= maxAllowedEventAge;
+			if (!isAlive) {
+				LOGGER.warn(
+					"Last binlog event seen " + lastEventAge + "ms ago, exceeding " + maxAllowedEventAge + "ms allowance " +
+					"(" + binlogHeartbeatInterval + " * " + binlogHeartbeatIntervalAllowance + ")");
+			}
```
osheroff, line 274: have you managed to trigger this codepath (and the ensuing reconnection logic) locally?  Like by say, turning on maxwell's expectation of heartbeats but turning off the actual heartbeat mechanism? 

I don't necessarily expect you to write an integration test here, probably too hard, but I'd like to know that you at least reproduced a synthetic run of the code.

osheroff:
generally looking good.

timbertson-zd:
>  I too was wondering if there were any places where we might timeout inelegantly waiting for the connection to come up or something.

Actually, if we're reconnecting instead of dying, there's two edge cases - initial connection, and then on subsequent connection we'll still be remembering the last heartbeat (which is by definition too old 💥 ).

We really want to reset the lastEventSeenAt in `onConnect`. So I've broken it out into its own listener in a29cdaaedae9c717ba10104295cc4e1c424eeee1, which can be both a Lifecycle and Event listener. The current logic was split between Replicator and EventListener, which was already a bit awkward.

(and that way "are heartbeats enabled" equates to "is the heartbeat monitor non-null", which feels cleaner)

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
timbertson-zd: Yeah, I'll hack up something locally once we're happy with the PR and make sure it hangs together 👍

src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
timbertson-zd: It didn't seem worth adding a permanent reference to `mysqlConfig` which is otherwise only available in the constructor. But see below, I've changed it now.

osheroff:
> We really want to reset the lastEventSeenAt in onConnect. So I've broken it out into its own listener in a29cdaa, which can be both a Lifecycle and Event listener. 

Great, looks much cleaner.  I'm 👍 on this after you tell me you've run a test.

osheroff (approved):

timbertson-zd:
Tested, it (now 😉 ) works as advertised:

With the setHeartbeatInterval commented out, we get:

```
16:13:08,932 WARN  BinlogConnectorLivenessMonitor - Last binlog event seen 5027ms ago, exceeding 5000ms allowance (5000 * 1.0)
16:13:08,932 INFO  BinlogConnectorReplicator - Binlog disconnected.
16:13:08,934 WARN  BinlogConnectorReplicator - replicator stopped at position: master.000043:19424584 -- restarting
16:13:08,944 INFO  BinaryLogClient - Connected to mysql.docker:3306 at master.000043/19424584 (sid:6379, cid:281)
16:13:08,944 INFO  BinlogConnectorReplicator - Binlog connected.
```

(I also adjusted the threshold locally because mysql was too chatty)

I had to add a disconnect() before reconnecting, since now the replicator might still think it's connected and raise the already connected exception. The implementation looks idempotent, and it didn't complain when I ran `disconnect()` a few times in a row so I think that's safe.

I also moved the config off `replicator_heartbeat` into `binlog_heartbeat`, and it gets used for all 3 mysql configs. It's a replication setting, but it shouldn't need to be configured separately if you're using split server roles.

timbertson-zd:
I know it's been a while, but I finally remembered to check back for some historical data, and we've seen this log message 4x over the past month. We also haven't seen any recurrence of the initial symptoms (stuck instance) since we rolled this out in February, so it definitely seems to help.

PR #1646, opened 2021-02-04
feat: Nats producer for maxwell

vhnguyenae:
Hello zendesk friends,
Being desperate to use maxwell with Nats :smile: , we try to add a new producer for it.
We've been able to compile and test our solution locally.
Please let us know if you have any feedback or update requirement for the code.
We hope that you guys can integrate it soon so other users can benefit it as well.
Cheers,

### DIFF
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -227,9 +227,8 @@ To produce messages to Nats, you can have the configuration specify in `config.p
 The configurable properties are:
 
 - `nats_url` - defaults to **localhost:4222**
-- `nats_subject_prefix` - defaults to **maxwell**, can be specified as empty '' if you don't want any
-- `nats_subject_hierarchies` - defaults to **%db%.%table%.%type%**
-  - This config controls the routing key, where `%db%` and `%table%` and `%type%`are placeholders that will be substituted at runtime
+- `nats_subject` - defaults to **%{database}.%{table}**
+  - This interpolated string config controls the subject hierarchies, where `%{database}` and `%{table}` are placeholders that will be substituted at runtime
 
 # Google Cloud Pub/Sub
 ***
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -228,7 +228,7 @@ The configurable properties are:
 
 - `nats_url` - defaults to **localhost:4222**
 - `nats_subject` - defaults to **%{database}.%{table}**
-  - This interpolated string config controls the subject hierarchies, where `%{database}` and `%{table}` are placeholders that will be substituted at runtime
+  - This interpolated string config controls the subject hierarchies, where `%{database}` and `%{table}` and `%{type}` are placeholders that will be substituted at runtime. All non alphanumeric characters, all dots, all whitespaces, contained in the database, table or type will be replaced by underscores.
 
 # Google Cloud Pub/Sub
 ***
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -220,6 +220,16 @@ Set the output queue in the `config.properties` by setting the `sqs_queue_uri` p
 
 The producer uses the [AWS SQS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/AmazonSQSClient.html).
 
+# Nats
+***
+To produce messages to Nats, you can have the configuration specify in `config.properties`.
+
+The configurable properties are:
+
+- `nats_url` - defaults to **localhost:4222**
+- `nats_subject_prefix` - defaults to **maxwell**, can be specified as empty '' if you don't want any
+- `nats_subject_hierarchies` - defaults to **%db%.%table%.%type%**
+  - This config controls the routing key, where `%db%` and `%table%` and `%type%`are placeholders that will be substituted at runtime
 
 # Google Cloud Pub/Sub
 ***
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -226,7 +226,7 @@ To produce messages to Nats, you can have the configuration specify in `config.p
 
 The configurable properties are:
 
-- `nats_url` - defaults to **localhost:4222**
+- `nats_url` - defaults to **nats://localhost:4222**
 - `nats_subject` - defaults to **%{database}.%{table}**
   - This interpolated string config controls the subject hierarchies, where `%{database}` and `%{table}` and `%{type}` are placeholders that will be substituted at runtime. All non alphanumeric characters, all dots, all whitespaces, contained in the database, table or type will be replaced by underscores.
 
--- a/src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
@@ -22,18 +22,15 @@ public class NatsProducer extends AbstractProducer {
 
 	public NatsProducer(MaxwellContext context) {
 		super(context);
+		List<String> urls = Arrays.asList(context.getConfig().natsUrl.split(","));
+		Options.Builder optionBuilder = new Options.Builder();
+		urls.forEach(optionBuilder::server);
+		Options option = optionBuilder.build();
+
+		this.natsSubjectTemplate = context.getConfig().natsSubject;
+
 		try {
-			List<String> urls = Arrays.asList(context.getConfig().natsUrl.split(","));
-
-			Options.Builder optionBuilder = new Options.Builder();
-
-			urls.forEach(optionBuilder::server);
-
-			Options option = optionBuilder.build();
-
 			this.natsConnection = Nats.connect(option);
-			this.natsSubjectTemplate = context.getConfig().natsSubject;
-
 		} catch (IOException | InterruptedException e) {
 			throw new RuntimeException(e);
 		}
@@ -47,7 +44,7 @@ public class NatsProducer extends AbstractProducer {
 		}
 
 		String value = r.toJSON(outputConfig);
-		String natsSubject = new TopicInterpolator(this.natsSubjectTemplate).generateFromRowMapAndTrimAllWhitesSpaces(r);
+		String natsSubject = new TopicInterpolator(this.natsSubjectTemplate).generateFromRowMapAndCleanUpIllegalCharacters(r);
 
 		long maxPayloadSize = natsConnection.getMaxPayload();
 		byte[] messageBytes = value.getBytes(StandardCharsets.UTF_8);
--- a/src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
@@ -2,7 +2,7 @@ package com.zendesk.maxwell.producer;
 
 import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.row.RowMap;
-import com.zendesk.maxwell.util.InterpolatedStringsHandler;
+import com.zendesk.maxwell.util.TopicInterpolator;
 import io.nats.client.Connection;
 import io.nats.client.Nats;
 import io.nats.client.Options;
@@ -47,7 +47,7 @@ public class NatsProducer extends AbstractProducer {
 		}
 
 		String value = r.toJSON(outputConfig);
-		String natsSubject = new InterpolatedStringsHandler(this.natsSubjectTemplate).generateFromRowMapAndTrimAllWhitesSpaces(r);
+		String natsSubject = new TopicInterpolator(this.natsSubjectTemplate).generateFromRowMapAndTrimAllWhitesSpaces(r);
 
 		long maxPayloadSize = natsConnection.getMaxPayload();
 		byte[] messageBytes = value.getBytes(StandardCharsets.UTF_8);
--- a/src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
+++ b/src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
@@ -12,11 +12,18 @@ import java.io.IOException;
 public class NatsProducer extends AbstractProducer {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(NatsProducer.class);
-	private Connection natsConnection;
+	private final Connection natsConnection;
+	private final String natsSubjectPrefix;
+	private final String subjectHierarchiesTemplate;
+
 	public NatsProducer(MaxwellContext context) {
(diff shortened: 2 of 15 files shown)


osheroff:
thanks for the PR!  Generally this looks like fairly thorough work, so thanks for that. 

Let's do high-level comments first, and once we've addressed that we can get into smaller stuff.

- The subject-hierarchy code needs more work.  You've got the right approach, passing in a template string, but:

1.  Maxwell already has a format for interpolated strings, it should look like `%{db}.%{table}`, not `%db%.%table%`.   
2. We should put the interpolation code someplace where both the kafka producer and this producer can use it.
3. The default should probably be `%{db}.%{table}`, not `%{db}.%{table}.%{type}`.
4. We should accept any and all subject hierarchy strings.  Remove the checks on line 28 of the producer.
5. Are there other interpolations we want to make available in the subject hierarchy?  %{server_id}?  just thinking.  not totally necessary.
6. We should guard against spaces in table/database names as well as non alphanumberic characters or dots.  NATs is not exactly clear on what's OK in a subject, but we definitely should substitute out spaces and dots.
7.  There's no need for `nats_subject_prefix`.  We only need one string, `nats_subject`, and that can container interpolations.


- should we support connecting to multiple members of a cluster?  https://docs.nats.io/developing-with-nats/connecting/cluster
- authentication?  https://docs.nats.io/developing-with-nats/security/userpass
- does the NATs system have any limits on message size? 

It's ok if the answers to any of these is "I don't know/we'll get to it later."  But if we can think about some ahead of time, great.

When you say:

> We've been able to compile and test our solution locally.

Does that mean you have this running in some kind of production environment?  

Thanks, looking forward to getting another system supported!

-ben

imclem:
Hi @osheroff,

I've co-authored this pull-request with @vhnguyenae.

First of all thanks for your comments.

> 1. Maxwell already has a format for interpolated strings, it should look like `%{db}.%{table}`, not `%db%.%table%`.
> 2. We should put the interpolation code someplace where both the kafka producer and this producer can use it.
> 3. The default should probably be `%{db}.%{table}`, not `%{db}.%{table}.%{type}`.
> 4. We should accept any and all subject hierarchy strings.  Remove the checks on line 28 of the producer.
> 5. Are there other interpolations we want to make available in the subject hierarchy?  %{server_id}?  just thinking.  not totally necessary.
> 6. We should guard against spaces in table/database names as well as non alphanumberic characters or dots.  NATs is not exactly clear on what's OK in a subject, but we definitely should substitute out spaces and dots.
> 7. There's no need for `nats_subject_prefix`.  We only need one string, `nats_subject`, and that can container interpolations.

Totally agree on all the points above, we'll make the changes and make sure that we extract kafka logic to a common place.

> * should we support connecting to multiple members of a cluster? https://docs.nats.io/developing-with-nats/connecting/cluster

I don't know, on our testing/production environment we have a DNS load balancer before our nats nodes which means that when we don't know before hand to which node we connect to. If the node fails, the client will reconnect to one of the ohter nodes of the cluster since all nodes know the routes to each other. More informations [here](https://docs.nats.io/nats-server/configuration/clustering#running-a-simple-cluster).

> * authentication? https://docs.nats.io/developing-with-nats/security/userpass

We use authentication and the credentials can be put directly in the connection string, for example: `nats://user:pass@localhost:4222`, do you want us to update the doc ?

> * does the NATs system have any limits on message size?

Yes and that's something that can be configured server-side, this setting can be obtained from the client. [https://docs.nats.io/developing-with-nats/connecting/misc#get-the-maximum-payload-size](https://docs.nats.io/developing-with-nats/connecting/misc#get-the-maximum-payload-size)
What's your take on this ? Should we check that we won't overflow the message size before publishing a message ? Then what should we do, not send the message ? throw an exception ?

> > We've been able to compile and test our solution locally.
For now it's locally, we will deploy it on our testing environment today and if everything is working as expected to our production environment in the coming days.

Thanks 

Clem

osheroff:
> I don't know, on our testing/production environment we have a DNS load balancer before our nats nodes which means that we don't know before hand to which node we connect to. If the node fails, the client will reconnect to one of the other nodes of the cluster since all nodes and clients know the routes to each other. More informations

True, but one of the possible configs (as far as my reading goes) is that you configure the client with multiple nodes so that if one of them is down the others can be connected to (obviously you don't need this with a load-balanced config, but some people might have it)...

> We use authentication and the credentials can be put directly in the connection string, for example: nats://user:pass@localhost:4222, do you want us to update the doc ?

sure that's fine.

> What's your take on this ? Should we check that we won't overflow the message size before publishing a message ? Then what should we do, not send the message ? throw an exception ?

I think yeah, just log an error message and don't send the message.  None of the producers crash with message-too-large if I remember right.

vhnguyenae:
Hello @osheroff ,
We updated the PR and tackle most of the point that you mentioned before (only the substitution of dots is not done, since Nats uses it for [subject hierarchies](https://docs.nats.io/nats-concepts/subjects#subject-hierarchies).
We also deploy it on our QA environment and so far it's working fine.
Thanks again for the review.
Cheers,

osheroff:
>  (only the substitution of dots is not done, since Nats uses it for subject hierarchies). 

Yes, what I meant is this:  we should substitute dots or other illegal characters (spaces) *before* we interpolate it into the subject string.  This way we can ensure that we don't have extra, unintentional dots or spaces (or other characters, it's unclear whether the subject should be ascii or utf-8 or what) interpolated into the subject string.

src/main/java/com/zendesk/maxwell/util/InterpolatedStringsHandler.java
```diff
@@ -0,0 +1,81 @@
...
+		if (table == null)
+			table = "";
+
+		String type = r.getRowType();
+
+		if (type == null)
+			type = "";
+
+		if (this.isInterpolated)
+			return interpolate(r.getDatabase(), table, type);
```
osheroff, line 60: we may want to check for `database` being null too.  I think in rare cases this can happen.

src/main/java/com/zendesk/maxwell/util/InterpolatedStringsHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * An utility class to interpolate subscription, channel templates.
+ * <p>
+ * An input string can contains any combination of:
+ * - %{db}
+ * - %{table}
+ * - %{type}
+ */
+public class InterpolatedStringsHandler {
```
osheroff, line 14: can we rename this class something a bit better?  `TopicInterpolator` would be fine.

docs/docs/producers.md
```diff
@@ -220,6 +220,15 @@ Set the output queue in the `config.properties` by setting the `sqs_queue_uri` p
...
 
+# Nats
+***
+To produce messages to Nats, you can have the configuration specify in `config.properties`.
+
+The configurable properties are:
+
+- `nats_url` - defaults to **localhost:4222**
+- `nats_subject` - defaults to **%{database}.%{table}**
+  - This interpolated string config controls the subject hierarchies, where `%{database}` and `%{table}` are placeholders that will be substituted at runtime
```
osheroff, line 231: we should mention that `%{type}` can also be interpolated

osheroff:
generally looks good.  will merge after the minor fixes requested

src/main/java/com/zendesk/maxwell/util/InterpolatedStringsHandler.java
vhnguyenae: common method [emptyStringOnNull](https://github.com/GlobalSport/maxwell/pull/3/files#diff-45da18462c44ee750de35aea378ff0075b32a7f16d6c688a3cdbc5bfde4c5e9eR72) is created for this

src/main/java/com/zendesk/maxwell/util/InterpolatedStringsHandler.java
vhnguyenae: updated in https://github.com/zendesk/maxwell/pull/1646/commits/37ff1c30f8e41c6995e46ed0ee248dc2453d4289

docs/docs/producers.md
vhnguyenae: comment is updated https://github.com/GlobalSport/maxwell/pull/3/files#diff-172ffaeef540749bcbd24d92f8ce364cc9e22e1d632d8a11264e8d70c0e85431R231

vhnguyenae:
re @osheroff ,
Thanks for your speedy review, we updated the code according to your comment :-)

docs/docs/producers.md
```diff
@@ -220,6 +220,15 @@ Set the output queue in the `config.properties` by setting the `sqs_queue_uri` p
 
 The producer uses the [AWS SQS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/AmazonSQSClient.html).
 
+# Nats
+***
+To produce messages to Nats, you can have the configuration specify in `config.properties`.
+
+The configurable properties are:
+
+- `nats_url` - defaults to **localhost:4222**
```
vhnguyenae, line 229:
```suggestion
- `nats_url` - defaults to **nats://localhost:4222**
```

src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
```diff
@@ -0,0 +1,68 @@
...
+			Options.Builder optionBuilder = new Options.Builder();
+
+			urls.forEach(optionBuilder::server);
+
+			Options option = optionBuilder.build();
+
+			this.natsConnection = Nats.connect(option);
+			this.natsSubjectTemplate = context.getConfig().natsSubject;
+
+		} catch (IOException | InterruptedException e) {
```
osheroff, line 37: can you limit the `try {} catch() ` block to only surround the parts of the code that might throw the exception?  I Imagine that only line 34 might throw exceptions?

src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
```diff
@@ -0,0 +1,68 @@
...
+	public void push(RowMap r) throws Exception {
+		if (!r.shouldOutput(outputConfig)) {
+			context.setPosition(r.getNextPosition());
+			return;
+		}
+
+		String value = r.toJSON(outputConfig);
+		String natsSubject = new TopicInterpolator(this.natsSubjectTemplate).generateFromRowMapAndTrimAllWhitesSpaces(r);
+
+		long maxPayloadSize = natsConnection.getMaxPayload();
```
osheroff, line 52: have you looked at the source for this method?  Are you sure that `getMaxPayload()` doesn't make a network request?

src/main/java/com/zendesk/maxwell/util/TopicInterpolator.java
```diff
@@ -0,0 +1,89 @@
...
+
+	protected String interpolate(final String database, final String table, final String type) {
+		if (this.isInterpolated) {
+			final String typeReplacement = type != null ? type : "";
+
+
+			return this.inputString
+					.replaceAll("%\\{database\\}", cleanupIllegalCharacters(cleanupIllegalCharacters(emptyStringOnNull(database))))
+					.replaceAll("%\\{table\\}", cleanupIllegalCharacters(cleanupIllegalCharacters(emptyStringOnNull(table))))
+					.replaceAll("%\\{type\\}", cleanupIllegalCharacters(cleanupIllegalCharacters(emptyStringOnNull(typeReplacement))));
```
osheroff, line 66: - shouldn't call `cleanupIllegalCharacters` twice
- please only do this cleanup for "nats" -- kafka topic rules are different and I don't want to change behavior there.

src/main/java/com/zendesk/maxwell/util/TopicInterpolator.java
```diff
@@ -0,0 +1,89 @@
...
+		if (value == null) {
+			return "";
+		} else {
+			return value;
+		}
+	}
+
+	private String cleanupIllegalCharacters(final String value) {
+		return value
+				.replaceAll("([^A-Za-z0-9]+|[\\.\\s]+)", "_");
```
osheroff, line 82: do you need the second part of the regex?  The first part says "anything *not* alpha-numeric" -- doesn't that cover the same character range as the second part?

src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
vhnguyenae: done in https://github.com/GlobalSport/maxwell/pull/4/commits/b1111d20a66d9bfad0e83caeb5cefc0c9866811f

vhnguyenae: done in https://github.com/zendesk/maxwell/pull/1646/commits/d6e9b0077b5add7e9385ad05c2b097e63accc208

src/main/java/com/zendesk/maxwell/producer/NatsProducer.java
vhnguyenae: From what I saw in the source code of [getMaxPayload](https://github.com/nats-io/nats.java/blob/master/src/main/java/io/nats/client/impl/NatsConnection.java#L1503), it just return the value of a field from `Connection.serverInfo`, which was already achieved when we created the connection, so no network request for me :)

src/main/java/com/zendesk/maxwell/util/TopicInterpolator.java
vhnguyenae: updated in https://github.com/zendesk/maxwell/pull/1646/commits/d6e9b0077b5add7e9385ad05c2b097e63accc208#diff-45da18462c44ee750de35aea378ff0075b32a7f16d6c688a3cdbc5bfde4c5e9eR78

src/main/java/com/zendesk/maxwell/util/TopicInterpolator.java
vhnguyenae: you are right, we removed the second part https://github.com/zendesk/maxwell/pull/1646/commits/d6e9b0077b5add7e9385ad05c2b097e63accc208#diff-45da18462c44ee750de35aea378ff0075b32a7f16d6c688a3cdbc5bfde4c5e9eR103

osheroff:
i think we're good to go.  I'll look it over once more and then merge.

osheroff:
thanks for sticking it out through the code-review process guys.  Excited to have this feature in maaxwell.

imclem:
> thanks for sticking it out through the code-review process guys. Excited to have this feature in maaxwell.

Thank you! We are also excited to get this released !

PR #1662, opened 2021-02-26
Add stackdriver metrics logging to maxwell using opencensus

DPGrev:
This will add stackdriver logging for logging the maxwell metrics. Unfortunately there is no ready available dropwizard reporter for stackdriver. 

In order to keep the implementation simple I opted for using the opencensus stackdriver exporter. The downside of the solution is that we need to add opencensus as an external dependency. In my opinion this outweighs having to maintain our own dropwizard stackdriver exporter.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <mockito.version>3.7.7</mockito.version>
+    <opencensus.version>0.28.3</opencensus.version>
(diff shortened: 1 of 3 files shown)


osheroff:
seems ok to me.  what's the size of the dependencies?

DPGrev:
- **opencensus-contrib-dropwizard**: 5KB
- **opencensus-api**: 347KB
- **opencensus-impl**: 11KB
- **opencensus-exporter-stats-stackdriver**: 36KB

PR #1667, opened 2021-03-06
Add SNS producer and updated config

robwittman:
AWS SNS topics now support FIFO ordering and strict delivery. This allows maxwell to publish messages to an SNS topic, with optional `table` and `database` message attributes for subscription filtering

https://aws.amazon.com/about-aws/whats-new/2020/10/amazon-sns-introduces-fifo-topics-with-strict-ordering-and-deduplication-of-messages/

---
This change is [Reviewable](https://reviewable.io/reviews/zendesk/maxwell/1667)

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -90,6 +90,13 @@ option                         | argument                            | descripti
 -------------------------------|-------------------------------------| --------------------------------------------------- | -------
 sqs_queue_uri                  | STRING                              | SQS Queue URI |
 
+## sns producer 
+option                         | argument                            | description                                         | default
+-------------------------------|-------------------------------------| --------------------------------------------------- | -------
+sns_topic                      | STRING                              | The SNS topic to publish to. FIFO topics should end with `.fifo` |
+sns_database_attribute         | BOOLEAN                             | Add `database` as an attribute to the SNS message   | false
+sns_table_attribute            | BOOLEAN                             | Add `table` as an attribute to the SNS message      | false
+
 ## nats producer
 option                         | argument                            | description                                         | default
 -------------------------------|-------------------------------------| --------------------------------------------------- | -------
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -94,8 +94,7 @@ sqs_queue_uri                  | STRING                              | SQS Queue
 option                         | argument                            | description                                         | default
 -------------------------------|-------------------------------------| --------------------------------------------------- | -------
 sns_topic                      | STRING                              | The SNS topic to publish to. FIFO topics should end with `.fifo` |
-sns_database_attribute         | BOOLEAN                             | Add `database` as an attribute to the SNS message   | false
-sns_table_attribute            | BOOLEAN                             | Add `table` as an attribute to the SNS message      | false
+sns_attrs                      | STRING                              | Properties to set as attributes on the SNS message  |  
 
 ## nats producer
 option                         | argument                            | description                                         | default
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -220,6 +220,23 @@ Set the output queue in the `config.properties` by setting the `sqs_queue_uri` p
 
 The producer uses the [AWS SQS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/AmazonSQSClient.html).
 
+# SNS
+***
+
+## AWS Credentials
+You will need to obtain an IAM user that has the permission to access the SQS service. The SQS producer also uses [DefaultAWSCredentialsProviderChain](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) to get AWS credentials.
+
+See the [AWS docs](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html) on how to setup the IAM user with the Default Credential Provider Chain.
+
+In case you need to set up a different region also along with credentials then default one, see the [AWS docs](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html#setup-credentials-setting-region).
+
+## Options
+Set the topic arn in the `config.properties` by setting the `sns_topic` property to the topic name. FIFO topics should have a `.fifo` suffix. 
+
+Optionally, you can enable `sns_table_attribute` and/or `sns_database_attribute` values to have maxwell attach `table` and`database` attributes to the message for subscription filtering
+
+The producer uses the [AWS SNS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sns/AmazonSNSClient.html).
+
 # Nats
 ***
 The configurable properties for nats are:
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -233,7 +233,7 @@ In case you need to set up a different region also along with credentials then d
 ## Options
 Set the topic arn in the `config.properties` by setting the `sns_topic` property to the topic name. FIFO topics should have a `.fifo` suffix. 
 
-Optionally, you can enable `sns_table_attribute` and/or `sns_database_attribute` values to have maxwell attach `table` and`database` attributes to the message for subscription filtering
+Optionally, you can enable `sns_table_attribute` and/or `sns_database_attribute` values to have maxwell attach `table` and `database` attributes to the message for subscription filtering
 
 The producer uses the [AWS SNS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sns/AmazonSNSClient.html).
 
--- a/docs/docs/producers.md
+++ b/docs/docs/producers.md
@@ -224,7 +224,7 @@ The producer uses the [AWS SQS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest
 ***
 
 ## AWS Credentials
-You will need to obtain an IAM user that has the permission to access the SQS service. The SQS producer also uses [DefaultAWSCredentialsProviderChain](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) to get AWS credentials.
+You will need to obtain an IAM user that has the permission to access the SNS topic. The SNS producer also uses [DefaultAWSCredentialsProviderChain](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) to get AWS credentials.
 
 See the [AWS docs](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html) on how to setup the IAM user with the Default Credential Provider Chain.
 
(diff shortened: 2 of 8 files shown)


docs/docs/producers.md
```diff
@@ -220,6 +220,23 @@ Set the output queue in the `config.properties` by setting the `sqs_queue_uri` p
...
+You will need to obtain an IAM user that has the permission to access the SNS topic. The SNS producer also uses [DefaultAWSCredentialsProviderChain](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) to get AWS credentials.
+
+See the [AWS docs](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html) on how to setup the IAM user with the Default Credential Provider Chain.
+
+In case you need to set up a different region also along with credentials then default one, see the [AWS docs](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html#setup-credentials-setting-region).
+
+## Options
+Set the topic arn in the `config.properties` by setting the `sns_topic` property to the topic name. FIFO topics should have a `.fifo` suffix. 
+
+Optionally, you can enable `sns_table_attribute` and/or `sns_database_attribute` values to have maxwell attach `table` and`database` attributes to the message for subscription filtering
```
osheroff, line 236: white space issue here near `database`

docs/docs/producers.md
```diff
@@ -220,6 +220,23 @@ Set the output queue in the `config.properties` by setting the `sqs_queue_uri` p
...
+
+The producer uses the [AWS SNS SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sns/AmazonSNSClient.html).
+
```
osheroff, line 239: is there further configuration for the SNS client library possible?  If so it'd be nice to drop hints as to where to find more config info (rather than just pointing them to the top level SDK docs)

docs/docs/config.md
```diff
@@ -90,6 +90,13 @@ option                         | argument                            | descripti
 -------------------------------|-------------------------------------| --------------------------------------------------- | -------
 sqs_queue_uri                  | STRING                              | SQS Queue URI |
 
+## sns producer 
+option                         | argument                            | description                                         | default
+-------------------------------|-------------------------------------| --------------------------------------------------- | -------
+sns_topic                      | STRING                              | The SNS topic to publish to. FIFO topics should end with `.fifo` |
+sns_database_attribute         | BOOLEAN                             | Add `database` as an attribute to the SNS message   | false
+sns_table_attribute            | BOOLEAN                             | Add `table` as an attribute to the SNS message      | false
```
osheroff, line 98: this interface is a little weird.  can you explain what the function of SNS message attributes are?  In other words, are they just metadata, or do they perform some role in uniqueness, sorting, querying, something?

osheroff:
code looks pretty good for a first pass, but there's no command line flag to actually enable the producer.... if this is a WIP do let me know.

thanks!

docs/docs/config.md
robwittman: The `database` and `table` (and possibly other) MessageAttributes allow consumers of the SNS topic to filter which messages they receive from SNS. Using this, multiple different handlers could listen to a single maxwell stream through SNS, and only receive events for databases / tables they are interested in

https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html#example-filter-policies

docs/docs/producers.md
robwittman: fixed

docs/docs/producers.md
robwittman: First, I'll admit it was a copy - paste from the SQS producer, as they are fairly similar. :/

Regarding configuration, I'm not sure there's a whole lot of "configuration" on the client library itself. It pulls credentials from the environment, then you give it a topic and message for the `publishAsync()` call. I'll poke around to see if I can find something more appropriate

robwittman:
Not totally a WIP, but I do still want to run some manual tests around this to make sure it works 

Also, added the flag for SNS :)

docs/docs/config.md
robwittman: Follow up question to the above. Would it make more sense to maybe use 
```
sns_partition_by_table BOOLEAN 
sns_partition_by_database BOOLEAN
```
to reference these? I don't know if that makes sense, as you could partition by both fields at the same time, but it uses similar syntax to the other producers that support partitioning

osheroff: I was thinking that we might want to support other attributes copied into the record at some point in time. 

Maybe we're not gonna need it, but I was thinking the interface could be like:

```
--sns_attrs=database,table,xid,ts
```

with a default of `database,table`

docs/docs/producers.md
osheroff: If there's nothing more to configure, no big deal.

osheroff:
> Also, added the flag for SNS :)

cool, please also update the docs surrounding the `--producer` flag.

osheroff:
After we resolve around the interface for attributes, I can either:

1.  pull this in and mark it as alpha-quality code in the release, or
2. wait until you've had a chance to hammer some of ht issues in production before we release it to the general world.

I'd obviously prefer option (2) but it's not that big of a deal either way.

docs/docs/config.md
robwittman: My only hang up with that, after re-reading some of the documentation, is that the message attributes need to be specified as a specific type of value. https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html 

So we would have to split on the string, then also operate on each field differently, depending on the datatype. If we want the flexibility for that, happy to implement. 

I also don't know how effective some of the other fields are, at least as far as it relates to subscription filtering. There may be other uses I'm not aware of though

robwittman:
I'm good with option 2 as well. I'll get something setup and verify this all actually works in the wild

docs/docs/config.md
osheroff: so attributes are preconfigured in like the AWS console?  Or they're determined at runtime? 

> So we would have to split on the string, then also operate on each field differently, depending on the datatype.

Yeah, I think that's fine?  It's maybe a big uglier in maxwell code but it appears cleaner to the end user.  Or if you've got a better idea I'm all ears, I just want to avoid having to add a new cmdline flag for each attribute type.

> I also don't know how effective some of the other fields are, at least as far as it relates to subscription filtering.

I think it's fine if we only support database/table/ts for now (those are the hyper obvious ones), as long as our interface leaves some expansion room down the line. 

> There may be other uses I'm not aware of though.

In my experience at *some* point if enough people adopt the code someone will want to be doing shit like extracting data out of a row and passing it as a message attribute.  but that's still hypothetical for now.

robwittman: Attributes configuration has been updated to use `sns_attrs` instead, and the respective documentation was updated as well.

It might be my minimal Java experience, but I was having a real hard time accessing the `RowMap` properties dynamically. I met somewhere in the middle, and allow only `database` and `table` to be added, but the producer can be updated to handle more values in `sns_attrs` as needed. 

It would support passing some of the `data` / `old_data` dynamically as attributes, but that would probably require some large scope changes to the `RowMap`. Thoughts?

robwittman:
Glad I tested the full end-to-end integration with SNS / SQS, I found a couple small issues I resolved to get the producer working. Running the producer against a super basic table resulted in these messages in SQS (which is subscribed to the SNS topic)

```
### Without using sns_attrs
{
  "Type": "Notification",
  "MessageId": "ae40f7a1-ae15-5848-9aa9-af66a620b7b9",
  "SequenceNumber": "10000000000000000005",
  "TopicArn": "arn:aws:sns:us-east-2:<redacted>:maxwell-test-topic.fifo",
  "Message": "{\"database\":\"application\",\"table\":\"entries\",\"type\":\"insert\",\"ts\":1615660533,\"xid\":3209,\"commit\":true,\"data\":{\"id\":8,\"first_name\":\"testing7\"}}",
  "Timestamp": "2021-03-13T18:35:33.948Z",
  "UnsubscribeURL": "https://sns.us-east-2.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-2:<redacted>:maxwell-test-topic.fifo:ed100d31-9761-4621-b7df-562e227720b0"
}

## with sns_attrs
{
  "Type": "Notification",
  "MessageId": "a2a9b9f1-807c-5d3f-83c8-e13689cca90d",
  "SequenceNumber": "10000000000000000009",
  "TopicArn": "arn:aws:sns:us-east-2:<redacted>:maxwell-test-topic.fifo",
  "Message": "{\"database\":\"application\",\"table\":\"entries\",\"type\":\"insert\",\"ts\":1615660912,\"xid\":4591,\"commit\":true,\"data\":{\"id\":12,\"first_name\":\"testing10\"}}",
  "Timestamp": "2021-03-13T18:41:52.896Z",
  "UnsubscribeURL": "https://sns.us-east-2.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-2:<redacted>:maxwell-test-topic.fifo:ed100d31-9761-4621-b7df-562e227720b0",
  "MessageAttributes": {
    "database": {
      "Type": "String",
      "Value": "application"
    },
    "table": {
      "Type": "String",
      "Value": "entries"
    }
  }
}
```

All in all, this looks as I would expect it to.

src/main/java/com/zendesk/maxwell/producer/MaxwellSNSProducer.java
```diff
@@ -0,0 +1,112 @@
...
+						break;
+					case "table":
+						messageAttributes.put(
+							"table",
+							new MessageAttributeValue().withDataType("String").withStringValue(r.getTable())
+						);
+						break;
+				}
+			}
+		}
```
osheroff, line 62: i think this is probably fine but some purist-engineer part of me recoils and says you should parse the string once in the initializer and then compile down to enums and oh god we'll never get it merged

src/main/java/com/zendesk/maxwell/producer/MaxwellSNSProducer.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private AmazonSNSAsync client;
+	private String topic;
+	private String[] stringFelds = {"database", "table"};
+	private String[] numberFields = {"ts", "xid"};
+
+	public MaxwellSNSProducer(MaxwellContext context, String topic) {
+		super(context);
+		this.topic = topic;
+		System.out.println(topic);
```
osheroff, line 29: stray code line

osheroff:
99% there

src/main/java/com/zendesk/maxwell/producer/MaxwellSNSProducer.java
robwittman: :facepalm: removed

src/main/java/com/zendesk/maxwell/producer/MaxwellSNSProducer.java
robwittman: I thought that as well. I opted to keep most of the logic inside of the SNS producer for now, as that's what mostly would be in charge of it. If we started allowing specific `data` fields to be mapped for example, classes upstream would need to know how to handle something like 
```
sns_attrs=database,table,sometable.somefield:String,anothertable.otherfield:Boolean` (for type casting the attributes)
```
or whatever, depending on what those interfaces actually look like when they get implemented. Or alternatively, some logic would live in the config, some would still live in the producer. So I'm not totally sure what the best strategy would be

The current usage isn't something I'd brag about writing, but I agree, its probably fine.

Issue #1615, opened 2020-12-08
table columns process doesn't need to restart service


zhongyu211:
Hi,
we have already deployment a lot of maxwell serverless service on ecs, and we always need to do the  common column process, such as suppress and get substring of the column we need to change on filter.js script and restart the service, do we have any solution to avoid this scenario ?

osheroff:
there currently isn't a solution for reconfiguring maxwell without a restart.  I'll leave this issue open as a feature request.

PR #1670, opened 2021-03-20
[Ready for Review] Proposal for allowing live updates to maxwell filter

crispmark:
We are also having issues with wanting to quickly turn on and off maxwell filters. I saw with issue #1615 that we were not the only ones having this issue. My proposal is pretty simple: expose an http endpoint `PATCH: /config` that would allow a user to supply a json containing the new filter, and update it live.

```JSON
{
  "filter":"exclude: dynamic_db.*"
}
```

I have some sample code here with it working, @osheroff please let me know if this is an avenue that you think we could pursue, and I will add tests and move the servlet outside of the metricsRegistries group.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/monitoring/MaxwellConfigServlet.java
+++ b/src/main/java/com/zendesk/maxwell/monitoring/MaxwellConfigServlet.java
@@ -6,6 +6,9 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.zendesk.maxwell.MaxwellContext;
 import com.zendesk.maxwell.filtering.InvalidFilterException;
 
@@ -14,10 +17,9 @@ import java.io.PrintWriter;
 import java.util.stream.Collectors;
 
 /**
- * An HTTP servlets which allows for the live reconfiguration of maxwell.
+ * An HTTP servlet which allows for the live reconfiguration of maxwell.
  */
 public class MaxwellConfigServlet extends HttpServlet {
-  private static final long serialVersionUID = 3772654177231086757L;
   private static final ObjectMapper mapper = new ObjectMapper();
   private final MaxwellContext context;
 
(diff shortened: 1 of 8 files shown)


osheroff:
'sok by me

crispmark:
@osheroff i have added some tests as well as a new config option (http_config) to support this. The endpoint does very basic validation and will return an error in the format:
```JSON
{
  "error": {
    "code": 400,
    "message": "invalid filter"
  }
}
```

both PATCH and GET endpoints are exposed on `/config` and return the serialized config (right now only filter, but can be expanded if new support for config updates is added).
```
{
  "filter": "include: *.*"
}
```
The filter is only updated if the filter string is a non-null valid filter and when the filter is updated the new filter is logged.

Let me know if this looks good or if there are any changes you want to make!

crispmark:
@osheroff could you take a look to see if any major contract changes are needed?

src/main/java/com/zendesk/maxwell/monitoring/MaxwellConfigServlet.java
```diff
@@ -0,0 +1,118 @@
...
+    resp.setStatus(HttpServletResponse.SC_OK);
+    resp.setContentType(CONTENT_TYPE);
+      try (PrintWriter writer = resp.getWriter()) {
+        writer.println(getSerializedConfig());
+      }
+  }
+
+  @Override
+  public void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+    if (request.getMethod().equalsIgnoreCase("PATCH")) {
```
osheroff, line 99: I think we should probably support PUT/POST too.  I don't mind if the semantics aren't precisely correct, I'd rather it follow principle of least surprise.

osheroff:
maxwell uses tabs not spaces, one of the new files introduced had spaces

osheroff:
overall this looks ok.  I could see an argument for a richer representation of the filter in JSON as an array of strings (or even an array of objects, eg `[ { "exclude": { "database": "aa", "table": "bb" } }]`.  Otherwise consumers of the GET endpoint may have to implement their own filter parser.  

could also see an argument for accepting new filter configs in said JSON format *or* as strings.

Don't think that's a blocker to merging this as-is though.

osheroff:
if there's no filter, it should return
`{ filter: null }`
instead of what it's doing which is `{filter: ""}`

osheroff:
needs a scootch of documentation written.

crispmark:
@osheroff i have updated to make the following changes
- fixed the indentation on the servlet file
- added put and post handlers for updating the config
- added some documentation for http_config variable
- returns null for filter when filterstring is ""

osheroff:
ok.  merging.  thanks!

PR #1691, opened 2021-05-07
Allow access to the primary key columns, values as array or map

recurlybrad:
We are using a custom file producer and would like to have direct access to the primary keys/values for a RowMap.


osheroff:
would like a couple of tests around it.  Otherwise fine.

recurlybrad:
Great, thank you, we will update with tests.

recurlybrad:
@osheroff,
Thank you for taking the time to review this request.  I have added unit tests for the new methods on RowMap.  If there are other tests that we should consider adding, we'd be happy to include more.

PR #1696, opened 2021-05-17
update Dockerfile to support ARM

charlotte-xiao:
This Dokcerfile can support ARM and AMD64

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,17 @@
-FROM maven:3.6-jdk-11
-ENV MAXWELL_VERSION=1.33.0 KAFKA_VERSION=1.0.0
+#Docker image of MaxWell Can Support Amd64 and Arm64/v8
+FROM openjdk:11-slim
+#FROM openjdk:11
+ENV MAXWELL_VERSION=1.33.0
+ENV MAXWELL_FILE maxwell-1.33.0.tar.gz
+ENV MAXWELL_DOWNLOAD_PATH "https://github.com/zendesk/maxwell/releases/download/v1.33.0/maxwell-1.33.0.tar.gz"
 
-RUN apt-get update \
-    && apt-get -y upgrade \
-    && apt-get install -y make
-
-# prime so we can have a cached image of the maven deps
-COPY pom.xml /tmp
-RUN cd /tmp && mvn dependency:resolve
-
-COPY . /workspace
-RUN cd /workspace \
-    && KAFKA_VERSION=$KAFKA_VERSION make package MAXWELL_VERSION=$MAXWELL_VERSION \
-    && mkdir /app \
-    && mv /workspace/target/maxwell-$MAXWELL_VERSION/maxwell-$MAXWELL_VERSION/* /app/ \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /workspace/ /root/.m2/ \
-    && echo "$MAXWELL_VERSION" > /REVISION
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,23 +1,18 @@
-FROM maven:3.6-jdk-11
-ENV MAXWELL_VERSION=1.33.0 KAFKA_VERSION=1.0.0
+#Docker image of MaxWell Can Support Amd64 and Arm64/v8
+FROM openjdk:11-slim
+#FROM openjdk:11
+ENV MAXWELL_VERSION=1.33.0
+ENV MAXWELL_FILE maxwell-1.33.0.tar.gz
+ENV MAXWELL_DOWNLOAD_PATH "https://github.com/zendesk/maxwell/releases/download/v1.33.0/maxwell-1.33.0.tar.gz"
```
redcape, line 6: This creates a circular dependency. This would mean we need to make a public release in order to deploy docker. This doesn't allow for iteration / changes in a local environment like the current one does.

charlotte-xiao: Yes. But I think there is a another idea to resolve the docker support Arm64.  That is build a maven image to support Arm64 and then use the old dockefile. But I think it will be little complex.

redcape: It looks like maven has an openjdk-11-slim version: https://github.com/carlossg/docker-maven/blob/bdffb5117c33476d554325d8efe5866306004b99/openjdk-11-slim/Dockerfile.

https://hub.docker.com/_/maven?tab=description:
```
3.8.1-jdk-11-slim, 3.8-jdk-11-slim, 3-jdk-11-slim, 3.8.1-openjdk-11-slim, 3.8-openjdk-11-slim, 3-openjdk-11-slim
```


Does that work?

redcape: It looks like arm64 with the regular non-slim version works for arm64v8 too.

https://hub.docker.com/r/arm64v8/maven/

Looking at the git history, arm84v8 has been supported since 3.6.1.

https://github.com/docker-library/official-images/commit/ed6120b4d885ace6fb7d400021371bfcbcbe40c0#diff-79f56c4595d485f2c306310171f04dcd7d027f418e1a6bee7186dedd2fbedb7c

osheroff:
@redcape what's your opinion here?  I can't tell if this is a PR worth going deeper on or just snoozing.

I'm also a bit unclear why we need to change the fundamental method of building just to support arm

redcape:
> @redcape what's your opinion here? I can't tell if this is a PR worth going deeper on or just snoozing.
> 
> I'm also a bit unclear why we need to change the fundamental method of building just to support arm

@osheroff I'm not sure the exact steps because I'm not familiar with setting up docker hub, but you need make it so when publishing a new docker hub image, it publishes an output for linux/arm64/v8 in addition to linux/amd64. Here's a link on [multi-arch support in docker](https://docs.docker.com/desktop/multi-arch/). It seems like you need to use the `--platform linux/amd64,linux/arm64/v8` option.

For what it's worth, I was able to build the arm64/amd64 versions on an amd64 machine running Ubuntu 18.04 / Docker version 20.10.5, build 55c4c88:
```
docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
docker buildx create --driver docker-container --use
docker buildx inspect --bootstrap
docker buildx build --platform linux/amd64,linux/arm64/v8  -f Dockerfile -t registry.dev.example.com/maxwell/maxwell:1.33.0-ex
```

The first three commands came form [here](https://github.com/docker/buildx/issues/495).

osheroff:
this was fixed with some github actions.  for better or for worse.

PR #1698, opened 2021-05-21
Add env_config option for config.properties alternative

redcape:
Setting many kafka parameters is not possible using environment variables in Linux because Java is unable to read environment variables with '.' in them. Environment variables are supposed to be alphanumeric with underscores allowed only. Settings like kafka.bootstrap.servers can only be set in the config.properties. Some organizations prefer configuration via environment variables to prevent seeing sensitive data in `ps` or the docker image itself. [Related link](https://github.com/docker-library/openjdk/issues/135), there are also several stackoverflow questions/answers and I spent a lot of time debugging before realizing this was happening.

This change adds in a JSON-encoded environment variable that can be used as an alternative to config.properties. The input is expected to be a JSON object of string keys and string values. With this change, we are able to set all necessary configs in environment variables.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -179,6 +179,11 @@ public class MaxwellConfig extends AbstractConfig {
 		final MaxwellOptionParser parser = new MaxwellOptionParser();
 		parser.accepts( "config", "location of config.properties file" )
 				.withRequiredArg();
+		parser.accepts( "env_config", "json object encoded config in an environment variable, alternative to config option" )
+				.availableUnless("config")
+				.withRequiredArg();
+		parser.accepts( "config_prefix", "prefix of maxwell configuration variables inside of the properties file, case insensitive" )
+				.withRequiredArg();
 
 		parser.separator();
 
@@ -487,12 +492,15 @@ public class MaxwellConfig extends AbstractConfig {
 		MaxwellOptionParser parser = buildOptionParser();
 		OptionSet options = parser.parse(argv);
 
-		Properties properties;
+		final Properties properties;
+		final String configPrefix = options.has("config_prefix") ? (String) options.valueOf("config_prefix") : null;
 
 		if (options.has("config")) {
-			properties = parseFile((String) options.valueOf("config"), true);
+			properties = getPrefixedFilteredProperties(parseFile((String) options.valueOf("config"), true), configPrefix);
+		} else if (options.has("env_config")) {
+			properties = getPrefixedFilteredProperties(readPropertiesEnv((String) options.valueOf("env_config")), configPrefix);
 		} else {
-			properties = parseFile(DEFAULT_CONFIG_FILE, false);
+			properties = getPrefixedFilteredProperties(parseFile(DEFAULT_CONFIG_FILE, false), configPrefix);
 		}
 
 		String envConfigPrefix = fetchStringOption("env_config_prefix", options, properties, null);
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -182,8 +182,6 @@ public class MaxwellConfig extends AbstractConfig {
 		parser.accepts( "env_config", "json object encoded config in an environment variable, alternative to config option" )
 				.availableUnless("config")
 				.withRequiredArg();
-		parser.accepts( "config_prefix", "prefix of maxwell configuration variables inside of the properties file, case insensitive" )
-				.withRequiredArg();
 
 		parser.separator();
 
@@ -493,14 +491,13 @@ public class MaxwellConfig extends AbstractConfig {
 		OptionSet options = parser.parse(argv);
 
 		final Properties properties;
-		final String configPrefix = options.has("config_prefix") ? (String) options.valueOf("config_prefix") : null;
 
 		if (options.has("config")) {
-			properties = getPrefixedFilteredProperties(parseFile((String) options.valueOf("config"), true), configPrefix);
+			properties = parseFile((String) options.valueOf("config"), true);
 		} else if (options.has("env_config")) {
-			properties = getPrefixedFilteredProperties(readPropertiesEnv((String) options.valueOf("env_config")), configPrefix);
+			properties = readPropertiesEnv((String) options.valueOf("env_config"));
 		} else {
-			properties = getPrefixedFilteredProperties(parseFile(DEFAULT_CONFIG_FILE, false), configPrefix);
+			properties = parseFile(DEFAULT_CONFIG_FILE, false);
 		}
 
 		String envConfigPrefix = fetchStringOption("env_config_prefix", options, properties, null);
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
```diff
@@ -95,6 +100,46 @@ protected Properties readPropertiesFile(String filename, Boolean abortOnMissing)
...
+						.map(entry -> new AbstractMap.SimpleEntry<>(entry.getKey(), entry.getValue().toString()))
+						.forEach(entry -> properties.put(entry.getKey(), entry.getValue()));
+				return properties;
+			} catch (JsonProcessingException e) {
+				throw new IllegalArgumentException("Unparseable JSON in env variable " + envConfig, e);
+			}
+		} else {
+			System.err.println("No JSON-encoded environment variable named: " + envConfig);
+			System.exit(1);
+			return null; // unreachable
```
osheroff, line 122: throw rather than exit here.

osheroff:
I think I'm good with the JSON ENV thing (although the gyrations gone through to conform to docker's way of things sometimes amuses and infuriates me), but I don't think that the prefix-filter-thing is really maxwell's responsibility, and don't really want to accept that part of this PR unless there's a compelling argument outside of your specific implementation...

redcape:
> I think I'm good with the JSON ENV thing (although the gyrations gone through to conform to docker's way of things sometimes amuses and infuriates me), but I don't think that the prefix-filter-thing is really maxwell's responsibility, and don't really want to accept that part of this PR unless there's a compelling argument outside of your specific implementation...

It's only compelling for use-cases that have a flattened application config, including whatever might be wrapping around maxwell.

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
redcape: fixed

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -487,10 +490,12 @@ private void parse(String [] argv) {
 		MaxwellOptionParser parser = buildOptionParser();
 		OptionSet options = parser.parse(argv);
 
-		Properties properties;
+		final Properties properties;
 
 		if (options.has("config")) {
 			properties = parseFile((String) options.valueOf("config"), true);
+		} else if (options.has("env_config")) {
+			properties = readPropertiesEnv((String) options.valueOf("env_config"));
```
osheroff, line 498: thinking maybe we should also parse the config file?  maybe parse the config file first and then override with the env?  something like that.

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
```diff
@@ -95,6 +100,29 @@ protected Properties readPropertiesFile(String filename, Boolean abortOnMissing)
 		return p;
 	}
 
+	protected Properties readPropertiesEnv(String envConfig) {
+		LOGGER.debug("Attempting to read env_config param: {}", envConfig);
+		String envConfigJsonRaw = System.getenv(envConfig);
+		if (envConfigJsonRaw != null && envConfigJsonRaw.startsWith("{")) {
+			LOGGER.debug("Parsing envConfig");
```
osheroff, line 107: probably more useful to log each key/value pair to the debug log than to say this

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
```diff
@@ -95,6 +100,29 @@ protected Properties readPropertiesFile(String filename, Boolean abortOnMissing)
 		return p;
 	}
 
+	protected Properties readPropertiesEnv(String envConfig) {
+		LOGGER.debug("Attempting to read env_config param: {}", envConfig);
+		String envConfigJsonRaw = System.getenv(envConfig);
+		if (envConfigJsonRaw != null && envConfigJsonRaw.startsWith("{")) {
```
osheroff, line 106: might want to be forgiving of whitespace?

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
```diff
@@ -95,6 +100,29 @@ protected Properties readPropertiesFile(String filename, Boolean abortOnMissing)
...
+			LOGGER.debug("Parsing envConfig");
+			ObjectMapper mapper = new ObjectMapper();
+			try {
+				Map<String, Object> stringMap = mapper.readValue(envConfigJsonRaw, Map.class);
+				Properties properties = new Properties();
+				stringMap.entrySet().stream()
+						.map(entry -> new AbstractMap.SimpleEntry<>(entry.getKey(), entry.getValue().toString()))
+						.forEach(entry -> properties.put(entry.getKey(), entry.getValue()));
```
osheroff, line 114: i can never make heads or tails of this kinda code.  I'm going to trust you that it works.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
redcape: I disagree, making a smaller number of ways to configure makes it easier to reason about. In this case, the user chooses either `config` or `env_config` for configuration and it throws an error if they attempt to choose both. This model is easier for a user trying to configure it in multiple places and needing to understand which configuration is higher in the hierarchy.

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
redcape: will change to the following:
```
	Properties properties = new Properties();
	for (Map.Entry<String, Object> entry : stringMap.entrySet()) {
		properties.put(entry.getKey(), entry.getValue().toString());
	}
	return properties;
```

The point is to allow it to be an Object to be forgiving in the case where a user specifies a Number instead of a String and turn it into a String from there, the same way a properties file would read it.

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
redcape: added `trim()` to cover whitespace including newline, carriage return, spaces, tabs

src/main/java/com/zendesk/maxwell/util/AbstractConfig.java
redcape: I will log the keys, we should avoid logging the values which might contain passwords to the logs, even in debug mode. Went ahead and added this elsewhere too be consistent.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
osheroff: sure, possibly harder to reason about.  but this is how the other configuration methods work; if you're passing in command line flags we still parse `config.properties`.  If you use `env_config_prefix`, we grab what's in the environment and still parse `config.properties`.  None of the other methods are exclusive so this one shouldn't be either.

redcape: I don't entirely agree with the reasoning, but I respect your opinion on project direction as the project owner. Changed to make it additive on top of config.properties.

Ordering of overrides is now `cmd line options > env variable prefix > env config > config file`

osheroff: thanks for that.

PR #1703, opened 2021-06-02
Make all debug statements parameterized

redcape:
Most of the logging in Maxwell is currently done by concatenating strings, however one of the main benefits of using SLF4j is that you can skip the cost of concatenating log message strings and calling the toString method on parameters when you don't log something. This is particularly relevant for debug logging because it is usually turned off. The same should be done for all other logging calls for consistency, but those changes are not part of this PR and provide less benefit as logging for INFO and above is usually on.

There are some places where there is both a isDebugEnabled() check and parameterization, I added isDebugEnabled where it was relevant (like calls to an alternative toString method for debugging purposes) and left it alone when it was already there. While parameterization is  redundant in these cases I think consistency of using parameterization is more important.

Reference: http://www.slf4j.org/faq.html#logging_performance

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellContext.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellContext.java
@@ -231,8 +231,9 @@ public class MaxwellContext {
 					// ignore
 				}
 
-				LOGGER.debug("Shutdown complete: " + shutdownComplete.get());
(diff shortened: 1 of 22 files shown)


osheroff:
okey doke.  probably micro in the grand scheme but the style is cleaner anyway

PR #1716, opened 2021-06-30
SchemaCapturer - minor query optimization and better connection cleanup

redcape:
Query Optimization:
 - Call `isMySQLAtLeast56` only once during initialization instead of at each call to `captureTables` (one call per DB)
 - Save a call to Connection#prepareStatement in `captureDatabase` by retaining the PreparedStatement as a member variable, similar to the other statements. (one call per DB)

Connection cleanup
 - Made SchemaCapturer AutoClosable
 - Used try-with-resources everywhere in the class, and made calls to SchemaCapturer do the same.

For SchemaCapturer, I recommend reviewing the diff in an IDE like IntelliJ, it makes it super-clear what lines are indent-only.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -305,7 +305,7 @@ public class MysqlSavedSchema {
 			MysqlSavedSchema savedSchema = new MysqlSavedSchema(serverID, caseSensitivity);
 
 			savedSchema.restoreFromSchemaID(conn, schemaID);
-			savedSchema.handleVersionUpgrades(conn);
+			savedSchema.handleVersionUpgrades(pool);
 
 			return savedSchema;
 		}
@@ -679,13 +679,13 @@ public class MysqlSavedSchema {
 		int caseDiffs = 0;
 
 		for ( Pair<ColumnDef, ColumnDef> pair : schema.matchColumns(recaptured) ) {
-			ColumnDef cA = pair.getLeft();
-			ColumnDef cB = pair.getRight();
+			ColumnDef schemaCol = pair.getLeft();
+			ColumnDef recapturedCol = pair.getRight();
 
-			if ( !cA.getName().equals(cB.getName()) ) {
-				LOGGER.info("correcting column case of `" + cA.getName() + "` to `" + cB.getName() + "`.  Will save a full schema snapshot after the new DDL update is processed.");
+			if ( !schemaCol.getName().equals(recapturedCol.getName()) ) {
+				LOGGER.info("correcting column case of `" + schemaCol.getName() + "` to `" + recapturedCol.getName() + "`.  Will save a full schema snapshot after the new DDL update is processed.");
 				caseDiffs++;
-				cA.setName(cB.getName());
+				schemaCol.setName(recapturedCol.getName());
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -721,17 +721,23 @@ private void fixColumnLength(Schema recaptured) throws SQLException {
...
 
 			if ( this.schemaVersion < 1 ) {
 				if ( this.schema != null && this.schema.findDatabase("mysql") == null ) {
 					LOGGER.info("Could not find mysql db, adding it to schema");
-					SchemaCapturer sc = new SchemaCapturer(conn, sensitivity, "mysql");
-					Database db = sc.capture().findDatabase("mysql");
-					this.schema.addDatabase(db);
-					this.shouldSnapshotNextSchema = true;
+					try (Connection conn = pool.getConnection();
+						 SchemaCapturer sc = new SchemaCapturer(conn, sensitivity, "mysql")) {
```
osheroff, line 736: whitespace issue here

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
```diff
@@ -261,4 +269,13 @@ private void captureTablesPK(Database db, HashMap<String, Table> tables) throws
...
 	}
 
+	@Override
+	public void close() throws SQLException {
+		try (PreparedStatement p1 = tablePreparedStatement;
+			 PreparedStatement p2 = columnPreparedStatement;
+			 PreparedStatement p3 = pkPreparedStatement) {
+			// auto-close shared prepared statements
+		}
+	}
```
osheroff, line 279: this is a bit goofy, is there not a `PreparedStatement#close` or something?

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -721,17 +721,23 @@ private void fixColumnLength(Schema recaptured) throws SQLException {
 		}
 	}
 
-	protected void handleVersionUpgrades(Connection conn) throws SQLException, InvalidSchemaError {
+	protected void handleVersionUpgrades(ConnectionPool pool) throws SQLException, InvalidSchemaError {
 		if ( this.schemaVersion < 3 ) {
-			Schema recaptured = new SchemaCapturer(conn, sensitivity).capture();
+			final Schema recaptured;
+			try (Connection conn = pool.getConnection();
+				 SchemaCapturer sc = new SchemaCapturer(conn, sensitivity)) {
```
osheroff, line 728: whitespace issue here

osheroff (changes requested):
Hi, only minor changes necessary.  I wish I had had time to look at this before, just to say:

I don't mind doing some code-cleanup while you're going, but if you can, please separate out functional changes from cosmetic (or very minor) changes -- if your PR is like "fixes coding style, moves some log lines around, no functionality changes" then I don't have to turn on the deep inspection brain. 

Similarly, if I can view a PR as just changing functionality then I can focus far more intently on those changes will be able to give much better PR feedback.

src/main/java/com/zendesk/maxwell/schema/SchemaCapturer.java
redcape: There is, but I think the alternative when dealing with multiple resources is weirder:
```
...
finally {
  try {
    pkPreparedStatement.close();
  } finally {
    try {
      columnPreparedStatement.close();
    } finally {
      tablePreparedStatement.close();
    }
  }
}
```

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
redcape: not seeing it, can you be more specific?

redcape:
> please separate out functional changes from cosmetic (or very minor) changes

reverted out-of-place MysqlSavedSchema.java variable renames.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
feng-tao: L736 doesn't line up with L737 and has additional space? it seems to happen in other `SchemaCapturer` as well

PR #1718, opened 2021-07-06
Improve Schema database traversal speed / startup time

redcape:
For diffs, n^2 -> n.
Change Schema's database list to a LinkedHashMap from an ArrayList to greatly improve startup for instances with many databases/tables. Since case sensitivity doesn't change, we can just pre-compute the name and do comparisons against the normalized name. Moved mutations to the database list (add/remove) to `Schema` methods so it's easier to maintain this normalization logic.

Baseline test: 70K DB, 1M Table, 6M columns
Measured Performance: 12min startup to 3 min startup with this change

Startup was seeing very high sustained CPU usage and thread dumping showed the Schema#findDatabase call was taking the majority time to scan the ArrayList.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/Schema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/Schema.java
@@ -6,44 +6,50 @@ import com.zendesk.maxwell.schema.ddl.InvalidSchemaError;
 import org.apache.commons.lang3.tuple.Pair;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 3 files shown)


osheroff:
this is great!  Should we do the same with `Database#findTable` or is that not a noticable part of startup?

redcape:
> this is great! Should we do the same with `Database#findTable` or is that not a noticeable part of startup?

For my use-case, it wasn't noticeable since the table list average is quite low. It could have similar improvements for instances with a large number of databases with a large number of tables.

PR #1719, opened 2021-07-06
Intern ColumnDefs to reduce memory by more than half for multi-tenant DB use-cases

redcape:
This is part of a series of PRs to better support multi-tenant DB use-cases that are typically not served well by most CDC systems. This PR does not include any special modes for multi-tenant, it simply reorganizes the memory/code to make it more performant for this use-case.

Baseline instance: 70K DBa, 1M Tablea, 6M columns
Before/After Change Schema retained size (from heap dump): 350MB/140MB
 - There is opportunity to shrink this down by another ~110MB by moving PK to TableColumnList or otherwise interning that list as well, but this is more complex and not going to be part of this PR.

Make ColumnDefs and TableColumnLists immutable and intern them. This is all about optimizing on the fact that we see the same column and table definitions over and over in multi-tenant DBs. There is a negligible hit on modify performance when DDL changes occur to rebuild these immutable objects. This doesn't hurt existing use-case performance where DDL changes occur infrequently and greatly improves memory usage for multi-tenant use-cases.

**Reviewer notes**
**ColumnDef and subclasses**
To make maintaining immutability easier, made all member variables private and replaced all `setXXX` with `withXXX`. The latter function will clone the instance with the appropriate changed variable. Tests validate that all member variables are private and there are no setters, or at least no methods starting with `set` which should be enough to prevent casual changes from accidentally introducing issues.

Related to this change, I created the `TableColumnList#replace` and `Table#replaceColumn` method since replacing a field is now the common way to modify it.

To allow for easy interning, I created a `create` method on each concrete type and made the constructors private. All columns share a single `Interner` with weakKey mapping, this is why any classes with newly introduced member variables need to implement hashCode/equals (validated by test that the method exists). Since all creations and modifications go through `create` / `withXXX` methods, they all use the Interner to keep the memory footprint small when seeing lots of duplicates fields.

**TableColumnList**
There are some `synchronized` blocks on TableColumnList that I tried to keep the same. Uncontended locks are very cheap in Java. I don't understand why they are there as the methods themselves (indexed add, remove, get) would create race conditions anyway in any concurrent write scenario.

Internally, it keeps an immutable object containing a list of the ordered ColumnDefs. ColumnDefs are immutable, so the only way to change them is to modify the list. Since the list of columns repeats itself constantly in multi-tenant DBs, we intern this list for memory savings (reduce duplicate ArrayList/backing Object[]s). My long-term plan is that the internal reduced memory representations will be able to be represented in a similar compact way in the stored maxwell schema.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -16,7 +16,6 @@ import com.zendesk.maxwell.schema.columndef.*;
 import com.zendesk.maxwell.util.ConnectionPool;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.text.StrTokenizer;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -635,27 +634,29 @@ public class MysqlSavedSchema {
 	private void fixUnsignedColumns(Schema recaptured) throws SQLException, InvalidSchemaError {
 		int unsignedDiffs = 0;
 
-		for ( Pair<ColumnDef, ColumnDef> pair : schema.matchColumns(recaptured) ) {
-			ColumnDef cA = pair.getLeft();
-			ColumnDef cB = pair.getRight();
+		for ( Pair<Schema.FullColumnDef, Schema.FullColumnDef> pair : schema.matchColumns(recaptured) ) {
+			Table schemaTable = pair.getLeft().getTable();
+			ColumnDef schemaCol = pair.getLeft().getColumnDef();
+			ColumnDef recapturedCol = pair.getRight().getColumnDef();
 
-			if (cA instanceof IntColumnDef) {
-				if (cB != null && cB instanceof IntColumnDef) {
-					if (((IntColumnDef) cA).isSigned() && !((IntColumnDef) cB).isSigned()) {
-						((IntColumnDef) cA).setSigned(false);
+			if (schemaCol instanceof IntColumnDef) {
+				if (recapturedCol != null && recapturedCol instanceof IntColumnDef) {
+					if (((IntColumnDef) schemaCol).isSigned() && !((IntColumnDef) recapturedCol).isSigned()) {
+						schemaTable.replaceColumn(schemaCol.getPos(), ((IntColumnDef) schemaCol).withSigned(false));
 						unsignedDiffs++;
 					}
 				} else {
-					LOGGER.warn("warning: Couldn't check for unsigned integer bug on column " + cA.getName() +
+					LOGGER.warn("warning: Couldn't check for unsigned integer bug on column " + schemaCol.getName() +
 						".  You may want to recapture your schema");
 				}
-			} else if (cA instanceof BigIntColumnDef) {
-				if (cB != null && cB instanceof BigIntColumnDef) {
-					if (((BigIntColumnDef) cA).isSigned() && !((BigIntColumnDef) cB).isSigned())
-						((BigIntColumnDef) cA).setSigned(false);
+			} else if (schemaCol instanceof BigIntColumnDef) {
+				if (recapturedCol != null && recapturedCol instanceof BigIntColumnDef) {
+					if (((BigIntColumnDef) schemaCol).isSigned() && !((BigIntColumnDef) recapturedCol).isSigned()) {
+						schemaTable.replaceColumn(schemaCol.getPos(), ((BigIntColumnDef) schemaCol).withSigned(false));
+					}
 					unsignedDiffs++;
 				} else {
-					LOGGER.warn("warning: Couldn't check for unsigned integer bug on column " + cA.getName() +
+					LOGGER.warn("warning: Couldn't check for unsigned integer bug on column " + schemaCol.getName() +
 						".  You may want to recapture your schema");
 				}
 			}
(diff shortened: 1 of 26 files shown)


src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -675,43 +676,42 @@ private void fixUnsignedColumns(Schema recaptured) throws SQLException, InvalidS
 		}
 	}
 
-	private void fixColumnCases(Schema recaptured) throws SQLException {
+	private void fixColumnCases(Schema recaptured) throws InvalidSchemaError {
```
osheroff, line 679: I'm going to start a review, but this is a big PR and we'll have to go in chunks.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
```diff
@@ -635,27 +634,29 @@ public Position getPosition() {
 	private void fixUnsignedColumns(Schema recaptured) throws SQLException, InvalidSchemaError {
```
osheroff, line 634: I'm thinking we could just get rid of some of this `fixXXX` code.  Any installations that had to have data fixed this way are either fixed already or long dead.

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -142,7 +142,7 @@ private void diffColumnList(List<String> diffs, Table a, Table b, String nameA,
 					EnumeratedColumnDef enumA, enumB;
 					enumA = (EnumeratedColumnDef) column;
 					enumB = (EnumeratedColumnDef) other;
-					if ( !Arrays.deepEquals(enumA.getEnumValues(), enumB.getEnumValues()) ) {
+					if ( !enumA.getEnumValues().equals(enumB.getEnumValues()) ) {
```
osheroff, line 145: I'm not sure I remember or understand why `deepEquals` was used here, or why the change...

src/main/java/com/zendesk/maxwell/schema/Table.java
```diff
@@ -223,8 +223,10 @@ public void diff(List<String> diffs, Table other, String nameA, String nameB) {
 	}
 
 	public void setDefaultColumnCharsets() {
+		String newCharset = this.getCharset();
 		for ( StringColumnDef c : getStringColumns() ) {
-			c.setDefaultCharset(this.getCharset());
+			int index = c.getPos();
+			columns.replace(index, c.withDefaultCharset(newCharset));
```
osheroff, line 229: seems a bit odd to rely on the result of `getPos` here.. for that matter,  I'm not totally sure why `position` would be part of our intern'ed columns...

src/main/java/com/zendesk/maxwell/schema/TableColumnList.java
```diff
@@ -2,36 +2,40 @@
...
 import com.zendesk.maxwell.schema.columndef.ColumnDef;
 
 
 public class TableColumnList implements Iterable<ColumnDef> {
-	private final List<ColumnDef> columns;
-	private Set<String> columnNames;
+	// reduce count of duplicate ArrayLists/Sets for column lists by providing mutability for the class
+	// through references to an internal immutable object that gets interned. This greatly reduces overhead for
+	// table definitions that are duplicated across databases
+	private InternalImmutableTableColumnList internalState;
```
osheroff, line 16: I'd like a better name here than `internalState`.  just `columnList` will be fine.

src/main/java/com/zendesk/maxwell/schema/TableColumnList.java
```diff
@@ -42,43 +46,97 @@ public synchronized int indexOf(String name) {
...
 	public int size() {
-		return columns.size();
+		return internalState.getColumns().size();
 	}
 
-	private void renumberColumns() {
-		short i = 0 ;
-		for ( ColumnDef c : columns ) {
-			c.setPos(i++);
+	private static final class InternalImmutableTableColumnList {
```
osheroff, line 90: This can just be named `ImmutableColumnList`

osheroff (commented):
First part of the review.  Will have to finish later.

src/main/java/com/zendesk/maxwell/schema/Table.java
redcape: `Arrays.deepEquals` is supposed to be used for nested arrays, which this isn't. It probably should've been `Arrays.equals` from the beginning. You can't check `(String[]a).equals((String[]b)` or use `==` so you need one of the Array equals helper methods. With this changed to a (immutable) List to avoid allowing the array to be mutated, we need to use the normal equals method again.

src/main/java/com/zendesk/maxwell/schema/Table.java
redcape: Using the index is a common pattern in Table to retain column order (see addColumn(int, ColumnDef), removeColumn, renameColumn, changeColumn). Though looking deeper, it seems that the code usually uses `findIndex` and only sometimes `getPos`. I think these are identical though, a ColumnDef's `pos` always == TableColumnList index per the constructor of `ImmutableColumnList`.

On interning, the idea is to create less objects overall. Multi-tenant DBs will have the same ordered columns over and over, even small variations on some won't impact the overall savings. However, we need to intern whatever provides equality and restrict mutability to keep the contract for Interner/hashCode of never changing. We can remove `pos` from ColumnDef and always scan the column list with findIndex (to find self-position) or keep it interned.

src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
redcape: Some orgs are way behind and rarely upgrade making long-term upgrade path maintenance a really great feature. I'd rather keep it if that's okay. I've definitely had the misfortune of working places with 8+ year old external dependencies. Continuous upgrade paths make life much easier for the few people willing to improve things.

osheroff:
what does the memory usage look like if we back out the interning of the column lists themselves?  I think ColumnDefs being intern'ed makes a lot a lof of sense, and the code to support it seems a lot easier/cleaner than the code to support all the intern'ing happening at the `ColumnList` level.

src/main/java/com/zendesk/maxwell/schema/columndef/BigIntColumnDef.java
```diff
@@ -3,17 +3,39 @@
...
+import java.util.Objects;
 
 public class BigIntColumnDef extends ColumnDef {
 	static private final BigInteger longlong_max = BigInteger.ONE.shiftLeft(64);
 
-	protected boolean signed;
+	private boolean signed;
 
-	public BigIntColumnDef(String name, String type, short pos, boolean signed) {
+	private BigIntColumnDef(String name, String type, short pos, boolean signed) {
```
osheroff, line 13: hmmm... I think it's kinda-sorta-ok to keep these public rather than force an API change (include someone using maxwell-as-JAR is creating them).  Maybe we could add a deprecation annotation suggesting that people move to the `#create` api instead?

osheroff:
gonna run code through its paces locally.  does look mostly quite good.

redcape:
> what does the memory usage look like if we back out the interning of the column lists themselves? I think ColumnDefs being intern'ed makes a lot a lof of sense, and the code to support it seems a lot easier/cleaner than the code to support all the intern'ing happening at the `ColumnList` level.

We could potentially back out interning ColumnDefs (while keeping the immutability changes). Multi-tenant DB creations create the same tables over and over. Iteratively, each change provides a  significant memory savings.

I looked over the heap dumps I took while developing this and got a more exact size for the live/strongly reachable objects for each change:
``
columndef package retained size:
Original: 138MiB
after ColumnDef interning: 27KiB
after TableColumnList interning: 27KiB

TableColumnList retained size:
Original: 241MiB
After ColumnDef interning: 102MiB
After TableColumnList interning: 14MiB

Schema retained size:
Original: 362MiB
After ColumnDef interning: 223MiB
After TableColumnList interning: 135MiB
```

Interning TableColumnList saves a bunch of duplicate ArrayLists. Removing ColumnDef interning (while keeping immutability changes) would probably keep similar savings since the ColumnDefs would be intern'd by their parent ImmutableTableColumnList objects.

You can see similar effects of duplicate ArrayLists in Table. After all these changes, Table has a retained size of 118MiB, 67MiB of which is ArrayLists for the primary key, 14MiB for TableColumnList. As stated in the description, I'm not trying to fix that in this PR, but that would be the next logical step if we could make sure the API remains clear.

src/main/java/com/zendesk/maxwell/schema/columndef/BigIntColumnDef.java
redcape: I thought it was generally okay to change API per https://maxwells-daemon.io/embedding/. The changes to enforce immutability also removes a bunch of setters, change visibility of public member fields to private, and changes expectations for ColumnDefs to all be implemented as immutable. Why are the constructor changes important to retain?

redcape: I should've asked - what cases would someone be manually creating ColumnDefs instead of using `ColumnDef.build`?

osheroff: yeah I mean who knows what will be in the wild.   To your other point, if we're changing the api dramatically I guess this change is ok too.

osheroff:
ok.  you have anything else to clean up before merge?  I can't say I feel like 100% confident that I'm able to read the full net effect of these changes but I'm willing to go with it.

redcape:
> ok. you have anything else to clean up before merge? I can't say I feel like 100% confident that I'm able to read the full net effect of these changes but I'm willing to go with it.

Thanks, I did one last review pass over it and made some minor changes you can see in the latest commit. It's good to commit from my perspective.

PR #1726, opened 2021-07-21
Change non-test SQL code to try-with-resources

redcape:
This commit is intended to only add try-with-resources to Connection/PreparedStatement/Statement/ResultSets.

Goal when changing this was to keep the same number of resources opened/closed and avoid changing logic/queries  to avoid creating issues.

Reviewer recommendation: review in an IDE that will help you ignore whitespace changes like IntelliJ IDEA


osheroff:
ok i hate the way this shapes the code but I guess it's techincally more correct. 👍 after merge conflict fixes

redcape:
merged and re-ran tests to verify.

PR #1772, opened 2021-11-18
Build non-root image and smaller

milosjanda:
Hi i would like use non-root image, because it's more secure.
I need need it, because our K8s cluster doesn't allow image with root priviledges.

So I prepare build image with non-root priviledges and also with less layer, which is more efficienty.
I used multi-stage build.

Miloš from Livesport

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM maven:3.6-jdk-11
+FROM maven:3.6-jdk-11 as builder
 ENV MAXWELL_VERSION=1.34.1 KAFKA_VERSION=1.0.0
 
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,6 +18,16 @@ RUN cd /workspace \
...
 
+# Build clean image with non-root priveledge
+FROM openjdk:11-jdk-slim
+
+COPY --from=builder /app /app
+
 WORKDIR /app
 
+RUN chown 1000:1000 /app && \
+    KAFKA_VERSION=$KAFKA_VERSION && MAXWELL_VERSION=$MAXWELL_VERSION && echo "$MAXWELL_VERSION" > /REVISION
```
osheroff, line 29: why are we setting the env variables here?

milosjanda: I overlooked that they are no longer used, so i will remove it.

milosjanda: I removed it

Issue #1773, opened 2021-11-20
Maxwell cannot handle dropping table level CHECK Constraints


ndiaz-addepar:
Repro:
```
CREATE TABLE t1 (
col1 int,
col2 int,
CONSTRAINT `ck_for_1` CHECK ( (col1 > 1) AND (col2 >1) )
);

```

When executing `ALTER TABLE t1 DROP CONSTRAINT ck_for_1`

c.z.m.s.d.SchemaChange: Error parsing SQL:  'ALTER TABLE t1 DROP CONSTRAINT ck_for_1'

ndiaz-addepar:
Patch:
[patchmaxwell.txt](https://github.com/zendesk/maxwell/files/7582452/patchmaxwell.txt)

ndiaz-addepar:
PR: https://github.com/zendesk/maxwell/pull/1774

PR #1774, opened 2021-11-22
allow table level check constraints

ndiaz-addepar:
Ignore table level check constraints in the Parser.


ndiaz-addepar:
Related Issue: https://github.com/zendesk/maxwell/issues/1773

osheroff:
thanks!

ndiaz-addepar:
@osheroff Thank you very much for the quick response. Would it be possible to cut a release? It has been a while :D

PR #1775, opened 2021-11-22
feat: add arm64 support for docker images

VictorVasconcelos:


osheroff:
looks good!  let's see if it works...

osheroff:
Hey @VictorVasconcelos looks like I'm running afoul of zendesk's policy of not allowing foreign actions into their repos.  maybe we can just setup `qemu` and `buildx` with shell scripts?

VictorVasconcelos:
Ok @osheroff, I will send another PR with custom actions

VictorVasconcelos:
Sent PR #1777

PR #1779, opened 2021-11-27
datetime columns timezone conversion fix

lpmarques:


osheroff:
kinda scary and deep.  might work, but would need far more tests around various behaviors.

lpmarques:
@osheroff, as mentioned in the issue, I've run some tests and confirmed the fix works for datetimes of any precision, making the bootstrapping process coherent with CDC's behavior of replicating datetime values as stored.

What further tests would you suggest?

PR #1812, opened 2022-01-13
Support DDL Topic Interpolation

davers:
DDL topic does not support interpolation and causes INVALID_TOPIC_EXCEPTION error

👍 - @osheroff


src/main/java/com/zendesk/maxwell/producer/MaxwellKafkaProducer.java
```diff
@@ -269,7 +271,8 @@ void sendAsync(ProducerRecord<String, String> record, Callback callback) {
 		String value = r.toJSON(outputConfig);
 		ProducerRecord<String, String> record;
 		if (r instanceof DDLMap) {
-			record = new ProducerRecord<>(this.ddlTopic, this.ddlPartitioner.kafkaPartition(r, getNumPartitions(this.ddlTopic)), key, value);
+			String topic = this.ddlTopicInterpolator.generateFromRowMap(r);
+			record = new ProducerRecord<>(topic, this.ddlPartitioner.kafkaPartition(r, getNumPartitions(topic)), key, value);
```
osheroff, line 275: the reason I've had my head in the sand and haven't implemented this feature already is that you have to be a bit more careful here: let's say you had a ddlTopic of `ddl_updates_%{table}` -- but then you do "drop database foobar" -- where does that update go?  `ddl_updates_`?

PR #1848, opened 2022-04-04
Make SNS/SQS procuders to respect outputConfig

ciscorn:
I found that output options such as `output_binlog_positions` are not available for SNS/SQS producers.


osheroff:
ok  have you tested this change?  these producers have very little test coverage...

ciscorn:
@osheroff 

Yes, I've tested this change with a real Amazon SNS/SQS queue and was able to enable `output_xoffset`, `output_binlog_position`, etc.

Issue #1863, opened 2022-04-26
Maxwell can crash on restart if binlog disconnection causes connection to different db host


tamimkh:
It seems like when maxwell is [disconnected from a host](https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java#L762) it attempts to [re-connect](https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java#L494) to the [host](https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java#L505-L506) but it does no checks to determine if the db host it just re-connected to is the same db host as before. 

This has the potential to cause issues if maxwell can connect to multiple db hosts (ie: a DNS record re-presents the hosts maxwell can connect to or maxwell connects via a load balancer) and the cluster maxwell is connecting to does not use GTIDs. The most severe of these issues is if this re-connection succeeds without problems (ie: the binlog positions between the initial server (`DB1`) and server it failed over to (`DB2`) are "similar") and maxwell starts writing positions for `DB2` without realizing they are for `DB1`. If the positions drift over time the state of the maxwell positions table can make it impossible for maxwell to re-connect back to the `DB1`.  If maxwell is restarted after this sequence of events it will immediately crash since the positions it will try to use to connect to `DB1` are not valid and point to a random place in a binlog. 

Here is a diagram that explains the sequence of events: 
```mermaid
sequenceDiagram
    Maxwell->>+DB1: Binlog Connected
    Maxwell->>DB1: Binlog File N1, Binlog Pos X1
    Maxwell->>DB1: Network blip, Binlog Server Disconnected
    Maxwell->>+DB2: Reconnect, but DB2 (Doesn't check Server ID)
    Maxwell->>+DB2: Continues Binlog File N1, Binlog Pos X1
    Maxwell->>+DB2: Continues Binlog File N1, Binlog Pos X2
    Maxwell->>+DB2: Continues Binlog File N1, Binlog Pos X3
    Maxwell->>DB2: Network blip, Binlog Server Disconnected
    Maxwell->>+DB1: Binlog re-connected (Doesn't check Server ID)
    Maxwell->>+DB1: Tries Binlog File N1, Binlog Pos X3
    Maxwell->>+DB1: Crash Due to Invalid Binlog Pos
    loop Crash Loop
[...]

PR #1866, opened 2022-05-02
Make BinlogConnectorReplicator throw exception on server ids changing on binlog re-connect

tamimkh:
When maxwell is re-connects to db host it does no checks to see if the id of the server that
it connected to is the same server as it was previously connected to. Normally this results
in a crash since binlog coordinates between servers are generally not compatible. However,
in certain situations it is possible that this re-connection suceeds (ie: the coordinates
between the two servers are similar) and maxwell starts writing positions for the second
server as if they were for the first server. This will eventually make it impossible for
maxwell to re-connect to the first server and cause to crash on startup.

In order to correct this, make maxwell check if the server id has changed on re-connection
and throw an exception. This fixes https://github.com/zendesk/maxwell/issues/1863 which
also describes the problem in more detail

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -491,7 +491,13 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 				// standard binlog positioning is a lot easier; we can really reconnect anywhere
 				// we like, so we don't have to bail out of the middle of an event.
 				LOGGER.warn("replicator stopped at position: {} -- restarting", client.getBinlogFilename() + ":" + client.getBinlogPosition());
+
+				Long oldMasterId = client.getMasterServerId();
 				tryReconnect();
+				if (client.getMasterServerId() != oldMasterId) {
+					throw new Exception("Master id changed from " + oldMasterId + " to " + client.getMasterServerId()
+								+ " while using binlog coordinate positioning. Cannot continue with the info that we have");
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
```diff
@@ -491,7 +491,13 @@ private void ensureReplicatorThread() throws Exception {
 				// standard binlog positioning is a lot easier; we can really reconnect anywhere
 				// we like, so we don't have to bail out of the middle of an event.
 				LOGGER.warn("replicator stopped at position: {} -- restarting", client.getBinlogFilename() + ":" + client.getBinlogPosition());
+
+				Long oldMasterId = client.getMasterServerId();
 				tryReconnect();
+				if (client.getMasterServerId() != oldMasterId) {
+					throw new Exception(
+						"Master id changed from " + oldMasterId + " to " + client.getMasterServerId() + " while using binlog coordinate positioning. Cannot continue with the info that we have");
+				}
```
osheroff, line 500: this should only be an exception in non-gtid mode -- otherwise a logger.warn should be fine

tamimkh: Oh I think is a formatting issue, which I will correct. This is under the `else` block at 490, which if I am not mistaken re-presents non gtid mode

tamimkh: Actually, it seems like github oddity. I have the settings specified for a tab size of 8 and no expand tab. However, for some reason github is showing a tab size of 4 for my changes _only_ on this PR but on the view version of the [file](https://github.com/zendesk/maxwell/blob/733249adc2c0c25ff6aee4545a0455060420ac48/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java#L491). 

Any ideas as to why the PR has odd formatting vs the actual files?

osheroff: i do not know.  it's interesting but i have no idea what would be happening

osheroff: > This is under the else block at 490, which if I am not mistaken represents non gtid mode

ahhh you're right i'm dumb

PR #1868, opened 2022-05-05
fix: localstack sqs

krzysztofzaucha:
An attempt to fix an issue with the localstack sqs connection. Required service endpoint url flag was missing in the cli.

With the update the following should work with localstack:

```
bin/maxwell \
  --host=mysql \
  --port=3306 \
  --user=maxwell \
  --password=password \
  --producer=sqs \
  --sqs_service_endpoint=localstack:4566 \
  --sqs_signing_region=es-west-1 \
  --sqs_queue_uri=http://localstack:4566/000000000000/maxwell
```

I guess handling defaults for backward compatibility would be good to have. 🤔 Unfortunately I am not an expert of Java, a bit of help would be appreciated. ✌️

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -133,7 +133,7 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-core</artifactId>
-      <version>1.12.191</version>
+      <version>1.12.213</version>
     </dependency>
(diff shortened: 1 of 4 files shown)


osheroff:
merged by hand.  if you like, open a second PR to document it.

PR #1895, opened 2022-07-28
feat(replay): Replay binlog files

udyr-fire:
If binlog files expire and are cleared from the mysql service, you can
to use the backup binlog files to replay data


src/main/java/com/zendesk/maxwell/producer/AbstractProducer.java
```diff
@@ -39,6 +39,10 @@ public AbstractProducer(MaxwellContext context) {
 
 	abstract public void push(RowMap r) throws Exception;
 
+	public boolean isDone() {
```
osheroff, line 42: should be named different, like `flushAndClose()`

src/main/java/com/zendesk/maxwell/replay/FilePathParser.java
```diff
@@ -0,0 +1,103 @@
...
+	private List<ReplayFilePattern> parse() {
+		try (InputStreamReader inputStream = new InputStreamReader(new StringInputStream(path))) {
+			this.tokenizer = new StreamTokenizer(inputStream);
+			return doParse(inputStream);
+		} catch (IOException e) {
+			throw new RuntimeException(e.getMessage(), e);
+		}
+	}
+
+	private List<ReplayFilePattern> doParse(InputStreamReader inputStream) throws IOException {
```
osheroff, line 39: uh, this is a weird copy and paste from the filter pattern tokenizer... but I don't think you need to do any of this.

src/main/java/com/zendesk/maxwell/replay/ReplayConfig.java
```diff
@@ -0,0 +1,101 @@
...
+		String replayBinlog = fetchStringOption("replay_binlog", options, properties, null);
+		if (StringUtils.isBlank(replayBinlog)) {
+			usage("Please specify a replay_binlog: ", parser, "replay");
+		}
+		this.binlogFiles.addAll(validateReplayFiles(replayBinlog, parser));
+
+	}
+
+	private List<ReplayFilePattern> validateReplayFiles(String replayBinlog, MaxwellOptionParser parser) {
+		List<ReplayFilePattern> filePatterns = FilePathParser.parse(replayBinlog);
```
osheroff, line 78: yeah, don't do any of this parsing -- just take a list of binlog files as arguments.

osheroff: to be clear, invoking `maxwell-replay` should look like:

```
bin/maxwell-replay --host=XXX --user=YYY /var/log/binlogfile_1 /var/log/binlogfile_2 /tmp/binlogfile_3
```

src/main/java/com/zendesk/maxwell/replay/ReplayBinlogStore.java
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		synchronized (com.zendesk.maxwell.replay.ReplayBinlogStore.class) {
+			if (maxwellOnlySchema != null) {
+				return maxwellOnlySchema;
+			}
+
+			try (Connection conn = schemaConnectionPool.getConnection()) {
+				SchemaCapturer capturer = new SchemaCapturer(conn, caseSensitivity);
+				maxwellOnlySchema = capturer.capture();
```
osheroff, line 50: so this kinda sucks -- basically it means that the binlog files are only replayable as long as the schema hasn't changed.  at the very least we'd have to document this pretty severe limitation.  There is a somewhat better method using binlog heartbeats to try to figure out what schema to start with, but it's a LOT more complicated.

osheroff:
It's a good start but the feature needs some work before merging and there's serious limitations to your approach.

src/main/java/com/zendesk/maxwell/producer/AbstractProducer.java
udyr-fire: FlushAndClose is better, I have updated

src/main/java/com/zendesk/maxwell/replay/FilePathParser.java
udyr-fire: Yes, I referred to the Filter

src/main/java/com/zendesk/maxwell/replay/ReplayConfig.java
udyr-fire: I actually have a lot of binlog files, and I need to select dozens of files from hundreds of files, so I want to solve this problem with a regular expression.

`binlog.000001,binlog.000002,...,binlog.000233,....,binlog.000453`

I mount over 400 files from cloud disk in my directory

Through the binlog **position** I finally confirm that I need to replay  binlog.000233 ~ binlog.000289

osheroff: mmmm no.   Solve this problem outside of maxwell; parse your binlog position (in bash or whatever) and then filter your file list via other methods.  Then pass them all in as command line arguments.  The weird regex stuff is too particular to your specific use case.

udyr-fire: this is fine, I need a script to generate a long file configuration, I will provide a helper script

src/main/java/com/zendesk/maxwell/replay/ReplayBinlogStore.java
udyr-fire: Keeping the schema consistent with Binlog is difficult unless the mysql service is initialized using Maxwell without interruption. In addition, I used max_schemas=1000, which will cause the schema to be cleared

udyr-fire: I use mysqldump to back up the schema every day. When I need to replay binlog, I will restore the backup schema to a temporary mysql service and provide it to Maxwell to read the correct schema

udyr-fire: I used max_schemas =1000, which will cause the schema to be cleared. So I can't use heartbeats.

Using the correct schema and position, chase binlog from this positionto the desired output position.

This schema needs to be provided by the user. If you use Maxwell's schema, it may be missing.

The user didn't keep Maxwell running healthily all the time, and it may have been restarted. Maybe Maxwell has restored the latest schema

The **replay** does not need whether Maxwell has run before, but only needs a correct schema and all binlogs. Does this make the **replay** more simple and reliable?

src/main/java/com/zendesk/maxwell/replay/MaxwellReplayFile.java
```diff
@@ -0,0 +1,378 @@
...
+				row = getRow(reader, binlogFile.getName());
+			}
+		} catch (Exception e) {
+			throw new RuntimeException("Replay failed, Check from: " + position + ", error: " + e.getMessage(), e);
+		} finally {
+			LOGGER.info("End replay binlog file: {}", binlogFile.getAbsoluteFile());
+		}
+	}
+
+	private RowMap getRow(BinaryLogFileReader reader, String binlogName) throws Exception {
```
osheroff, line 146: maybe we could find a way to refactor the original class where this came from to re-use it?  Like maybe turn it into a static method that takes `rowBuffer` as an argument or something?

src/main/java/com/zendesk/maxwell/replay/ReplayBinlogStore.java
osheroff: > The replay does not need whether Maxwell has run before, but only needs a correct schema and all binlogs. Does this make the replay more simple and reliable?

it makes it simpler from the perspective of this code, but operationally it's terrible: you'd need to somehow have a copy of the schema from the point in time that you're playing the binlogs from -- if the schema has changed at all, you'd need a database backup or something.

src/main/java/com/zendesk/maxwell/replay/MaxwellReplayFile.java
udyr-fire: I added a BinlogConnectorEventProcessor class to handle the event - related data, removed the BinlogConnectorReplicator section method

Issue #1927, opened 2022-10-25
polardb-x(aliyun) failed to execute the create index statement


wanghangyu817:
**Execute SQL ： CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)**

**Exception:**

```
31077 [DEBUG] C3P0PooledConnectionPool: Testing PooledConnection [com.mchange.v2.c3p0.impl.NewPooledConnection@38d72334] on CHECKOUT.
31078 [DEBUG] C3P0PooledConnectionPool: Test of PooledConnection [com.mchange.v2.c3p0.impl.NewPooledConnection@38d72334] on CHECKOUT has SUCCEEDED.
31085 [DEBUG] SchemaChange: SQL_PARSE <- "# POLARX_ORIGIN_SQL=CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)
# POLARX_TSO=xxxxx
CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)"
31156 [ERROR] MysqlParserListener: (statement CREATE INDEX device_id_idx ON event_tracking_info_extra ( event , create_time ))
31156 [ERROR] SchemaChange: Error parsing SQL: '# POLARX_ORIGIN_SQL=CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)
# POLARX_TSO=xxxxx
CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)'
31156 [ERROR] AbstractSchemaStore: Error on bin log position Position[BinlogPosition[binlog.000001:329876], lastHeartbeat=1666677625205]
com.zendesk.maxwell.schema.ddl.MaxwellSQLSyntaxError: CREATE
        at com.zendesk.maxwell.schema.ddl.MysqlParserListener.visitErrorNode(MysqlParserListener.java:93)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:17)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:28)
        at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:28)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parseSQL(SchemaChange.java:94)
        at com.zendesk.maxwell.schema.ddl.SchemaChange.parse(SchemaChange.java:109)
[...]

PR #1928, opened 2022-10-25
Fix DDL multi-line comments

wanghangyu817:
As shown in the #1927 issue, the binlog comment in polardb-x（aliyun） was in line with the mysql annotation specification, when maxwell threw exceptions when executing specific sql, and tests found that the comment could be multiple lines in a piece of sql.
eg:
```
String sql = "# POLARX_ORIGIN_SQL=CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)\n" +
                    "# POLARX_TSO=699055265586505324815216178714011320320000000000000000\n" +
                    "# POLARX_TSO=69905526558650532481521617871401x320320000000000000000\n" +
                    "CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)";
```
execute : 
``` 
String s = sql.replaceAll("^\\s*#.*", "");
System.out.println("s = " + s); 
```
results : 
``` 
s = 
# POLARX_TSO=699055265586505324815216178714011320320000000000000000
# POLARX_TSO=69905526558650532481521617871401x320320000000000000000
CREATE INDEX device_id_idx ON event_tracking_info_extra (event, create_time)
```
eventually an exception occurs #1927 

the solution is to replace the comment section with a multi-line match, resulting in the actual executed sql

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
+++ b/src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
@@ -58,7 +58,7 @@ public abstract class SchemaChange {
 		// now strip out comments
 		sql = CSTYLE_COMMENTS.matcher(sql).replaceAll("");
 		sql = sql.replaceAll("\\-\\-.*", "");
-		sql = sql.replaceAll("^\\s*#.*", "");
+		sql = Pattern.compile("^\\s*#.*", Pattern.MULTILINE).matcher(sql).replaceAll("");
 
 		for (Pattern p : SQL_BLACKLIST) {
 			if (p.matcher(sql).find()) {
--- a/src/test/java/com/zendesk/maxwell/schema/ddl/DDLParserTest.java
+++ b/src/test/java/com/zendesk/maxwell/schema/ddl/DDLParserTest.java
@@ -602,4 +602,16 @@ public class DDLParserTest {
 		}
 		System.out.println(nFixed + " fixed, " + nErr + " remain.");
 	}
+
+	@Test
+	public void testPolardbXCreateIndexSQL(){
+
+		List<SchemaChange> changes = parse(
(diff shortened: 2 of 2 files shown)


osheroff:
Hi, can you add a test for your code please?

wanghangyu817:
> 嗨，你可以为你的代码添加一个测试吗？

Add to DDLParserTest? Or something else

osheroff:
Yeah thats good

src/main/java/com/zendesk/maxwell/schema/ddl/SchemaChange.java
```diff
@@ -58,7 +58,7 @@ private static boolean matchesBlacklist(String sql) {
 		// now strip out comments
 		sql = CSTYLE_COMMENTS.matcher(sql).replaceAll("");
 		sql = sql.replaceAll("\\-\\-.*", "");
-		sql = sql.replaceAll("^\\s*#.*", "");
+		sql = Pattern.compile("^\\s*#.*", Pattern.MULTILINE).matcher(sql).replaceAll("");
```
osheroff, line 61: uh this doesn't seem right to me now that i look closer at it -- this regex will strip out all the sql if anything starts with `#`.

wanghangyu817: sql to this point should be ddl, sql starting with #, is there any other use? The processing method before my test really cannot handle multiple lines of sql with "newline character" in one line of sql and starting with #. We should handle this problem. I hope to get your advice

osheroff: i take it back, your PR is fine.  it looks dangerous but it's not.

wanghangyu817: I hope I can get some help to solve this problem

PR #1932, opened 2022-11-01
match mariadb comment syntax of /*M! ... */ in blacklist

grooverdan:
Reference syntax - https://mariadb.com/kb/en/comment-syntax/


osheroff:
thanks maria.  can you add a parser test?

Issue #1931, opened 2022-10-30
Maxwell crashing/going down


29swastik:
We have maxwell running on docker in our production and it is configured to listen to changes on MariadDb tables and push them to a kafka topic. Our MariadDb has some 10-12 tables and we have configured maxwell to listen to only 4 tables and all these 4 tables are heavily used (number of writes is too high). 

Recently we have seen that our Maxwell is going down, following are the few error logs we found when maxwell is going down

```
Oct 28, 2022 @ 06:10:40.037	06:10:40,037 ERROR MysqlParserListener - (parse SET STATEMENT max_statement_time = 60 FOR flush table)

Oct 28, 2022 @ 06:10:40.040	06:10:40,040 ERROR SchemaChange - Error parsing SQL: 'SET STATEMENT max_statement_time=60 FOR flush table'

Oct 28, 2022 @ 06:10:40.047	06:10:40,045 ERROR AbstractSchemaStore - Error on bin log position Position[BinlogPosition[mysql-bin-changelog.049610:3209], lastHeartbeat=1666898518069]

```

As the last error log says **Error on bin log position Position[BinlogPosition[mysql-bin-changelog.049610:3209]**, so to verify whether bin log file with name `mysql-bin-changelog.049610:3209` exists or not we ran `SHOW binary logs;` and we found that file in list outputted by above command 

Please help me in understanding what is going wrong here and help in solving the issue

**Environment Details**

1. OS: Centos
2. Platform - AWS instance(T4g Medium)
3. CPU - 2 CPU
4. RAM - 4GB

**Maxwell Version** - `v1.37.7`

**Note**: We are not handling DDL & our log retention is 3 days

Detailed logs that was logged when Maxwell was going down, sorry for posting huge logs :)

> ```
> Oct 28, 2022 @ 06:10:39.084	06:10:39,082 INFO  ProducerConfig - ProducerConfig values: 
> Oct 28, 2022 @ 06:10:39.146	06:10:39,146 INFO  AppInfoParser - Kafka commitId : aaa7af6d4a11b29d
> 
> Oct 28, 2022 @ 06:10:39.146	06:10:39,146 INFO  AppInfoParser - Kafka version : 1.0.0
> 
[...]

PR #1934, opened 2022-11-02
DDL blacklist could be prefixed with SET STATEMENT .. FOR

grooverdan:
This results in blacklisted DDL statements still passing through.

Closes: #1931

I've got the basic case working. My regex around quotes values still isn't right so assistance there would be appreciated. Its been a while since I've done professional regexes.


grooverdan:
0bb64f7612523b3908cb1ac472a6b2e85a1fd634 might be sufficient. `set statement` can be applied to any SQL however and I'm concerned this might over-filter.

osheroff:
I see.  yeah this approach is better, in theory someone can do `set statement for alter table`...

osheroff:
trying to do quote matching and such in regexs is pretty nasty usually.  I might try something like:

```
"SET\\s+STATEMENT\\s+(\\w+\\s*=('.*'|".*"|[\\d\\.]+,?\\s+)+\\s*FOR"
```

grooverdan:
Done. Non-greedy helped. as does the duplicate prevention of multiple items. Thanks.

29swastik:
hi @osheroff when is this fix expected to available?

osheroff:
thanks!

PR #1943, opened 2022-11-09
WIP implementation of Vitess VStream replicator support

### LINKED ISSUE #1757: Connect to Vitess
Is it possible to read CDC from [Vitess](https://vitess.io/)? I am not sure how Maxwell will read the MySQL binlogs behind it and store their files and GTID's in the meta db.

kovyrin:
Attempts to implement/fix #1757.

Things that work:
* The code successfully connects to `vtgate` via the VStream API
* Maxwell follows a given keyspace in a Vitess cluster (a specific shard or all shards) and receives all relevant events
* Maxwell uses FIELD events from VStream to maintain an in-memory representation of the database schema (meaning we don't need to load it from MySQL).
* Using the in-memory schema, Maxwell converts VStream ROW events into `RowMap` objects and pushes those to a given producer.

Still need to implement/fix/figure out:
* Position store is stubbed out for now and position recovery is not implemented, so it always starts at the current stream position.
* I'm not sure I'm handling GRPC stream shutdown properly.
* Had to disable OpenCensus dependency for now because they depend on an ancient grpc version that conflicts with the grpc dependency used by vitess.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,12 @@
       <version>15.3</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.vitess</groupId>
+      <artifactId>vitess-grpc-client</artifactId>
+      <version>14.0.1</version>
+    </dependency>
+
     <!-- utils and support libs -->
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
@@ -261,8 +267,9 @@
       <artifactId>simpleclient_servlet</artifactId>
       <version>0.9.0</version>
     </dependency>
+
     <!-- OpenCensus used for pushing metrics to stackdriver -->
-    <dependency>
+    <!-- <dependency>
         <groupId>io.opencensus</groupId>
         <artifactId>opencensus-contrib-dropwizard</artifactId>
         <version>${opencensus.version}</version>
@@ -282,7 +289,7 @@
         <groupId>io.opencensus</groupId>
         <artifactId>opencensus-exporter-stats-stackdriver</artifactId>
         <version>${opencensus.version}</version>
-    </dependency>
+    </dependency> -->
 
     <!-- producer libs -->
     <dependency>
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -9,6 +9,7 @@ import com.zendesk.maxwell.recovery.RecoveryInfo;
 import com.zendesk.maxwell.replication.BinlogConnectorReplicator;
 import com.zendesk.maxwell.replication.Position;
 import com.zendesk.maxwell.replication.Replicator;
+import com.zendesk.maxwell.replication.VStreamReplicator;
 import com.zendesk.maxwell.row.HeartbeatRowMap;
 import com.zendesk.maxwell.schema.*;
 import com.zendesk.maxwell.schema.columndef.ColumnDefCastException;
@@ -239,58 +240,72 @@ public class Maxwell implements Runnable {
 	}
 
 	private void startInner() throws Exception {
-		try ( Connection connection = this.context.getReplicationConnection();
-		      Connection rawConnection = this.context.getRawMaxwellConnection() ) {
-			MaxwellMysqlStatus.ensureReplicationMysqlState(connection);
+		try (Connection rawConnection = this.context.getRawMaxwellConnection()) {
 			MaxwellMysqlStatus.ensureMaxwellMysqlState(rawConnection);
-			if (config.gtidMode) {
-				MaxwellMysqlStatus.ensureGtidMysqlState(connection);
-			}
 
 			SchemaStoreSchema.ensureMaxwellSchema(rawConnection, this.config.databaseName);
 
-			try ( Connection schemaConnection = this.context.getMaxwellConnection() ) {
+			try (Connection schemaConnection = this.context.getMaxwellConnection()) {
 				SchemaStoreSchema.upgradeSchemaStoreSchema(schemaConnection);
 			}
 		}
 
 		AbstractProducer producer = this.context.getProducer();
 
-		Position initPosition = getInitialPosition();
-		logBanner(producer, initPosition);
-		this.context.setPosition(initPosition);
+    if (config.vitessEnabled) {
+      this.replicator = new VStreamReplicator(
+        config.vitessConfig,
+        producer,
+        context.getMetrics(),
+        context.getFilter(),
+        config.bufferMemoryUsage,
+        config.binlogEventQueueSize
+      );
+    } else {
+      try(Connection connection = context.getReplicationConnection()) {
+        MaxwellMysqlStatus.ensureReplicationMysqlState(connection);
 
-		MysqlSchemaStore mysqlSchemaStore = new MysqlSchemaStore(this.context, initPosition);
-		BootstrapController bootstrapController = this.context.getBootstrapController(mysqlSchemaStore.getSchemaID());
+        if (config.gtidMode) {
+          MaxwellMysqlStatus.ensureGtidMysqlState(connection);
+        }
+      }
 
-		this.context.startSchemaCompactor();
+      Position initPosition = getInitialPosition();
+      logBanner(producer, initPosition);
+      this.context.setPosition(initPosition);
 
-		if (config.recaptureSchema) {
-			mysqlSchemaStore.captureAndSaveSchema();
-		}
+      MysqlSchemaStore mysqlSchemaStore = new MysqlSchemaStore(this.context, initPosition);
+      BootstrapController bootstrapController = this.context.getBootstrapController(mysqlSchemaStore.getSchemaID());
 
-		mysqlSchemaStore.getSchema(); // trigger schema to load / capture before we start the replicator.
+      this.context.startSchemaCompactor();
 
-		this.replicator = new BinlogConnectorReplicator(
-			mysqlSchemaStore,
-			producer,
-			bootstrapController,
-			config.replicationMysql,
-			config.replicaServerID,
-			config.databaseName,
-			context.getMetrics(),
-			initPosition,
-			false,
-			config.clientID,
-			context.getHeartbeatNotifier(),
-			config.scripting,
-			context.getFilter(),
-			context.getConfig().getIgnoreMissingSchema(),
-			config.outputConfig,
-			config.bufferMemoryUsage,
-			config.replicationReconnectionRetries,
-			config.binlogEventQueueSize
-		);
+      if (config.recaptureSchema) {
+        mysqlSchemaStore.captureAndSaveSchema();
+      }
+
+      mysqlSchemaStore.getSchema(); // trigger schema to load / capture before we start the replicator.
+
+      this.replicator = new BinlogConnectorReplicator(
+        mysqlSchemaStore,
+        producer,
+        bootstrapController,
+        config.replicationMysql,
+        config.replicaServerID,
+        config.databaseName,
+        context.getMetrics(),
+        initPosition,
+        false,
+        config.clientID,
+        context.getHeartbeatNotifier(),
+        config.scripting,
+        context.getFilter(),
+        context.getConfig().getIgnoreMissingSchema(),
+        config.outputConfig,
+        config.bufferMemoryUsage,
+        config.replicationReconnectionRetries,
+        config.binlogEventQueueSize
+      );
+    }
 
 		context.setReplicator(replicator);
 		this.context.start();
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -82,6 +82,16 @@ public class MaxwellConfig extends AbstractConfig {
 	public Boolean gtidMode;
 
 	/**
+	 * If Maxwell is running against a vitess cluster.
+	 */
+	public Boolean vitessEnabled;
+
+	/**
+	 * Vitess (vtgate) connection config
+	 */
+  public MaxwellVitessConfig vitessConfig;
+
+	/**
 	 * Name of database in which to store maxwell data (default `maxwell`)
 	 */
 	public String databaseName;
@@ -641,6 +651,10 @@ public class MaxwellConfig extends AbstractConfig {
 		this.schemaMysql = new MaxwellMysqlConfig();
 		this.masterRecovery = false;
 		this.gtidMode = false;
+
+    this.vitessEnabled = false;
+    this.vitessConfig = new MaxwellVitessConfig();
+
 		this.bufferedProducerSize = 200;
 		this.outputConfig = new MaxwellOutputConfig();
 		setup(null, null); // setup defaults
@@ -1043,9 +1057,11 @@ public class MaxwellConfig extends AbstractConfig {
 		this.replicationMysql   = parseMysqlConfig("replication_", options, properties);
 		this.schemaMysql        = parseMysqlConfig("schema_", options, properties);
 		this.gtidMode           = fetchBooleanOption("gtid_mode", options, properties, System.getenv(GTID_MODE_ENV) != null);
-
 		this.databaseName       = fetchStringOption("schema_database", options, properties, "maxwell");
 		this.maxwellMysql.database = this.databaseName;
+
+    this.vitessEnabled      = fetchBooleanOption("vitess", options, properties, false);
+    this.vitessConfig       = parseVitessConfig(options, properties);
 
 		this.producerFactory    = fetchProducerFactory(options, properties);
 		this.producerType       = fetchStringOption("producer", options, properties, "stdout");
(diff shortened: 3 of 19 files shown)


kovyrin:
The `config.properties` I'm using for my development/testing:

```
# tl;dr config
log_level=debug

producer=stdout
# producer=kafka
# kafka.bootstrap.servers=localhost:9092

# Enable vitess support
vitess=true
vitess_keyspace=commerce
vitess_shard=
vtgate_host=localhost
vtgate_port=15991

# MySQL login info for storing maxwell metadata
host=storefront-renderer.railgun
user=root
password=
```

osheroff:
> You will notice, that there is a set of files within the vitess directory that have Apache 2.0 headers. Those are derivatives of the code used by the [Vitess connector in Debezium (https://debezium.io/documentation/reference/stable/connectors/vitess.html). Will need guidance on how to properly integrate that code into the codebase (if there is a NOTICE file where it needs mentioning or something like that) or if it needs to be re-implemented (it mostly revolves around schema caching and VStream field value parsing/conversion).

Definitely don't re-implement.  Maxwell is apache2 as well, copying it on in is fine.  The debezium guys are friends, of sorts, and a NOTICE file will be fine.

osheroff:
Nice start!  Let me know if there's anything you're unclear on; I presume that `maxwell`.`positions` will have enough fields to store your position.

I won't get deep into source-level comments just now, on a very surface level you should know that maxwell uses tabs-not-spaces :)

kovyrin:
Added support for Vitess position tracking and recovery.

Few notes on the implementation:

* I've added a new `TEXT` field to the `positions` table for storing vgtid values since they may get pretty large for large Vitess clusters and I did not want to risk storing it in the `gtid_set` column (+ the format is different, so I was worried it may cause parsing problems down the road).

* When initializing a new deployment of Maxwell, instead of asking for a current position and persisting it right away (like we do for MySQL), I have opted for waiting for the initial `VGTID` event that is sent to the client immediately upon connection and then using that value as the initial vgtid (so the semantics of initial positioning have changed a bit).

* The `Position` class looks gross and polluted with the VGTID stuff now. I will refactor it as the code stabilizes.

* Finally, I have not implemented heartbeat support for Vitess since at the moment I'm targeting a deployment model where Maxwell has a separate MySQL cluster for storing its management schema (since it does not fit into the sharded model of Vitess directly as all the tables miss a sharding key) and those heartbeats from the Maxwell database will never be visible within the Vitess change stream.

kovyrin:
OK, polished it all a bit, cleaned up some changes we didn't need and tomorrow will try to deploy in staging within our infrastructure and run on a relatively large Vitess cluster to see how it goes. 🤞🏻

osheroff:
> I've added a new TEXT field to the positions table for storing vgtid values since they may get pretty large for large Vitess clusters and I did not want to risk storing it in the gtid_set column 

That's fine, two things:

1.  need some code in upgradeSchemaStoreSchema to add the column to existing installs
2. probably name it `vitess_gtid` for clarity

osheroff:
> I have not implemented heartbeat support for Vitess since at the moment I'm targeting a deployment model where Maxwell has a separate MySQL cluster for storing its management schema (since it does not fit into the sharded model of Vitess directly as all the tables miss a sharding key) and those heartbeats from the Maxwell database will never be visible within the Vitess change stream.

That's fine, we'll need some code strewn around the codebase to turn off the expectation of heartbeats.

osheroff:
do you have a docker-compose type setup I could use to step through the code?

osheroff:
can I get a 10k foot view of how the in-memory schema works?  Does vitess keep track of what tables a stream listeners has seen already and then send any unseen tables + changes?

osheroff:
> I have not implemented heartbeat support for Vitess

Honestly the whole heartbeating thing even for mysql is of dubious value; it was written in a world where GTID wasn't hugely common to support a wild and overly engineered desire to not lose data during primary server flips.  I may start trudging down the ugly and unpleasant path of deprecation at some point.

kovyrin:
> can I get a 10k foot view of how the in-memory schema works? Does vitess keep track of what tables a stream listeners has seen already and then send any unseen tables + changes?

Exactly! For each VStream connection, they keep track of what tables and what versions of those tables a client has seen and before sending any data for a given table, they make sure the client first gets a `FIELD` event explaining that the table schema should look like.

Here is a demo of how it works in reality:

1. Let's connect to VStream and listen for updates:

```
❯ grpcurl -plaintext -d '{"vgtid":{"shard_gtids":[{"keyspace":"commerce", "gtid":"current"}]}}' localhost:15991 vtgateservice.Vitess.VStream
```

It will immediately send the `VGTID` event with the current position (which we use to set record the initial position in Maxwell).

```json
{
  "events": [
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "commerce",
            "shard": "0",
            "gtid": "MySQL56/ff450168-5ed6-11ed-9415-954189eafa28:1-2060"
          }
        ]
      },
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "OTHER",
      "keyspace": "commerce",
      "shard": "0"
    }
  ]
}
```

2. Now, let's create a simple table with `create table bananas(name varchar(100));`. The stream will receive a `DDL` event with the DDL statement:

```json
{
  "events": [
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "commerce",
            "shard": "0",
            "gtid": "MySQL56/ff450168-5ed6-11ed-9415-954189eafa28:1-2061"
          }
        ]
      },
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "DDL",
      "timestamp": "1669215464",
      "statement": "create table bananas (\n\t`name` varchar(100)\n)",
      "currentTime": "1669215464575035000",
      "keyspace": "commerce",
      "shard": "0"
    }
  ]
}
```

3. Now, we insert some data `insert into bananas set name="test1";`. The stream receives a `FIELD` event with the schema for the new table, followed by a `ROW` event with the actual data:

```json
{
  "events": [
    {
      "type": "BEGIN",
      "timestamp": "1669215490",
      "currentTime": "1669215490981788000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "FIELD",
      "timestamp": "1669215490",
      "fieldEvent": {
        "tableName": "commerce.bananas",
        "fields": [
          {
            "name": "name",
            "type": "VARCHAR",
            "table": "bananas",
            "orgTable": "bananas",
            "database": "vt_commerce",
            "orgName": "name",
            "columnLength": 400,
            "charset": 45,
            "columnType": "varchar(100)"
          }
        ],
        "keyspace": "commerce",
        "shard": "0"
      },
      "currentTime": "1669215490990077000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "ROW",
      "timestamp": "1669215490",
      "rowEvent": {
        "tableName": "commerce.bananas",
        "rowChanges": [
          {
            "after": {
              "lengths": [
                "5"
              ],
              "values": "dGVzdDE="
            }
          }
        ],
        "keyspace": "commerce",
        "shard": "0"
      },
      "currentTime": "1669215490990132000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "commerce",
            "shard": "0",
            "gtid": "MySQL56/ff450168-5ed6-11ed-9415-954189eafa28:1-2063"
          }
        ]
      },
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "COMMIT",
      "timestamp": "1669215490",
      "currentTime": "1669215490990165000",
      "keyspace": "commerce",
      "shard": "0"
    }
  ]
}
```

4. Let's insert another row `insert into bananas set name="test2";`. Now the stream only gets the transaction, but no schema updates:

```json
{
  "events": [
    {
      "type": "BEGIN",
      "timestamp": "1669215502",
      "currentTime": "1669215502107840000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "ROW",
      "timestamp": "1669215502",
      "rowEvent": {
        "tableName": "commerce.bananas",
        "rowChanges": [
          {
            "after": {
              "lengths": [
                "5"
              ],
              "values": "dGVzdDI="
            }
          }
        ],
        "keyspace": "commerce",
        "shard": "0"
      },
      "currentTime": "1669215502107932000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "commerce",
            "shard": "0",
            "gtid": "MySQL56/ff450168-5ed6-11ed-9415-954189eafa28:1-2064"
          }
        ]
      },
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "COMMIT",
      "timestamp": "1669215502",
      "currentTime": "1669215502107961000",
      "keyspace": "commerce",
      "shard": "0"
    }
  ]
}
```

5. Let's change the schema: `alter table bananas change name name varchar(50);`. The stream gets a `DDL` event:

```json
{
  "events": [
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "commerce",
            "shard": "0",
            "gtid": "MySQL56/ff450168-5ed6-11ed-9415-954189eafa28:1-2065"
          }
        ]
      },
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "DDL",
      "timestamp": "1669215523",
      "statement": "alter table bananas change column `name` `name` varchar(50)",
      "currentTime": "1669215523734671000",
      "keyspace": "commerce",
      "shard": "0"
    }
  ]
}
```

6. Let's insert more data: `insert into bananas set name="test3";`. Now VTgate detects that we have not seen the new schema yet, so it sends us a transaction with a `FIELD` event inside before sending us the data:

```json
{
  "events": [
    {
      "type": "BEGIN",
      "timestamp": "1669215535",
      "currentTime": "1669215535606806000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "FIELD",
      "timestamp": "1669215535",
      "fieldEvent": {
        "tableName": "commerce.bananas",
        "fields": [
          {
            "name": "name",
            "type": "VARCHAR",
            "table": "bananas",
            "orgTable": "bananas",
            "database": "vt_commerce",
            "orgName": "name",
            "columnLength": 200,
            "charset": 45,
            "columnType": "varchar(50)"
          }
        ],
        "keyspace": "commerce",
        "shard": "0"
      },
      "currentTime": "1669215535619492000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "ROW",
      "timestamp": "1669215535",
      "rowEvent": {
        "tableName": "commerce.bananas",
        "rowChanges": [
          {
            "after": {
              "lengths": [
                "5"
              ],
              "values": "dGVzdDM="
            }
          }
        ],
        "keyspace": "commerce",
        "shard": "0"
      },
      "currentTime": "1669215535619512000",
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "commerce",
            "shard": "0",
            "gtid": "MySQL56/ff450168-5ed6-11ed-9415-954189eafa28:1-2066"
          }
        ]
      },
      "keyspace": "commerce",
      "shard": "0"
    },
    {
      "type": "COMMIT",
      "timestamp": "1669215535",
      "currentTime": "1669215535619533000",
      "keyspace": "commerce",
      "shard": "0"
    }
  ]
}
```

osheroff:
nice, very ergonomic.  

Just something to keep in the back of your head:

Maxwell supports configurations with near-psychotic sized schemas, hundreds of thousands of tables.  To this end, it's important to keep in-memory schema representations as compact as possible.  The default in-memory representation jumps through crazy `intern` hoops to do so.  Not saying you should do this in v1, but if you can design your data to be somewhat compact it'll help us down the line.

kovyrin:
> do you have a docker-compose type setup I could use to step through the code?

Working today on setting that up (so far it has been based on a bunch of internal tooling we use at Shopify + a bunch of duct tape around official scripts shopped with Vitess for local dev environment setup).

kovyrin:
> do you have a docker-compose type setup I could use to step through the code?

Added a `vitess` directory to the project, including a set of scripts and a README file explaining how to run it in development/test mode. That should allow you to run the whole setup with just two commands.

kovyrin:
> The default in-memory representation jumps through crazy intern hoops to do so. Not saying you should do this in v1, but if you can design your data to be somewhat compact it'll help us down the line.

I've added intern calls to all keyspace/table/field/type names and interned enum values. Didn't go as far as your approach with interning whole column definitions, etc, but I think the object model used for Vitess schemas in memory is compatible with that approach as well, so could later be improved (or we could potentially even reuse your existing schema def stuff for Vitess since they are very close in what they do).

kovyrin:
Added TLS support to TVgate connections.

src/main/java/com/zendesk/maxwell/replication/Position.java
```diff
@@ -69,8 +67,10 @@ public int hashCode() {
...
 
 	public boolean newerThan(Position other) {
-		if ( other == null )
-			return true;
-		return this.getBinlogPosition().newerThan(other.getBinlogPosition());
+		if (other instanceof Position) {
+			Position mOther = (Position) other;
+			return binlogPosition.newerThan(mOther.binlogPosition);
+		}
+		return true;
```
osheroff, line 74: I think this function read better before...

kovyrin: Sorry, those are artifacts from my attempt to refactor position classes in a slightly different manner, they don't make much sense anymore.  Thank you for spotting those, I've reverted the changes now.

kovyrin:
Ok, we have our first instance of Maxwell running on Vitess deployed in Staging on top of a real cluster and it seems to work as expected. More testing, benchmarks, etc over the next couple of weeks.

kovyrin:
Added and tested both TLS certificate and user/password auth on our staging instances - both work as expected.

osheroff:
Very nice.  I’m pretty much good to merge and release whenever you feel confident On Dec 7, 2022, at 12:17, Oleksiy Kovyrin ***@***.***> wrote:﻿
Added and tested both TLS certificate and user/password auth on our staging instances - both work as expected.

kovyrin:
I'd like it to bake for a bit in our staging environments (until January?) and then we can be more confident about it. But thank you for your vote of confidence! 😊

osheroff:
SGTMOn Dec 7, 2022, at 12:53, Oleksiy Kovyrin ***@***.***> wrote:﻿
I'd like it to bake for a bit in our staging environments (until January?) and then we can be more confident about it. But thank you for your vote of confidence! 😊

kovyrin:
@osheroff Noticed an issue today:

Due to the way VStream ROW events come in before the VGTID event for a given transaction and they do not have their own position values, I end up simply setting the position on the last ("commit") RowMap event manually at the very end. \

I've looked at your existing producers and they seem to use the `nextPosition` value from RowMap to commit the latest position, so I've assumed the nextPosition is what I need to set. It works in all my testing with built-in producers.

Interestingly though, your example CustomProducer (and, as a result, our a producer we use for our project here at Shopify) has `context.setPosition(r.getPosition());`. Since I do not set those values on VStream-originating RowMap events (the value is a static field and I wasn't sure I should touch it), the whole position tracking thing ends up not working, leading to all kinds of fun problems.

So, the question is: which is the right way to update progress from a producer? And, if both the position and the last position field could potentially be used, should I just set them to the same value to make the system work for people who may have followed the example producer code and have the same logic in place?

Thank you for any insight you can provide into this.

osheroff:
ugh, I'm sure that's just a dumb bug with the example `CustomerProducer`, or maybe `RowMap#getPosition` _used_ to actually return the next position, or something.

The proper behavior is most assuredly to do `context.setPosition(r.getNextPosition())`.  Does that work with vitess' view of how positions are tracked?

kovyrin:
> The proper behavior is most assuredly to do `context.setPosition(r.getNextPosition())`.  

Great, I'll just use that in our custom producer.

> Does that work with vitess' view of how positions are tracked?  

Yes, absolutely, that works!

osheroff:
how'd it all go @kovyrin?  stall a bit or you in prod?

kovyrin:
> all a bit or you in prod?

Hey, Ben! We're slowly rolling it out in some of our environments and ironing out rough edges, etc. Currently exploring different options for coordination of multiple instances of Maxwell running with the same config (sometimes happens during rolling upgrades, k8s cluster failovers, etc). I'm considering adding Zookeeper support for leader election or something like that (will propose it in a separate PR if we decide to go that route).

I'll try to polish the Vitess PR enough over the next couple of weeks and then request a final review/merge if that's OK with you. Thank you for your support!

kovyrin:
@dbmurphy Unfortunately, due to [some changes](https://globalnews.ca/news/9673162/shopify-layoffs-may-2023/) at Shopify, we had to pivot and change our approach for the project, making Maxwell redundant in our infrastructure.  We have run this code in production for about a couple of months before shutting down, so I am fairly certain in it as an experimental feature, but may not be able to push it over the finish line at this point. :(

Issue #1924, opened 2022-10-17
Dependency fixes


mayamessinger:
Hello from GitHub engineering! We're big fans of Maxwell and really appreciate y'all making this library open source. Our security team brought some vulnerabilities to our attention in a project we have that uses Maxwell, that we'd like to contribute back (and then use the newly-secured release version ourselves)

The vulnerabilities are:
* libtirpc-common CVE-2021-46828
	* 1.3.1-1 -> needs 1.3.1-1+deb11u1
* libtirpc3 CVE-2021-46828
	* 1.3.1-1 -> needs 1.3.1-1+deb11u1
* libpcre2-8-0 CVE-2022-1586
	* 10.36-2 -> needs 10.36-2+deb11u1
* libgnutls30  CVE-2022-2509
	* 3.7.1-5+deb11u1 -> needs 3.7.1-5+deb11u2
* zlib1g CVE-2022-37434
	* 1:1.2.11.dfsg-2+deb11u1 -> needs 1:1.2.11.dfsg-2+deb11u2
* libc-bin CVE-2021-3999
	* 2.31-13+deb11u3 -> needs 2.31-13+deb11u4
* libc6 CVE-2021-3999
	* 2.31-13+deb11u3 -> needs 2.31-13+deb11u4
* jackson-databind GHSA-57j2-w4cx-62h2
	* 2.11.2 -> needs 2.14.0-rc1
* protobuf-java GHSA-h4h5-3hr4-j3g2
	* 3.21.4 -> needs 3.21.7
* jetty-http GHSA-cj7v-27pg-wf7q
	* 9.4.41.v20210516 -> needs 9.4.47

The lib* dependencies are likely from the Docker container [being used](https://github.com/zendesk/maxwell/blob/master/Dockerfile#L1), `maven:3.6-jdk-11`. There's a `maven:3.8-jdk-11` available on [DockerHub](https://hub.docker.com/layers/library/maven/3.8-jdk-11/images/sha256-37a94a4fe3b52627748d66c095d013a17d67478bc0594236eca55c8aef33ddaa?context=explore)
Edit: the new maven image does not fix the lib* dependencies, they will have to be fixed independently

[...]

PR #1945, opened 2022-11-19
GH requested security updates

osheroff:
https://github.com/zendesk/maxwell/issues/1924

LMK if this works for you; i didn't pull in your apt installs that pinned libc, but hopefully just the apt-get update will pull in the latest?


mayamessinger:
Thanks so much for handling this @osheroff!

I think another `apt-get update` and upgrade is needed after line 22 in the Dockerfile, since pulling a new image (`FROM openjdk:11-jdk-slim`) will not keep the library updates done earlier. Java deps are all good though!

osheroff:
ok good call.

Issue #1946, opened 2022-11-22
Multiple HA deployment modes


wanghangyu817:
maxwell is currently highly available through jgroups, which is a good choice.  However, in many cases, users are not familiar with jgroups.  On the contrary, users who deploy maxwell in a big data architecture will be more familiar with zookeeper.  Whether maxwell can co-exist with multiple HA deployment modes?  Users can choose to use the distributed coordination technology of the underlying high availability.  If so, I will submit maxwellHA on ZK

osheroff:
sure, if you like.  Honestly kubernetes is a much better answer here but hey

wanghangyu817:
Yes, maxwell is more flexible by providing users with a variety of options

PR #1948, opened 2022-11-28
maxwellHA on zookeeper

wanghangyu817:
#1946 
zookeeper is used to implement maxwell's high availability mode.

Sorry, I modified high_availability.md and added the usage mode. If it does not conform to the document specification, I will roll it back.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellConfig.java
@@ -609,9 +609,9 @@ public class MaxwellConfig extends AbstractConfig {
 	public Scripting scripting;
 
 	/**
-	 * Enable high available support (via jgroups-raft)
+	 * Enable high available support (via jgroups-raft or zookeeper)
 	 */
-	public boolean haMode;
+	public String haMode;
 
 	/**
 	 * Path to raft.xml file that configures high availability support
@@ -628,6 +628,32 @@ public class MaxwellConfig extends AbstractConfig {
 	 * {@link com.zendesk.maxwell.replication.BinlogConnectorReplicator}
 	 */
 	public int binlogEventQueueSize;
+
+	/**
+	 * HA zookeeper address
+	 */
+	public String zookeeperServer;
+
+	/**
+	 * session time
+	 */
+	public int zookeeperSessionTimeoutMs;
+
+	/**
+	 * connection time
+	 */
+	public int zookeeperConnectionTimeoutMs;
+
+	/**
+	 * maxRetries
+	 */
+	public int zookeeperMaxRetries;
+
+	/**
+	 * retryWaitMs
+	 */
+	public int zookeeperRetryWaitMs;
+
 
 	/**
 	 * Build a default configuration object.
@@ -741,12 +767,22 @@ public class MaxwellConfig extends AbstractConfig {
 				.withRequiredArg();
 		parser.separator();
 
-		parser.accepts( "ha", "enable high-availability mode via jgroups-raft" )
-				.withOptionalArg().ofType(Boolean.class);
+		parser.accepts( "ha", "enable high-availability mode via jgroups-raft or zookeeper" )
+				.withOptionalArg();
 		parser.accepts( "jgroups_config", "location of jgroups xml configuration file" )
 				.withRequiredArg();
 		parser.accepts( "raft_member_id", "raft memberID.  (may also be specified in raft.xml)" )
 				.withRequiredArg();
+		parser.accepts("zookeeper_server","enable maxwell High Availability using zookeeper")
+				.withRequiredArg();
+		parser.accepts("zookeeper_session_timeout_ms","session timeout duration (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("zookeeper_connection_timeout_ms","connection timeout duration (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("zookeeper_max_retries","maximum retry (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("zookeeper_retry_wait_ms","initial retry wait time (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
 
 		parser.separator();
 
@@ -1206,10 +1242,16 @@ public class MaxwellConfig extends AbstractConfig {
 
 		setupEncryptionOptions(options, properties);
 
-		this.haMode = fetchBooleanOption("ha", options, properties, false);
+		this.haMode = fetchStringOption("ha", options, properties, null);
 		this.jgroupsConf = fetchStringOption("jgroups_config", options, properties, "raft.xml");
 		this.raftMemberID = fetchStringOption("raft_member_id", options, properties, null);
 		this.replicationReconnectionRetries = fetchIntegerOption("replication_reconnection_retries", options, properties, 1);
+
+		this.zookeeperServer = fetchStringOption("zookeeper_server", options, properties, null);
+		this.zookeeperSessionTimeoutMs = fetchIntegerOption("zookeeper_session_timeout_ms", options, properties, 6000);
+		this.zookeeperConnectionTimeoutMs = fetchIntegerOption("zookeeper_connection_timeout_ms", options, properties, 6000);
+		this.zookeeperMaxRetries = fetchIntegerOption("zookeeper_max_retries", options, properties, 3);
+		this.zookeeperRetryWaitMs = fetchIntegerOption("zookeeper_retry_wait_ms", options, properties, 1000);
 
 		this.binlogEventQueueSize = fetchIntegerOption("binlog_event_queue_size", options, properties, BinlogConnectorReplicator.BINLOG_QUEUE_SIZE);
 	}
--- a/src/main/java/com/zendesk/maxwell/MaxwellHA.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellHA.java
@@ -1,5 +1,9 @@
 package com.zendesk.maxwell;
 
+import com.zendesk.maxwell.util.CuratorUtils;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.recipes.leader.LeaderLatch;
+import org.apache.curator.framework.recipes.leader.LeaderLatchListener;
 import org.jgroups.JChannel;
 import org.jgroups.protocols.raft.Role;
 import org.jgroups.raft.RaftHandle;
@@ -9,13 +13,15 @@ import org.slf4j.LoggerFactory;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
- * Class that joins a jgroups-raft cluster of servers
+ * Class that joins a jgroups-raft cluster of servers or zookeeper
  */
 public class MaxwellHA {
 	static final Logger LOGGER = LoggerFactory.getLogger(MaxwellHA.class);
 
 	private final Maxwell maxwell;
-	private final String jgroupsConf, raftMemberID, clientID;
+	private String jgroupsConf, raftMemberID, clientID;
+	private String zookeeperServer;
+	private int sessionTimeoutMs, connectionTimeoutMs, maxRetries, baseSleepTimeMs;
 	private boolean hasRun = false;
 	private AtomicBoolean isRaftLeader = new AtomicBoolean(false);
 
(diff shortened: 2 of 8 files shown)


src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -741,12 +771,24 @@ protected MaxwellOptionParser buildOptionParser() {
...
+		parser.accepts("session_time_out_ms","session timeout duration (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("connection_time_out_ms","connection timeout duration (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("max_retries","maximum retry (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("base_sleep_time_ms","initial retry wait time (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
+		parser.accepts("ha_modality","high availability mode : jgroups-raft or zookeeper")
+				.withRequiredArg();
```
osheroff, line 791: instead of introducing a new parameter, let's make the `ha` parameter accept and optional server string.  So:

`--ha` still enables HA via raft, but `--ha=zookeeper` enables HA via ZK.

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -741,12 +771,24 @@ protected MaxwellOptionParser buildOptionParser() {
...
+		parser.accepts( "ha", "enable high-availability mode via jgroups-raft or zookeeper" )
 				.withOptionalArg().ofType(Boolean.class);
 		parser.accepts( "jgroups_config", "location of jgroups xml configuration file" )
 				.withRequiredArg();
 		parser.accepts( "raft_member_id", "raft memberID.  (may also be specified in raft.xml)" )
 				.withRequiredArg();
+		parser.accepts("zookeeper_server","enable maxwell High Availability using zookeeper")
+				.withRequiredArg();
+		parser.accepts("session_time_out_ms","session timeout duration (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
```
osheroff, line 783: rename `zookeeper_session_timeout_ms`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -741,12 +771,24 @@ protected MaxwellOptionParser buildOptionParser() {
...
+		parser.accepts("connection_time_out_ms","connection timeout duration (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
```
osheroff, line 785: rename `zookeeper_connection_timeout_ms`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -741,12 +771,24 @@ protected MaxwellOptionParser buildOptionParser() {
...
+		parser.accepts("max_retries","maximum retry (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
```
osheroff, line 787: rename `zookeeper_max_retries`

src/main/java/com/zendesk/maxwell/MaxwellConfig.java
```diff
@@ -741,12 +771,24 @@ protected MaxwellOptionParser buildOptionParser() {
...
+		parser.accepts("base_sleep_time_ms","initial retry wait time (maxwellHA on zk)")
+				.withRequiredArg().ofType(Integer.class);
```
osheroff, line 789: rename `zookeeper_retry_wait_ms`

src/main/java/com/zendesk/maxwell/MaxwellHA.java
```diff
@@ -83,4 +109,43 @@ public void startHA() throws Exception {
 
 		Thread.sleep(Long.MAX_VALUE);
 	}
+
+	/**
+	 * indicates that Ha is started in zookeeper mode
+	 * @param flag the overloading of the method facilitates the subsequent expansion of multiple modes
+	 * @throws Exception
+	 */
+	public void startHA(String flag) throws Exception {
```
osheroff, line 118: rename `startHAZookeeper`

osheroff: (and rename the other method to `startHAJGroups`)

src/main/java/com/zendesk/maxwell/MaxwellHA.java
```diff
@@ -83,4 +109,43 @@ public void startHA() throws Exception {
...
+			@Override
+			public void isLeader() {
+				cu.register();
+				run();
+				LOGGER.debug("starting maxwell");
+			}
+
+			@Override
+			public void notLeader() {
+
```
osheroff, line 144: put an `info` statement here so the standby instances print something.

src/main/java/com/zendesk/maxwell/MaxwellHA.java
```diff
@@ -83,4 +109,43 @@ public void startHA() throws Exception {
...
+		cu.init();
+		CuratorFramework client = cu.getClient();
+		LeaderLatch leader = new LeaderLatch(client, cu.getElectPath());
+		leader.start();
+		leader.addListener(new LeaderLatchListener() {
+			@Override
+			public void isLeader() {
+				cu.register();
+				run();
+				LOGGER.debug("starting maxwell");
```
osheroff, line 139: make this an info message like:

```
LOGGER.info("node is current leader, starting Maxwell")
```

src/main/java/com/zendesk/maxwell/util/CuratorUtils.java
```diff
@@ -0,0 +1,130 @@
...
+
+	public CuratorFramework getClient() {
+		return client;
+	}
+
+	public void register() {
+		try {
+			String rootPath = masterPath;
+			String hostAddress = InetAddress.getLocalHost().getHostAddress();
+			client.create().creatingParentsIfNeeded().withMode(CreateMode.EPHEMERAL_SEQUENTIAL).forPath(rootPath + "/" + hostAddress);
```
osheroff, line 111: I don't like that we're pulling in all of `org.apache.zookeeper` for just the `EPHEMERAL_SEQUENTIAL` constant.  Is there a way around that?

src/main/java/com/zendesk/maxwell/util/CuratorUtils.java
```diff
@@ -0,0 +1,130 @@
...
+			LOGGER.error("register exception", e);
+		}
+	}
+
+	public List<String> getChildren(String path) {
+		List<String> childrenList = new ArrayList<>();
+		try {
+			childrenList = client.getChildren().forPath(path);
+		} catch (Exception e) {
+			LOGGER.error("There was an error getting the child nodes", e);
```
osheroff, line 122: same here, don't catch this exception here

osheroff:
the docs need work but I can do some rewrites when this is ready to merge.  thanks for the contribution!

src/main/java/com/zendesk/maxwell/util/CuratorUtils.java
wanghangyu817: Is the jar package too large or something else

osheroff: it's 1MB yeah.

wanghangyu817: all the ways I can think of need to refer to this dependency

osheroff: ok the curator-client depends on it anyway.  nevermind.

kovyrin:
Hey, @wanghangyu817! Thank you for your contribution!

We're thinking about different options for coordinating multiple instances of Maxwell in our infrastructure (mostly happens during k8s cluster failovers, etc) and using Zookeeper feels like a good option for us.

I'm planning to try it out over the next couple of days and will report if we see any issues.

src/main/java/com/zendesk/maxwell/MaxwellHA.java
```diff
@@ -83,4 +109,51 @@ public void startHA() throws Exception {
...
+					System.exit(1);
+				}
+				LOGGER.info("node is current leader, starting Maxwell....");
+				run();
+				cu.stop();
+			}
+
+			@Override
+			public void notLeader() {
+				//LeaderLatch.CloseMode.SILENT mode will not invoke this method
```
kovyrin, line 152: Are you expecting this to never happen or what makes it OK for us to ignore the loss of leadership here?

src/main/java/com/zendesk/maxwell/MaxwellHA.java
```diff
@@ -128,28 +137,48 @@ public void startHAZookeeper() throws Exception {
...
 			}
 
 			@Override
 			public void notLeader() {
-				//LeaderLatch.CloseMode.SILENT mode will not invoke this method
+				try {
+					lock.lock();
+					LOGGER.warn("node:" + hostAddress + " lost leader");
+					LOGGER.warn("master-slave switchover......");
+					LOGGER.warn("The leadership went from " + hostAddress + " to " + leader.getLeader());
```
kovyrin, line 176: Should this case shut down the current maxwell process given that it has lost the leadership status?

wanghangyu817: I don't understand what you mean

kovyrin: Sorry, I mean by the time we get this call, it means the current Maxwell process has lost leadership. That means we should probably stop the process altogether or shut down the replicator and go back into the election mode waiting for our turn once again.

Logging a warn does nothing and means we are going to keep replicator threads running and pumping duplicate data into whatever producer is configured. Additionally, the positions store will start getting conflicting writes from two different processes.

@osheroff Do I understand it correctly?

wanghangyu817: I understand. I have tested this point. When I try to kill the process, the code will go straight to this location and print the results we want, and also print the node information of the next leader. If there are other exceptions that make it impossible to execute this code, please tell me and I will solve it

kovyrin: I'll leave it to Ben to make a call on what should happen in this case, but I feel simply shutting down the process gracefully may be the easiest way to avoid conflicts after a leadership loss. Alternatively, it may be possible to call `maxwell.terminate();` the way the JGroups-based HA implementation does it.

wanghangyu817: In my opinion, when the `leader` becomes a `follower`, it only means that the zookeeper connection is abnormal or the current node does not meet the conditions of the leader. During this period, the program does not care about maxwell connector status (of course, maxwell status monitoring can be added later). It's a simple switch from one node to another. If the zookeeper connection problem causes the master/slave switchover, the program will not quit, but become a `follower`. Next, I will add maxwell to the indicator monitoring in iterations to determine whether to switch the master/slave according to these indicators, which requires Ben @osheroff  guidance

kovyrin: >  During this period, the program does not care about maxwell connector status
If we have lost ZK connection and, consequently, lost the leader status, then the current Maxwell instance will at the very least start producing duplicate events since the other instance that is the leader now is replicating the same set of changes already. Additionally, there is a chance of both instances overwriting each other's position information in maxwell's database, which AFAIU can have negative consequences as well.


> If the zookeeper connection problem causes the master/slave switchover, the program will not quit, but become a follower

What do you feel a follower Maxwell instance should do? (AFAIU, there is no notion of a follower mode in current maxwell codebase)

wanghangyu817: For example,if I start three maxwell instances,we label them as 1,2,3, where 1 is the leader and 2,3 are the followers. When 1 is the leader,2,3 is just a daemon process that doesn't do anything. When 1 exit the leader(not the exit caused by maxwell,but the server failure:For example,exit caused by restart,memory overflow,disk space, etc.),then 2 or 3 takes over from 1 to continue the collection task. If the maxwell process caused by mysql exits,no matter how many instances are started, the problem still presists. This is not something that can be fixed by high availability. What I need to do is to ensure that maxwell itself is highly available.

kovyrin: Your solution absolutely solves the scenarios described. One thing I feel it is missing (or I may be confused!) is a scenario when a leader, doing its leader stuff, replicating data data, etc, loses its leadership while remaining alive and seemingly healthy (due to ZK connectivity issues, ZK restart, any other issues that force a new election). In those cases the old leader needs to step down and stop doing its usual leader things and move to a quiet follower mode (stop binlog replicator, don't write into the position store anymore, etc).

wanghangyu817: I have tested the scenario you described and got corresponding results.  Please let me know if you have any other problems

PR #1957, opened 2022-12-13
Fix NPE when reading null values from env_config

redcape:
Ran into this when the default config got a null value added to it.

1) It was hard to debug since it didn't show up in the regular logs, only in stderr.
2) Due to the way the configuration for our custom producers are setup, we don't have a good way to prevent these null values from getting added

Fix:
 - Don't add null values to the map. When we `get` them, it will come back as null anyway.
 - Add logging of the exception/stack trace and added stack traces to logs around it. If maxwell is going to shutdown the process, I'd rather have a stacktrace than not if I need to debug what's going on.

Add unit test for null value case.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/Maxwell.java
+++ b/src/main/java/com/zendesk/maxwell/Maxwell.java
@@ -343,13 +343,14 @@ public class Maxwell implements Runnable {
 		} catch ( URISyntaxException e ) {
 			// catch URISyntaxException explicitly as well to provide more information to the user
 			LOGGER.error("Syntax issue with URI, check for misconfigured host, port, database, or JDBC options (see RFC 2396)");
-			LOGGER.error("URISyntaxException: " + e.getLocalizedMessage());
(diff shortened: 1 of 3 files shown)


src/main/java/com/zendesk/maxwell/Maxwell.java
```diff
@@ -343,13 +343,14 @@ public void run() {
...
-			LOGGER.error("URISyntaxException: " + e.getLocalizedMessage());
+			LOGGER.error("URISyntaxException: " + e.getLocalizedMessage(), e);
 			System.exit(1);
 		} catch ( ServerException e ) {
-			LOGGER.error("Maxwell couldn't find the requested binlog, exiting...");
+			LOGGER.error("Maxwell couldn't find the requested binlog, exiting...", e);
 			System.exit(2);
 		} catch ( Exception e ) {
 			e.printStackTrace();
+			LOGGER.error("Maxwell saw an exception and is exiting...", e);
```
osheroff, line 353: this one feels superflous due to the log line above.  Or is it that `printStackTrace` just dumps to stdout?

PR #1969, opened 2023-01-16
Parse "ADD COLUMN IF NOT EXISTS" which is supported by MariaDB (fixes #1968)

### LINKED ISSUE #1968: Error parsing SQL: 'ALTER TABLE task ADD COLUMN IF NOT EXISTS
maxwell version 1.35.3
Maxwell has stopped running, probably because of the error below. An SQL validator like https://www.eversql.com/sql-syntax-check-validator/ seems to suggest that the "IF NOT EXISTS" addition is invalid even though MariaDB has accepted the query.

Could you maybe give us some pointers on how we can make Maxwell accept this query?

```
ERROR MysqlParserListener - (alter_partition_specification ADD COLUMN IF NOT EXISTS deleted_at DATETIME NULL)
ERROR SchemaChange - Error parsing SQL: 'ALTER TABLE task ADD COLUMN IF NOT EXISTS deleted_at DATETIME NULL'
ERROR AbstractSchemaStore - Error on bin log position Position[BinlogPosition[master.006382:12395005], lastHeartbeat=0]
```

bartvw:
Fixes #1968

### DIFF
--- a/src/main/antlr4/imports/mysql_alter_table.g4
+++ b/src/main/antlr4/imports/mysql_alter_table.g4
@@ -25,8 +25,8 @@ alter_specification:
   ;
 
 // the various alter_table commands available
-add_column: ADD COLUMN? column_definition col_position?;
-add_column_parens: ADD COLUMN? '(' (column_definition|index_definition) (',' (column_definition|index_definition))* ')';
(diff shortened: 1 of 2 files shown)


osheroff:
lgtm, sorry for delay

Issue #1973, opened 2023-02-08
query field missing from some transactions


mavendonovanhubbard:
For some reason the `query` field is missing from about half of our transactions. All the other fields are present and Maxwell does not seem to be logging any errors related to this. It only occurs in our production environment and we cannot reproduce it. We've tried to find any correlations to see what types of transactions are failing, whether it's certain tables or such but we haven't found any. 

Below are the fields we get when the query field is missing
```json
{
    "database":... ,
    "table": ...,
    "type": ...,
    "ts": ..,
    "xid": ...,
    "xoffset": ...,
    "position": ...,
    "push_ts": ...,
    "primary_key": ...,
    "primary_key_columns": ...,
    "data": { ... },
    "old": { ... }
}
```

Below is what we get when the query field is present
```json
{
    "database":... ,
    "table": ...,
    "query": ...,
    "type": ...,
    "ts": ..,
    "xid": ...,
    "xoffset": ...,
    "position": ...,
    "push_ts": ...,
    "primary_key": ...,
    "primary_key_columns": ...,
    "data": { ... },
    "old": { ... }
}
```

osheroff:
did you set the `binlog_rows_query_log_events` while the server was still running?  any chance you have very-long-lived connections?  

I ask because clients get the value of global variables (and it controls the binlog behavior) when they first establish their server connection.

mavendonovanhubbard:
`output_row_query` is set to true. The maxwell process is running inside a kubernetes container so I killed and restarted the pod to turn it on. The pod restarts with some regularity so I doubt the problem is long lived connections. I imagine the problem isn't the `output_row_query` setting as it seems to be working for some transactions just not all.

osheroff:
it's not the maxwell process that needs to restart but *other* connections -- the ones actually generating the transactions.

mavendonovanhubbard:
[...]

PR #1974, opened 2023-02-09
No longer clears currentQuery after making a change to the db.

mavendonovanhubbard:
Some MySQL transactions are missing query data even though `output_row_query` is set to true, issue #1973. 

When Maxwell reads from the binlog, the `BinlogConnectorReplicator` class calls [getTransactionRows ](https://github.com/zendesk/maxwell/blob/56f458dd2eee3651a82cfbd1c365b6d24f937404/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java#L537-L637) for each transaction. The function enters a while (true) loop as it reads from the events, and when it receives a `ROWS_QUERY` event it sets the `currentQuery` local variable. When the it receives a database write operation such as `EXT_UPDATE_ROWS` it uses the `currentQuery` variable to get the query information, but then it clears the `currentQuery` variable. This causes a problem if there are any other write operations that occur before another `ROWS_QUERY` event does. Those events will have null query values.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -586,7 +586,6 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 								buffer.add(r);
 							}
 					}
-					currentQuery = null;
 					break;
 				case TABLE_MAP:
 					TableMapEventData data = event.tableMapData();


osheroff:
nice fix, but I feel like we should clear `currentQuery` at *some point* in the code -- otherwise if we have a mixed binlog stream where some events have the query data and some don't, we'll end up assigning bad sql to a query.

maybe clear it on getting a TABLE_MAP_EVENT?  is that correct?  the other option is when seeing the COMMIT.

mavendonovanhubbard:
The transactions I'm seeing typically follow this sequence:

1. ROWS_QUERY
2. TABLE_MAP
3. a write event like EXT_DELETE_ROWS

So we can't have the TABLE_MAP event clear it because then it would never get to the write events where it's needed.

Despite the fact that this is in a while true loop, the function does actually exit when it finds a commit event [here](https://github.com/zendesk/maxwell/blob/56f458dd2eee3651a82cfbd1c365b6d24f937404/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java#L562) so there is no risk of a query bleeding into another transaction. 

I don't have anything with a MIXED binlog on hand, so I'm not sure what the behavior is. Each transaction is isolated though.

osheroff:
roger that.  little rusty.  your pr is fine then

PR #1977, opened 2023-02-14
Rename pk bugfix

trallarn:
Bugfix renaming pk column

Renaming a primary key column yielded the following error:

```
java.lang.RuntimeException: Couldn't serialize [com.zendesk.maxwell.schema.ddl.ResolvedTableAlter@29a69a35] to JSON.
```

The issue was that the renamed column wasn't reflected in the internal
data structure of primary keys.

This commit updates the internal primary key data structure when RenameColumnMod
is applied in the same way it is updated when a ChangeColumnMod is applied.


trallarn:
@osheroff This PR fixes a bug when processing a "RENAME COLUMN" statement on a primary key column. We are currently running with our fork that has this fix applied and I would love to merge it to the main repository.

PR #1986, opened 2023-03-09
Proposal for a custom maxwell health check factory

crispmark:
Right now users are not able to tailor a healthcheck to their individual needs. At my company we have tried running checks through a shell script alongside the health check, but it seems that this code execution is much slower in some cases and not as reliable as native java.

This is a proposal based on this longstanding TODO
https://github.com/zendesk/maxwell/blob/master/src/main/java/com/zendesk/maxwell/monitoring/MaxwellHealthCheck.java#L17

and uses the same strategy employed for custom producers.


osheroff:
it's ok by me.  can you add some textual documentation around healthchecks and custom healthchecks here? https://github.com/zendesk/maxwell/blob/08a7bd5079eeae0f7a8f322dc12b90a02ceea2e0/docs/docs/monitoring.md#L41

crispmark:
@osheroff thanks for the quick review, I have added more documentation and a unit test around fetching the factory based on the config.

PR #1994, opened 2023-04-04
Row query max length

mavendonovanhubbard:
In some cases we've found that SQL queries that run in our environment can be extremely long. This causes problems as it bumps us over the limit for the max size permitted by AWS SNS. I've added the following enhancement to allow users to configure the max length of the output SQL query. Any queries that are longer than this length will be truncated. We've been running this in production for about a month and a half now with no problems.


osheroff:
cool, thanks for the PR!

PR #2008, opened 2023-05-17
enable localstack support for sns

nenlpl:
there is likely a cleaner way to do this but I followed the same pattern that was approved for SQS support - https://github.com/zendesk/maxwell/pull/1868


osheroff:
Hi, I merged this before realized that it fails the build.  happy to merge but fix tests first

Issue #2030, opened 2023-07-26
MySQL ANTLR grammar doesn't recognize "rename tables" (plural)


s-huk:
The DDL command `rename tables ...` (table**s** in plural) causes the following Maxwell error:

```
2023-07-26 14:21:59 ERROR SchemaChange - Error parsing SQL: 'rename tables MyTableA to _trash_MyTableA, MyTableB to MyTableA'
2023-07-26 14:21:59 ERROR MysqlParserListener - (rename_table rename <missing TABLE> (rename_table_spec (table_name (name (id tables))) MyTableA to (table_name (name (id _trash_MyTableA)))) , (rename_table_spec (table_name (name (id MyTableB))) to (table_name (name (id MyTableA)))))
```

So, your grammar just recognizes `rename table ...`, not `rename tables`. 

As a result, Maxwell doesn't work anymore as soon as this command occurs in the MySQL binlogs.

I will propose a fix.

PR #2031, opened 2023-07-26
Fix ddl grammar to recognize 'RENAME TABLES'

s-huk:
Proposed solution for issue #2030


src/main/antlr4/imports/mysql_literal_tokens.g4
```diff
@@ -192,6 +192,7 @@ STORED: S T O R E D;
 SUBPARTITION: S U B P A R T I T I O N;
 SUBPARTITIONS: S U B P A R T I T I O N S;
 TABLE: T A B L E;
+TABLES: T A B L E S;
```
osheroff, line 195: this file is autogenerated from `generate_tokens.rb`, plz add this token there

s-huk: ok, I kept it consistent

s-huk: everything ok like this?

Issue #2063, opened 2023-11-16
Allow persistent variables in Javascript filters


kellertobias:
In our company we have the issue that there are some big data jobs that should not end up in or cdc events.

Currently we have a date based filter in place, however that solution is not very good. 

We know when the data jobs start and when they end, because they start and end always with the same commands.

We would like to do something like this:

```
let currentlyDisabled = false;

function process_row(row) {
    ...
    // enable/ disabled based on changes
    if(first command of data job found) {
       currentlyDisabled = true
    }
    if(last command of data job found) {
       currentlyDisabled = false
    }
    if(currentlyDisabled) {
      row.suppress()
   }
}

```

osheroff:
reasonable enough, I think maybe easier would be to do like:

```
function process_row(row, state) {
    ...
    // enable/ disabled based on changes
    if(first command of data job found) {
       state.currentlyDisabled = true
    }
    if(last command of data job found) {
       state.currentlyDisabled = false
    }
    if(state.currentlyDisabled) {
      row.suppress()
   }
}
```

any desire to take a pass at a PR?

kellertobias:
Well, I am usuailly a typescript/ node developer and also do not have direct access to our deployment of maxwell at the moment, just to the repo were our filter scripts lay in.

kellertobias:
Let's do the same for the DDL function

kellertobias:
Added yet another PR: https://github.com/zendesk/maxwell/pull/2066

PR #2064, opened 2023-11-17
add state object for scripting engine

kellertobias:
This is my first shot on tackling https://github.com/zendesk/maxwell/issues/2063

What this changes:
- it introduces a new state that can be carried over multiple calls of `process_row`
- It adds tests for scripting engine and the new changes


osheroff:
well, can you install java?  This looks plausible, but it doesn't compile...

kellertobias:
@osheroff I have finally found the time to install java and fix the problem. I also have written some tests and updated the documentation.

osheroff:
thanks keller.

PR #2084, opened 2024-02-15
Add aws-msk-iam-auth dependency to pom.xml

turuncuofke:


turuncuofke:
With this "aws-msk-iam-auth" library, maxwell can connect to AWS MSK clusters that are managed by iam authentication when configured to do so. 

You can check this link for extra information: https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html

PR #2113, opened 2024-09-25
Updated README.md to fix example

gpsinghsandhu:
Corrected the example which wrongly showed update query to produce type='insert' event.


osheroff (approved):
how did this live for so many years?

gpsinghsandhu:
@osheroff Amazing isn't it 😃

PR #2129, opened 2024-11-08
feat(datetime): enhance datetime parsing and validation

CoreyWinkelmannPP:
Add validation and parsing for DateColumnDef and DateTimeColumnDef to handle invalid date formats and zero dates as null. Introduce a DateValidator with regex matching for date formats. Extend tests to cover invalid dates and malformed datetime strings.

MySQL and MariaDB allow invalid dates by default unless strict SQL mode is on, enabling entries like 2020-00-01 to be stored without error, which can lead to data processing issues. In older systems, data may be in poor condition, making it difficult to switch to strict mode, so these cases often need handling when using CDC.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
+++ b/src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
@@ -32,15 +32,19 @@ public class DateColumnDef extends ColumnDef {
 		if ( value instanceof String ) {
 			String dateString = (String) value;
 
-			if ( config.zeroDatesAsNull && "0000-00-00".equals(dateString) )
-				return null;
+			if ("0000-00-00".equals(dateString)) {
+				if ( config.zeroDatesAsNull  )
+					return null;
+				else
+					return "0000-00-00";
+			} else {
+				if ( !DateValidator.isValidDateTime(dateString) )
+					return null;
 
-			if ( !DateValidator.isValidDateTime(dateString) )
-				return null;
-
-			value = parseDate(dateString);
-			if (value == null) {
-				return null;
+				value = parseDate(dateString);
+				if (value == null) {
+					return null;
+				}
 			}
 		} else if ( value instanceof Long && (Long) value == Long.MIN_VALUE ) {
 			if ( config.zeroDatesAsNull )
--- a/src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
+++ b/src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
@@ -1,8 +1,14 @@
 package com.zendesk.maxwell.schema.columndef;
 
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
 import com.zendesk.maxwell.producer.MaxwellOutputConfig;
 
 public class DateColumnDef extends ColumnDef {
+    private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+
 	private DateColumnDef(String name, String type, short pos) {
 		super(name, type, pos);
 	}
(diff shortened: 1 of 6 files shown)


osheroff:
what was the behavior of the bootstrapper wrt to "0000-00-00" style dates (if the as-null flag was off) before this change?  

I know, it absolutely sucks to output invalid datetimes into the stream but all things equal the `0000-00-00` case is so common I'd prefer not to change that behavior without changing a flag.  I'm good with your other changes that nullify stuff like zero months tho.

CoreyWinkelmannPP:
I isolated the date validation logic outside of the all zero case and I believe kept the logic of the all zero case the same as what existed before. From my testing locally it seems to work in the same way as before.

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
```diff
@@ -19,7 +23,24 @@ public static DateTimeColumnDef create(String name, String type, short pos, Long
...
+		if ( value instanceof String) {
+			String dateString = (String) value;
+
+			if ( "0000-00-00 00:00:00".equals(dateString) ) {
+				if ( config.zeroDatesAsNull )
+					return null;
+				else 
+					return appendFractionalSeconds("0000-00-00 00:00:00", 0, getColumnLength());
+			} else {
+				if ( !DateValidator.isValidDateTime(dateString) )
```
osheroff, line 35: sorry for the delay, I'm coming back to this PR since I'm prepping a release of your mariadb stuff and some other stuff too.  I'm a little worried about running a regex in a very hot code-path...

But I guess this only runs on bootstrapping so it might be ok?  can you confirm that?

src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
```diff
@@ -24,11 +30,22 @@ public String toSQL(Object value) {
...
+					return "0000-00-00";
+			} else {
+				if ( !DateValidator.isValidDateTime(dateString) )
+					return null;
+
+				value = parseDate(dateString);
+				if (value == null) {
+					return null;
+				}
+			}
```
osheroff, line 48: mildly, same concerns here -- is this just for bootstrapping?

osheroff:
would be nice to slap some integration tests around this code (esp checking behavior in bootstrapping, which I believe is where this change targets).

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
CoreyWinkelmannPP: Sorry, just making it back to looking at this to see where it was at. From my testing, the value instanceof String is only triggered through the bootstrapping and not during normal processing. It goes down the Long path at least from my debugging I was doing. If it is just bootstrapping are you good with it or should I look for a more efficient way to verify that Date/DateTime instead of using Regex?

src/main/java/com/zendesk/maxwell/schema/columndef/DateColumnDef.java
CoreyWinkelmannPP: I am only seeing bootstrapping taking the String path. Otherwise, during normal processing it takes the Long path.

src/main/java/com/zendesk/maxwell/schema/columndef/DateTimeColumnDef.java
osheroff: hmmm yeah I don't know.  I had a performance benchmark setup at some point, let me see if I can maybe measure the impact of the regex on bootstrapping.

osheroff:
sorry to say it's a big perf hit.

Bootstrapping a table with 2 timestamps, 2 datetimes:

before your patch:

```
Producer received 1000076 rows in 8.22 seconds
121738 rows per second
```

after your patch:

```
Producer received 1000082 rows in 39.52 seconds
25307 rows per second
```

I tried to profile it in visualVM for more proof but it turns out that profiler is lousy.  I pushed the bootstrap benchmarker into your branch, LMK if you need pointers in running it

CoreyWinkelmannPP:
I am going to close this for now. I will look at ways to improve the full load process timing as we have time to look into it and then submit something else down the road. Thanks!

PR #2132, opened 2024-11-12
refactor(mysql-status): Refined compatibility checks for MySQL and Ma…

CoreyWinkelmannPP:
…riaDB versions

This PR updates the getShowBinlogSQL method to enhance compatibility and accuracy in determining the correct SQL command for checking binary log status based on the database type and version.

MySQL 8.2+ uses SHOW BINARY LOG STATUS
MariaDB 10.5+ uses SHOW BINLOG STATUS
Older or unspecified versions default to SHOW MASTER STATUS

### DIFF
--- a/src/main/java/com/zendesk/maxwell/MaxwellMysqlStatus.java
+++ b/src/main/java/com/zendesk/maxwell/MaxwellMysqlStatus.java
@@ -34,11 +34,40 @@ public class MaxwellMysqlStatus {
 		}
 	}
 
+       /**
+       * Generates the appropriate SQL command to retrieve binary log status based on
+       * the database type and version.
+       *
+       * This method checks the database product name and version to determine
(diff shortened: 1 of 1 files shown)


osheroff:
please undo all the whitespace changes your editor put in, the code changes look fine and mergable, thanks

CoreyWinkelmannPP:
Sorry, going to need to hold off on this. I need to better handle how the `String productName = md.getDatabaseProductName();` returns mysql whether mariadb or not. I have to use something similar to what the isMaria method is doing. But need to test if I can also verify mysql or if I need to do that differently as well.

CoreyWinkelmannPP:
Alright, I am using a similar technic as the isMaria function. And then for now just creating a helper isMySQL variable to handle the opposite of the mariadb check. I tested it out and it was working for mysql and mariadb locally.

PR #2133, opened 2024-11-14
refactor(schema): add database filtering to JSON detection

CoreyWinkelmannPP:
Updated `detectMariaDBJSON` to accept `dbName` as a parameter, enabling filtering by `CONSTRAINT_SCHEMA` for more precise JSON constraint detection within specified databases.

The INFORMATION_SCHEMA tables, especially in production, can be a bottleneck if queried without constraints due to the high volume of metadata. By narrowing down the scope of JSON validation queries, we reduce potential load on INFORMATION_SCHEMA, improving efficiency and response times. This targeted approach allows for more precise schema capture while avoiding the pitfalls of exhaustive queries, especially in systems with many databases or tables.


osheroff:
again, too much whitespace churn.  does this really have a noticable impact on the database in a production system or is this kinda a theoretical win?

CoreyWinkelmannPP:
Sorry, I will stop my neovim from updating the whitespace on PRs. This affects our systems and we do need it. If we don't focus our queries to the information schema we do experience issues.

PR #2139, opened 2024-12-16
feat: add partitioner for sns and sqs

m1heng:
Add a parttioner for sns and sqs, so they can choose which key to used for GroupId (where is actually play as the role as partition key like kafka)


osheroff:
looks good.  can you update the docs to reflect this change?

m1heng:
@osheroff just added some doc to mark sqs/sns supports current producer partition config. Let me know if I missed any.

Issue #2098, opened 2024-06-13
Maxwell v1.41.2 which based on openjdk:11-jkd-slim is vulnerable and outdated


gstamatov-gl:
Hello team,

I would like to bring to your attention that the base image openjdk:11-jdk-slim used in this project is outdated and contains several vulnerabilities. The last update for this image was 2 years ago, and it currently has 4 critical and 9 high vulnerabilities.

Using an outdated and vulnerable base image can pose significant security risks. It is recommended to update the base image to a more recent version or switch to a more secure alternative.

Link to openjdk:11-jkd-slim docker scanning:
https://hub.docker.com/layers/library/openjdk/11-jdk-slim/images/sha256-d2b6af2093e823ba0cdee4bcd45a905afe3fa054d08bde55b1d850515da69a08?context=explore

Thank you for your attention to this matter.
Regards, Georgi

PR #2140, opened 2024-12-19
fix: bump jdk version in the dockerfile

artemvovk:
https://github.com/zendesk/maxwell/issues/2098

use eclipse because it's easy part of the maven image

remove deprecated flag/option


osheroff:
thanks for the PR!  verifying & will merge

PR #2198, opened 2025-06-05
[GO-2025]: Updated kafka-clients to 3.7.1

PrajwalKathwate:
Description: Updated **kafka-clients** to **3.7.1** thereby addressing vulnerability [CVE-2024-31141](https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2Fzendesk%2Fmaxwell%2Fsecurity%2Fdependabot%2F109)

JIRA: https://zendesk.atlassian.net/browse/GO-2025
Risks: NA

Level: Low
Notes for Rollback: NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ clean:
 depclean: clean
 	rm -f $(CLASSPATH)
 
-package: depclean kafka-0.8.2.2 kafka-0.9.0.1 kafka-0.10.0.1 kafka-0.10.2.1 kafka-0.11.0.1 kafka-1.0.0 kafka-2.7.0 kafka-3.7.1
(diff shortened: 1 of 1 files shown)


osheroff:
two things here:

the first is i'm a little sketched out removing a kafka version that people might depend on in their scripts (3.4.0) -- historically we've kept around old client versions.  I guess if the upgrade from 3.4.0 -> 3.7.1 is all minor revs it's ok, but I don't want to be doing this all the time -- we'll break anybody specifying an explicit kafka version. 

(note to self, I guess i should write --kafka_version=3 support)

The second is that when you bump a kafka version there's other parts of the code base that you need to change, please `git grep '3.4.0'` and see what else needs to be upgraded.

PrajwalKathwate:
@osheroff  I agree that removing Kafka 3.4.0 outright could break users who explicitly specify that version, so for now I’ve added support for 3.7.1 but kept 3.4.0 available as well. 

I’ve also updated all references to supported Kafka versions in the Makefile, pom.xml, and MaxwellConfig.java to include 3.7.1.

PR #2245, opened 2026-02-28
Add automatic recovery from purged binlog files

CodeMarker0:
Summary

  This PR adds automatic recovery functionality when the requested binlog file has been purged from the MySQL server.
  Previously, when Maxwell attempted to reconnect to a binlog file that no longer existed on the server, it would fail
  with error code 1236 and require manual intervention.

  With this change, Maxwell can now detect when a binlog file is unavailable and automatically attempt to recover by
  connecting to the oldest available binlog file on the server.

  Problem Scenario

  Consider the following situation:

  1. MySQL binlog expiration is set to 7 days (binlog_expire_logs_seconds = 604800)
  2. However, due to low write traffic, the current binlog file takes 30 days to rotate to a new file
  3. When the binlog finally rotates, MySQL immediately purges the old file (since it's older than 7 days)
  4. Maxwell attempts to reconnect to the old binlog file that no longer exists
  5. Maxwell receives error 1236 (ER_MASTER_FATAL_ERROR_READING_BINLOG) and crashes
  6. Manual intervention is required to reset Maxwell's position

  This scenario is particularly problematic because:
  - It occurs unexpectedly after long periods of stable operation
  - The root cause (binlog rotation + immediate purge) is not obvious
  - Recovery requires manual inspection and position reset

  Changes

  - Added isBinlogFileAvailable() method to check if a binlog file exists on the server
  - Added getAvailableBinlogFiles() method to query all available binlog files
  - Added attemptRecoverFromAvailableBinlogs() method to handle recovery logic
  - Added getReplicationConnection() helper method for database queries
  - Modified tryReconnect() to handle binlog file not found errors (1236) and attempt recovery
  - Added pre-check in ensureReplicatorThread() to proactively detect unavailable binlogs

  Limitations & Known Risks

  ⚠️ Potential Data Loss: This recovery mechanism works by resetting Maxwell's position to the oldest available binlog
  file. Any events that existed between the original binlog position and the oldest available binlog will be permanently
   lost.

  Example:
  - Maxwell was reading binlog.000100 at position 12345
  - Server only has binlog.000150 through binlog.000200 available
  - Recovery will start from binlog.000150 position 4
  - All events from binlog.000100-000149 are lost

  Trade-off: This PR prioritizes service availability over data completeness. Maxwell will remain running and continue
  capturing new events, but there may be a gap in the data stream. Users should:

  - Monitor for recovery events in logs (search for "Attempting recovery" or "Successfully recovered")
  - Implement alerting when recovery occurs
  - Consider whether automatic recovery is appropriate for their use case
  - Potentially make this behavior configurable in a future PR

  Test Plan

  - Unit tests for new helper methods
  - Integration test simulating binlog purge scenario
  - Manual testing with a MySQL instance where binlogs are purged
  - Verify normal reconnection still works when binlogs are available

  Breaking Changes

  - tryReconnect() now throws Exception in addition to TimeoutException. Callers may need to handle the additional
  exception type.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorReplicator.java
@@ -31,6 +31,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -80,6 +85,7 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 	private Histogram transactionExecutionTime;
 
 	private final Boolean gtidPositioning;
+	private final MaxwellMysqlConfig mysqlConfig;
 
 	private static Pattern createTablePattern =
 		Pattern.compile("^CREATE\\s+TABLE", Pattern.CASE_INSENSITIVE);
@@ -222,6 +228,7 @@ public class BinlogConnectorReplicator extends RunLoopProcess implements Replica
 		this.client.setServerId(replicaServerID.intValue());
 
 		this.replicationReconnectionRetries = replicationReconnectionRetries;
+		this.mysqlConfig = mysqlConfig;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


osheroff:
this is just the wrong-thing-to-do in many many cases - if there's been any schema changes at all in between binlog files maxwell will crash as soon as it hits an update in the changed table.  Really the only safe thing to do in these cases is to boot maxwell with `--recapture-schema`.  

One *could* make an argument that you could turn on the footgun with a commandline flag, but my suspicion is that if your binlog has been deleted you wouldn't have been careful enough to enable that option anyway.  

So I don't think I see a path forward to making this mergable unless you have a strong argument about a manner in which it can not create more problems than it solves.

CodeMarker0:
> this is just the wrong-thing-to-do in many many cases - if there's been any schema changes at all in between binlog files maxwell will crash as soon as it hits an update in the changed table. Really the only safe thing to do in these cases is to boot maxwell with `--recapture-schema`.
> 
> One _could_ make an argument that you could turn on the footgun with a commandline flag, but my suspicion is that if your binlog has been deleted you wouldn't have been careful enough to enable that option anyway.
> 
> So I don't think I see a path forward to making this mergable unless you have a strong argument about a manner in which it can not create more problems than it solves.

Thanks for your reply. I agree. Skipping schema recapture is risky and can easily cause crashes when schema changes exist. I will close this PR.

PR #2252, opened 2026-05-12
Fix schema restore selecting stale snapshot due to lexicographic binlog file ordering

### LINKED ISSUE #2251: Schema restore picks wrong (stale) schema due to lexicographic string comparison of binlog file numbers
### Summary

When Maxwell restores its schema on startup (non-GTID mode), it uses a SQL query that compares  values **as strings**. Because MySQL's  comparison is lexicographic, binlog file numbers with varying digit counts sort incorrectly — causing Maxwell to silently select a **stale schema snapshot** that is missing DDLs that happened between the selected schema's position and the actual replication start position. This results in a crash at runtime:

```
java.lang.RuntimeException: Couldn't find table <table> in database <database>
  at com.zendesk.maxwell.replication.TableCache.processEvent(TableCache.java:35)
```

---

### Affected Code

**File:** `src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java`  
**Method:** `findSchema()` (around line 581)

```java
String sql = "SELECT id from `schemas` "
        + "WHERE deleted = 0 "
        + "AND last_heartbeat_read <= ? AND ("
        + "(binlog_file < ?) OR "                         // <-- string comparison
        + "(binlog_file = ? and binlog_position < ? and base_schema_id is not null) OR "
        + "(binlog_file = ? and binlog_position <= ? and base_schema_id is null) "
        + ") AND server_id = ? "
        + "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1";
                                                              // ^^ also string sort
```

The `binlog_file` column is `VARCHAR`, and both the `WHERE binlog_file < ?` filter and the `ORDER BY binlog_file DESC` sort operate lexicographically.

---

### Root Cause: Lexicographic vs Numeric Ordering

AWS Aurora / RDS uses binlog file names of the form `mysql-bin-changelog.NNNNNN` where the numeric suffix grows without zero-padding. Over time the suffix crosses digit-count boundaries (e.g., 6 digits → 7 digits), and string comparison then gives **wrong results**:

| File A | File B | Numeric order | String (lexicographic) order |
|---|---|---|---|
| `mysql-bin-changelog.122046` | `mysql-bin-changelog.1215501` | 122046 **<** 1215501 ✓ | `'122046'` **>** `'1215501'` ✗ |
| `mysql-bin-changelog.979715` | `mysql-bin-changelog.1215501` | 979715 **<** 1215501 ✓ | `'979715'` **>** `'1215501'` ✗ |

The comparison fails because at character position 2, `'2'` > `'1'` and `'9'` > `'1'`, so any 6-digit file number starting with a digit ≥ 2 sorts **after** a 7-digit number starting with `1`.

---

### How to Reproduce

1. Run Maxwell long enough that binlog files cross a digit-count boundary (e.g., files go from 6-digit `~122000` to 7-digit `~1000000+`).
2. Let Maxwell accumulate schema delta entries in `maxwell.schemas` across this range. The `schemas` table will have entries at both 6-digit files (e.g. `mysql-bin-changelog.121345`) and 7-digit files (e.g. `mysql-bin-changelog.1203422`).
3. Stop Maxwell. Manually advance (or set) the replication position in `maxwell.positions` to a 7-digit file (e.g. `mysql-bin-changelog.1215501`).
4. Restart Maxwell.

**Expected:** Maxwell selects the schema entry closest to `mysql-bin-changelog.1215501` — i.e., the one at `mysql-bin-changelog.1203422`.

**Actual:** Maxwell selects the schema entry at `mysql-bin-changelog.121345` (the last 6-digit file in the string order), silently skipping all delta entries at 7-digit files like `1000xxx`, `1100xxx`, `1203xxx`. In the startup log this appears as:

```
INFO  MysqlSavedSchema - Restoring schema id <ROOT_ID> (last modified at Position[BinlogPosition[mysql-bin-changelog.121345:...]])
INFO  MysqlSavedSchema - played 60 deltas in 34ms          ← only 60, not thousands
INFO  BinlogConnectorReplicator - Setting initial binlog pos to: mysql-bin-changelog.1215501:...
```

Then Maxwell immediately crashes when it encounters a TABLE_MAP event for any table created between the incorrectly selected schema position and the actual replication start.

---

### String Comparison Demonstration

You can verify the broken ordering directly in MySQL:

```sql
-- These are all numerically BEFORE 1215501, but string comparison disagrees:
SELECT
  'mysql-bin-changelog.122046'  < 'mysql-bin-changelog.1215501' AS six_digit_result,   -- returns 0 (WRONG)
  'mysql-bin-changelog.979715'  < 'mysql-bin-changelog.1215501' AS six_digit_result2,  -- returns 0 (WRONG)
  'mysql-bin-changelog.1203422' < 'mysql-bin-changelog.1215501' AS seven_digit_result; -- returns 1 (correct)
```

```
six_digit_result | six_digit_result2 | seven_digit_result
0                | 0                 | 1
```

---

### Impact

- Maxwell silently restores a schema that is potentially **millions of binlog events stale**.
- Any table created (via DDL) between the stale schema position and the replication start position will be unknown to Maxwell.
- Maxwell crashes at the first TABLE_MAP event for such a table with `Couldn't find table X in database Y`.
- This is a **silent data loss / availability** bug: no warning is emitted during startup about the schema gap.

---

### Suggested Fix

Extract the numeric suffix from `binlog_file` and compare it as an integer. One approach in MySQL:

```sql
SELECT id FROM `schemas`
WHERE deleted = 0
  AND last_heartbeat_read <= ?
  AND server_id = ?
  AND (
    CAST(SUBSTRING_INDEX(binlog_file, '.', -1) AS UNSIGNED) < CAST(SUBSTRING_INDEX(?, '.', -1) AS UNSIGNED)
    OR (
      binlog_file = ?
      AND (
        (binlog_position < ?  AND base_schema_id IS NOT NULL) OR
        (binlog_position <= ? AND base_schema_id IS NULL)
      )
    )
  )
ORDER BY last_heartbeat_read DESC,
         CAST(SUBSTRING_INDEX(binlog_file, '.', -1) AS UNSIGNED) DESC,
         binlog_position DESC
LIMIT 1;
```

Alternatively, zero-pad the numeric suffix to a fixed width on write (e.g., `printf("%010d", fileNumber)`) so lexicographic and numeric orders agree. This would require a schema migration for existing data.

---

### Environment

- **Maxwell version:** 1.40.0 (confirmed; likely affects all versions using non-GTID binlog position tracking)
- **MySQL:** AWS Aurora (variable-length binlog file numbers, no zero-padding)
- **Mode:** Non-GTID (file+offset based replication)

---

### Related Issues

- #1823 — describes a similar symptom (old schema selected on restart) from a different trigger (positions table truncation + `lastHeartbeatRead = 0`), but the same `findSchema` query is implicated.

hemanth132:
Fixes #2251

## Problem

`findSchema()` in `MysqlSavedSchema` compares and sorts `binlog_file` values as `VARCHAR` strings. This breaks when binlog file numbers have varying digit counts — a situation that arises naturally on long-running Aurora/RDS instances as the file counter grows past 6 digits.

**Example:** `mysql-bin-changelog.122046` sorts lexicographically *after* `mysql-bin-changelog.1215501` because at character position 2, `'2' > '1'`. Numerically, 122046 < 1215501.

The broken query:
```sql
WHERE (binlog_file < ?)                          -- string comparison
ORDER BY binlog_file DESC                        -- string sort
```

With a target of `mysql-bin-changelog.1215501`, this incorrectly **excludes** all schemas at files like `122046`, `979715`, etc. and then picks the "most recent" schema in string order — which may be thousands of real DDL events behind the actual replication start position. Maxwell then crashes with:

```
RuntimeException: Couldn't find table <X> in database <Y>
    at TableCache.processEvent(TableCache.java:35)
```

See the issue for a full analysis including a real production case where 6,964 schema delta entries (covering ~880,000 binlog files worth of DDLs) were silently skipped on restart.

## Fix

When the target binlog file has a parseable numeric suffix (the common case for `name.NNNNNN` format), use `CAST(SUBSTRING_INDEX(binlog_file, '.', -1) AS UNSIGNED)` in both the `WHERE` filter and `ORDER BY` so comparison is numeric rather than lexicographic.

If the suffix cannot be parsed as a number, the code falls back to the original string comparison — preserving compatibility with non-standard binlog file formats as requested.

```java
Long targetFileNumber = targetBinlogPosition.getFileNumber();
if ( targetFileNumber != null ) {
    // numeric path: CAST(SUBSTRING_INDEX(binlog_file, '.', -1) AS UNSIGNED)
} else {
    // fallback: original string comparison (unchanged)
}
```

`BinlogPosition.getFileNumber()` already implements this parse-with-null-fallback pattern (used by `newerThan()`); this PR applies the same logic at the SQL layer.

## Test

Added `testFindSchemaUsesNumericFileOrdering()` in `MysqlSavedSchemaTest` which sets up three schema snapshots at files `121345`, `122046`, and `1203422` against a target of `1215501`. Under the old code, `122046` would be filtered out and `121345` would be picked (wrong). Under the fix, `1203422` is correctly selected as the most recent numerically-prior snapshot.

### DIFF
--- a/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
+++ b/src/main/java/com/zendesk/maxwell/schema/MysqlSavedSchema.java
@@ -592,29 +592,65 @@ public class MysqlSavedSchema {
 		} else {
 			// Only consider binlog positions before the target position on the current server.
 			// Within those, sort for the latest binlog file, then the latest binlog position.
-			String sql = "SELECT id from `schemas` "
-					+ "WHERE deleted = 0 "
-					+ "AND last_heartbeat_read <= ? AND ("
-					+ "(binlog_file < ?) OR "
-					+ "(binlog_file = ? and binlog_position < ? and base_schema_id is not null) OR "
-					+ "(binlog_file = ? and binlog_position <= ? and base_schema_id is null) "
-					+ ") AND server_id = ? "
-					+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1";
-			try ( PreparedStatement s = connection.prepareStatement(sql) ) {
+			//
+			// Binlog file names contain a numeric suffix (e.g. "mysql-bin-changelog.1215501").
+			// When that suffix is parseable, we compare and sort numerically via CAST to avoid
+			// lexicographic ordering bugs where a shorter number string like "122046" compares
+			// greater than a longer one like "1215501". If the suffix cannot be parsed we fall
+			// back to string comparison so that non-standard formats still work.
+			Long targetFileNumber = targetBinlogPosition.getFileNumber();
+			if ( targetFileNumber != null ) {
+				String sql = "SELECT id from `schemas` "
+						+ "WHERE deleted = 0 "
+						+ "AND last_heartbeat_read <= ? AND ("
+						+ "(CAST(SUBSTRING_INDEX(binlog_file, '.', -1) AS UNSIGNED) < ?) OR "
+						+ "(binlog_file = ? and binlog_position < ? and base_schema_id is not null) OR "
+						+ "(binlog_file = ? and binlog_position <= ? and base_schema_id is null) "
+						+ ") AND server_id = ? "
+						+ "ORDER BY last_heartbeat_read DESC, "
+						+ "CAST(SUBSTRING_INDEX(binlog_file, '.', -1) AS UNSIGNED) DESC, "
+						+ "binlog_position DESC limit 1";
+				try ( PreparedStatement s = connection.prepareStatement(sql) ) {
+					s.setLong(1, targetPosition.getLastHeartbeatRead());
+					s.setLong(2, targetFileNumber);
+					s.setString(3, targetBinlogPosition.getFile());
+					s.setLong(4, targetBinlogPosition.getOffset());
+					s.setString(5, targetBinlogPosition.getFile());
+					s.setLong(6, targetBinlogPosition.getOffset());
+					s.setLong(7, serverID);
 
-				s.setLong(1, targetPosition.getLastHeartbeatRead());
-				s.setString(2, targetBinlogPosition.getFile());
-				s.setString(3, targetBinlogPosition.getFile());
-				s.setLong(4, targetBinlogPosition.getOffset());
-				s.setString(5, targetBinlogPosition.getFile());
-				s.setLong(6, targetBinlogPosition.getOffset());
-				s.setLong(7, serverID);
+					try ( ResultSet rs = s.executeQuery() ) {
+						if (rs.next()) {
+							return rs.getLong("id");
+						} else
+							return null;
+					}
+				}
+			} else {
+				// Fallback: file name has no parseable numeric suffix; use string comparison.
+				String sql = "SELECT id from `schemas` "
+						+ "WHERE deleted = 0 "
+						+ "AND last_heartbeat_read <= ? AND ("
+						+ "(binlog_file < ?) OR "
+						+ "(binlog_file = ? and binlog_position < ? and base_schema_id is not null) OR "
+						+ "(binlog_file = ? and binlog_position <= ? and base_schema_id is null) "
+						+ ") AND server_id = ? "
+						+ "ORDER BY last_heartbeat_read DESC, binlog_file DESC, binlog_position DESC limit 1";
+				try ( PreparedStatement s = connection.prepareStatement(sql) ) {
+					s.setLong(1, targetPosition.getLastHeartbeatRead());
+					s.setString(2, targetBinlogPosition.getFile());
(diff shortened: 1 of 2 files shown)


osheroff (approved):
lgtm

hemanth132:
@osheroff can you create a tag please? This will push the docker image which I can use from my side

osheroff:
I’d like to, currently I’m in  never ending custody battle with Zendesk for control of this repo.  Stay tuned On May 18, 2026, at 13:22, hemanth132 ***@***.***> wrote:﻿hemanth132 left a comment (zendesk/maxwell#2252)
@osheroff can you create a tag please? This will push the docker image which I can use from my side

hemanth132:
Hi @osheroff Any ETA on when the docker image will be pushed ? Will plan things accordingly from my side.

PR #2254, opened 2026-06-08
Support binlog_transaction_compression (MySQL 8.0.20+ compressed binlog)

alexpacio:
## Problem

When the source MySQL is configured with `binlog_transaction_compression=ON` (MySQL 8.0.20+), Maxwell connects successfully, reports the binlog as connected, advances its binlog position, and emits **zero row events**. No error is logged.

## Root cause

With compression on, MySQL wraps each transaction's events (`TABLE_MAP`, `WRITE/UPDATE/DELETE_ROWS`, `XID`, ...) into a single zstd-compressed `TRANSACTION_PAYLOAD` event.

In `mysql-binlog-connector-java` (verified through the current `0.30.3`):

- `TransactionPayloadEventDataDeserializer` decompresses the payload and parses the inner events into `TransactionPayloadEventData.uncompressedEvents`, but
- `BinaryLogClient` never unwraps that list — it delivers the single `TRANSACTION_PAYLOAD` event to listeners and advances the position.

Maxwell's `BinlogConnectorEventListener` only handles `GTID`/`MARIADB_GTID`/`QUERY`/`XID` + row events, so the `TRANSACTION_PAYLOAD` event is silently ignored and the inner rows — sitting decompressed but un-surfaced — are dropped.

## Fix

Two changes, both Maxwell-side (no connector bump is sufficient, since even the latest connector doesn't surface the inner events):

1. **`MaxwellTransactionPayloadDeserializer`** — a drop-in `TRANSACTION_PAYLOAD` deserializer registered via `setEventDataDeserializer(...)`. It decompresses the zstd payload and parses the inner events with the **same compatibility modes Maxwell already applies** to uncompressed events (`DATE_AND_TIME_AS_LONG_MICRO`, `CHAR_AND_BINARY_AS_BYTE_ARRAY`, `INVALID_DATE_AND_TIME_AS_MIN_VALUE`).

   This matters for correctness: the stock deserializer parses the inner events with a default `EventDeserializer`, which returns `String`/`java.util.Date` instead of `byte[]`/`Long`. That silently corrupts binary/non-utf8 columns (Base64 path is skipped) and drops sub-millisecond datetime precision. With this deserializer, a compressed transaction yields byte-for-byte the same `RowMap`s as an uncompressed one.

2. **Unwrap `TRANSACTION_PAYLOAD` in `BinlogConnectorEventListener.onEvent`** — iterate the inner events and route each through `onEvent`. Inner events carry positions relative to the in-memory decompressed buffer, so each header is re-stamped with the payload event's on-disk position. Because `EventHeaderV4.getPosition()` is derived as `nextPosition - eventLength`, setting both `eventLength` and `nextPosition` to the payload's values reproduces the payload's `[position, nextPosition]` on every inner event.

   A compressed transaction occupies a single binlog offset range, so all inner events share the payload's `nextPosition`. Position is persisted only on the TX-commit row (`MaxwellContext.setPosition(RowMap)` gates on `isTXCommit()`), so the committed offset equals the payload end — identical to where an uncompressed `XID` would land.

## Follow-up commit: gtid tracking, memory behavior, zstd hardening

A second commit fixes issues found while reviewing the first one:

- **gtid mode made no durable progress with compression on.** `BinaryLogClient.updateGtidSet` has no `TRANSACTION_PAYLOAD` case, and the `XID`/`COMMIT` events that commit the pending gtid into the client's set are inside the payload — so `getGtidSet()` froze at its startup value, `BinlogPosition.newerThan` (a gtid-set containment check) rejected every position update, and a restart replayed everything since startup. New `MaxwellBinaryLogClient` translates each payload into the synthetic `XID` event the stock bookkeeping expects (a payload always wraps exactly one complete transaction; the XID branch never reads the event argument). File/offset positioning was unaffected. Long-term this belongs in the connector (`case TRANSACTION_PAYLOAD:` in `updateGtidSet`, as upstream osheroff has) — happy to PR that separately.
- **Peak memory is now bounded.** Inner events were parsed eagerly into a list, putting the whole transaction's object graph on the heap before the first event was consumed (and `BinaryLogClient` pins the payload event until the next network event arrives, so the last large transaction stayed fully retained on a quiet stream). New `MaxwellTransactionPayloadEventData` parses inner events on demand through a single-pass cursor and releases the buffer and parser at end-of-payload.
- **zstd hardening:** one reused `ZstdDecompressCtx` instead of a native context malloc/free per transaction; `ZstdException` is caught and wrapped (zstd-jni reports errors by throwing, so the previous `Zstd.isError(rc)` check was dead code); the decompressed length is validated so truncated payloads fail loudly; compression type `NONE` is supported as passthrough; the compressed bytes are no longer retained on the event data.
- Parse failures during unwrap log at ERROR with the binlog position and skip the rest of that transaction (same outcome as the client's own handling of undeserializable events, but visible in Maxwell's logs).
- `zstd-jni` is declared explicitly in the pom (used directly; version matches the connector's transitive dependency).

Unit tests added: zstd round-trip (including integration through the outer `EventDeserializer`), NONE passthrough, corrupt/truncated/unsupported payloads, gtid advancement across compressed transactions, and listener unwrap/re-stamp behavior.

## Verification

- `mvn compile` and `mvn test-compile` pass; the new unit tests pass.
- Behavior traced against the connector bytecode (the registered `TRANSACTION_PAYLOAD` deserializer is honored; `deserializeEventData` bounds the stream to the event body excluding the checksum) and against Maxwell's position-store flow.
- **Not yet exercised against a live MySQL with `binlog_transaction_compression=ON`** — I don't have that environment handy. A maintainer-side integration test (compressed source → assert datetime/binary fidelity + position advance) would be the ideal gate before merge, and I'm happy to add one with guidance on the preferred test harness.

## Known limitation

For users running a `--javascript` filter that calls `row.suppress()` **and** an async producer (Kafka/Kinesis/SQS/SNS/Pub-Sub/BigQuery): a suppressed row advances the persisted position to the end of the whole compressed transaction (rather than a mid-transaction offset as in the uncompressed case), which widens the existing suppressed-row crash window from one row to one transaction. It does not affect normal table/database filters or sync producers. Happy to close this window (stamp non-terminal inner events with the payload start) if maintainers prefer.

## Alternative

Operators who can't wait for this can set `binlog_transaction_compression=OFF` on the source (note: already-compressed transactions before the current position stay invisible).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -161,6 +161,13 @@
       <artifactId>mysql-binlog-connector-java</artifactId>
       <version>0.30.1</version>
     </dependency>
+    <!-- used directly (transaction payload decompression); version matches the
+         binlog connector's transitive dependency -->
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <version>1.5.0-2</version>
+    </dependency>
     <dependency>
       <groupId>com.mchange</groupId>
       <artifactId>c3p0</artifactId>
--- a/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
+++ b/src/main/java/com/zendesk/maxwell/replication/BinlogConnectorEventListener.java
@@ -4,9 +4,11 @@ import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Timer;
 import com.github.shyiko.mysql.binlog.BinaryLogClient;
 import com.github.shyiko.mysql.binlog.event.Event;
+import com.github.shyiko.mysql.binlog.event.EventHeaderV4;
 import com.github.shyiko.mysql.binlog.event.EventType;
 import com.github.shyiko.mysql.binlog.event.GtidEventData;
 import com.github.shyiko.mysql.binlog.event.MariadbGtidEventData;
+import com.github.shyiko.mysql.binlog.event.TransactionPayloadEventData;
 import com.zendesk.maxwell.monitoring.Metrics;
 import com.zendesk.maxwell.producer.MaxwellOutputConfig;
 import org.slf4j.Logger;
@@ -52,6 +54,25 @@ class BinlogConnectorEventListener implements BinaryLogClient.EventListener {
 		boolean trackMetrics = false;
 
 		EventType eventType = event.getHeader().getEventType();
+
+		// binlog_transaction_compression=ON delivers a transaction's events compressed inside a
+		// single TRANSACTION_PAYLOAD event. The binlog client never unwraps it, so surface each
+		// inner event (TABLE_MAP / rows / XID ...) through the normal path here. Inner events carry
+		// positions relative to the in-memory decompressed buffer, not real binlog offsets; re-stamp
+		// each with the payload event's on-disk position so downstream position tracking stays correct
+		// (getPosition() is derived as nextPosition - eventLength, so we set both fields to the
+		// payload's values, giving every inner event the payload's position and nextPosition).
+		if ( eventType == EventType.TRANSACTION_PAYLOAD ) {
+			EventHeaderV4 payloadHeader = event.getHeader();
+			TransactionPayloadEventData payloadData = event.getData();
+			for ( Event innerEvent : payloadData.getUncompressedEvents() ) {
+				EventHeaderV4 innerHeader = innerEvent.getHeader();
+				innerHeader.setEventLength(payloadHeader.getEventLength());
+				innerHeader.setNextPosition(payloadHeader.getNextPosition());
+				onEvent(innerEvent);
+			}
+			return;
+		}
 
 		if ( eventType == EventType.GTID) {
 			gtid = ((GtidEventData)event.getData()).getGtid();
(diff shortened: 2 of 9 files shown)


marcomoauro:
I have the same problem, and so far the only way has been to disable compression to make the daemon work @osheroff .

I keep listening to the conversation

alexpacio:
As an update, this has been lately and manually tested against a MySQL 8.0.42 GTID-enabled server and it worked pretty well so far. 
Changes being highlighted in my previous message look good to me, even though i'm definitely not familiar with maxwell, nor java.
Perhaps would be worth to integrate/merge the custom deserializer's zstd handlers in the Maxwell's original deserializer instead of having a new custom class. Wondering if would even make sense to keep both.
As the commit authors suggest, this has been coded with AI.

Regarding the comment about --javascript startup option, i'm not sure of what this is about.
Also reducing comments' verbosity is definitely needed ahead of merging.

As a sidenote I can also mention that the docker image being used in the Dockerfile as base image is not available in the docker hub anymore as java 23 is EOL. I've used openjdk 21 lts in my local enviroment and it worked just fine.

osheroff:
thanks for the PR!

However, I think the correct way to fix this is to open a PR here: https://github.com/osheroff/mysql-binlog-connector-java -- just have the binlog connector transparently uncompress the payload and send it up through the connector.  Maxwell can just upgrade its library and not even know compression is enabled. 

thanks again!

alexpacio:
@osheroff sure. 
There's the related implementation in the connector side: https://github.com/osheroff/mysql-binlog-connector-java/pull/177
Closing this. 
We'll need to bump the pom file for dep update once released.

